PR #429, opened 2023-09-27
fix(sim): fix gas limit calcs

dancoombs:


dphilipson (approved):

crates/sim/src/gas/gas.rs
```diff
@@ -132,8 +114,10 @@ pub fn user_operation_execution_gas_limit(uo: &UserOperation, chain_id: u64) ->
 }
 
 /// Returns the maximum cost, in wei, of this user operation
-pub fn user_operation_max_gas_cost(uo: &UserOperation, chain_id: u64) -> U256 {
-    user_operation_gas_limit(uo, chain_id) * uo.max_fee_per_gas
+pub fn user_operation_max_gas_cost(uo: &UserOperation) -> U256 {
+    let mul = if uo.paymaster().is_some() { 3 } else { 1 };
```
alex-miao, line 118: In the no paymaster case why is this now 1 but in `user_operation_execution_gas_limit` it's still 2, we can still hit the `innerHandleOp` code and incur cost there

dancoombs: Good question. The `2` is there because of that incorrect check that requires us to have one `validationGasLimit` buffer left in the transaction's gas. However, since no `postOp` will ever run when not using a paymaster, the maximum a user will ever need to pay for is `1` `validationGasLimit`

alex-miao (approved):

PR #435, opened 2023-09-27
chore(codecov): add codecov reporting for CI pipeline

0xfourzerofour:
## Proposed Changes

  - Add codecov to the CI test runner

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -1,0 +1,52 @@
+on:
+  pull_request:
+  merge_group:
+  push:
+    branches: [main]
+
+env:
+  CARGO_TERM_COLOR: always
+
+name: unit
+jobs:
+  test:
+    name: unit-tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
+      - name: Install toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools-preview
+
+      - name: Install protobuf
+        run: sudo apt-get install -y protobuf-compiler
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+
+      - name: Install latest nextest release
+        uses: taiki-e/install-action@nextest
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Run tests
+        run: |
+          cargo llvm-cov nextest --lcov --output-path lcov.info \
+            --locked --all-features --workspace
+
+      - name: Upload coverage data to codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: lcov.info
+          flags: unit-tests
+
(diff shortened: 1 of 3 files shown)


.github/workflows/unit.yaml
```diff
@@ -0,0 +1,64 @@
...
+            -E 'kind(lib)' -E 'kind(bin)' -E 'kind(proc-macro)'
+
+      - name: Upload coverage data to codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: lcov.info
+          flags: unit-tests
+
+  unit-success:
```
dancoombs, line 54: What is this for?

.github/workflows/unit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Install latest nextest release
+        uses: taiki-e/install-action@nextest
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Run tests
+        run: |
+          cargo llvm-cov nextest --lcov --output-path lcov.info \
+            --locked --all-features --workspace --exclude examples --exclude ef-tests  \
```
dancoombs, line 44: What are these `--excludes`?

.github/workflows/unit.yaml
```diff
@@ -0,0 +1,64 @@
...
+            -E 'kind(lib)' -E 'kind(bin)' -E 'kind(proc-macro)'
```
dancoombs, line 45: What do these do?

.github/workflows/unit.yaml
0xfourzerofour: its something that I took from reth, By we need to get the unit tests running concurrently for it to work

.github/workflows/unit.yaml
0xfourzerofour: They just explicitly define what are being tests but I removed for now as we only have lib code being tested anyway

.github/workflows/unit.yaml
0xfourzerofour: I removed as they are not needed

.github/workflows/unit.yaml
dancoombs: Lets drop it for now unless its being used

codecov.yaml
```diff
@@ -0,0 +1,54 @@
+coverage:
```
dancoombs, line 1: Can we move this file into `.github/workflows`

.github/workflows/unit.yaml
```diff
@@ -0,0 +1,63 @@
+on:
```
dancoombs, line 1: We have `ci.yml` and `unit.yaml` lets consolidate on one, I think most people use `.yaml` these days

codecov.yaml
0xfourzerofour: Yeh I can, however I think it might be better to have a configs directory where we can put everything as its a bit weird testing locally using a config file in a directory specific to github

dancoombs: I don't think codecov is anything we'd use locally? Seems pretty specific to github to me

0xfourzerofour: Yeh i know but i just mean for general use for all of our config files

dancoombs: I think for now keeping config files top level is fine if they are used locally. If they get out of hand we can move them. For github specific we should move

dancoombs (approved):
LGTM, be sure to squash

PR #437, opened 2023-09-27
chore(codecov): add push format

0xfourzerofour:
Fix github action `on` selectors to match ci


.github/codecov.yaml
```diff
@@ -0,0 +1,53 @@
+on:
```
dancoombs, line 1: I think you copied the wrong file?

0xfourzerofour: oh my bad, serious brain fog today

dancoombs (approved):

Issue #349, opened 2023-09-07
builder: Improve bundle gas limit efficiency


dancoombs:
**Describe the feature**
Currently we only use ~20% of the gas limit of our bundle transactions. This is due to overheads required when estimating the gas to be used during validation and postOp (3x the validationGasLimit when a paymaster is used). We also add a 10% overhead to our limit.

We should look into ways to improve this.

The downside to an inefficient gas limit is that it limits the amount of UOs we can fit in a certain gas cap. If we improve this we can bundle more UOs per block.

PR #442, opened 2023-10-03
fix(gas): amortize evm transaction fixed cost over bundle

alex-miao:
#349

## Proposed Changes

  - Only count the 21000 transaction gas cost once in the entire bundle

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -45,6 +45,8 @@ const TIME_RANGE_BUFFER: Duration = Duration::from_secs(60);
 const BUNDLE_TRANSACTION_GAS_OVERHEAD_BUFFER: u64 = 5000;
 /// Extra buffer percent to add on the bundle transaction gas estimate to be sure it will be enough
 const BUNDLE_TRANSACTION_GAS_OVERHEAD_PERCENT: u64 = 5;
+/// The fixed gas overhead for any EVM transaction
+const EVM_TRANSACTION_GAS_OVERHEAD: u64 = 21000;
 
 #[derive(Debug, Default)]
 pub(crate) struct Bundle {
@@ -538,7 +540,7 @@ where
         let mut gas_left = U256::from(self.settings.max_bundle_gas);
         let mut ops_in_bundle = Vec::new();
         for op in ops {
-            let gas = gas::user_operation_gas_limit(&op.uo, self.settings.chain_id);
+            let gas = gas::user_operation_gas_limit(&op.uo, self.settings.chain_id, false);
             if gas_left < gas {
                 self.emit(BuilderEvent::skipped_op(
                     self.builder_index,
@@ -740,9 +742,10 @@ impl ProposalContext {
 
     fn get_total_gas_limit(&self, chain_id: u64) -> U256 {
         self.iter_ops()
-            .map(|op| gas::user_operation_gas_limit(op, chain_id))
+            .map(|op| gas::user_operation_gas_limit(op, chain_id, false))
             .fold(U256::zero(), |acc, c| acc + c)
             + BUNDLE_TRANSACTION_GAS_OVERHEAD_BUFFER
+            + EVM_TRANSACTION_GAS_OVERHEAD
     }
 
     fn iter_ops_with_simulations(&self) -> impl Iterator<Item = &OpWithSimulation> + '_ {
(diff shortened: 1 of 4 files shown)


crates/sim/src/gas/gas.rs
```diff
@@ -74,7 +72,7 @@ pub async fn calc_pre_verification_gas<P: Provider>(
     provider: Arc<P>,
     chain_id: u64,
 ) -> anyhow::Result<U256> {
-    let static_gas = calc_static_pre_verification_gas(full_op);
+    let static_gas = calc_static_pre_verification_gas(full_op, true);
```
alex-miao, line 75: We want this to be true here, because with `eth_estimateUserOperationGas` we have to assume that this is the only transaction in the bundle

crates/builder/src/bundle_proposer.rs
```diff
@@ -45,6 +45,8 @@ const TIME_RANGE_BUFFER: Duration = Duration::from_secs(60);
 const BUNDLE_TRANSACTION_GAS_OVERHEAD_BUFFER: u64 = 5000;
 /// Extra buffer percent to add on the bundle transaction gas estimate to be sure it will be enough
 const BUNDLE_TRANSACTION_GAS_OVERHEAD_PERCENT: u64 = 5;
+/// The fixed gas overhead for any EVM transaction
+const EVM_TRANSACTION_GAS_OVERHEAD: u64 = 21000;
```
dancoombs, line 49: Could we share this constant with the `21000` that we have hardcoded in the pvg calculation? Or somehow get it from the `GasOverheads` config?

crates/sim/src/gas/gas.rs
```diff
@@ -38,7 +38,6 @@ struct GasOverheads {
     per_user_op_word: U256,
     zero_byte: U256,
     non_zero_byte: U256,
-    bundle_size: U256,
```
dancoombs, line 41: What was the reasoning behind removing this?

crates/sim/src/precheck.rs
```diff
@@ -180,7 +180,7 @@ impl<P: Provider, E: EntryPoint> PrecheckerImpl<P, E> {
                 max_verification_gas,
             ));
         }
-        let total_gas_limit = gas::user_operation_gas_limit(op, chain_id);
+        let total_gas_limit = gas::user_operation_gas_limit(op, chain_id, false);
```
dancoombs, line 183: I think this one should be true?

crates/sim/src/gas/gas.rs
alex-miao: We don't use it anywhere now. Do you still want to keep this?

crates/builder/src/bundle_proposer.rs
alex-miao: refactored it so that the bundle gas calculation is now in `gas.rs`

crates/sim/src/gas/gas.rs
```diff
@@ -94,13 +94,31 @@ pub async fn calc_pre_verification_gas<P: Provider>(
     Ok(static_gas + dynamic_gas)
 }
 
+/// Compute the gas limit for the bundle composed of the given user operations
+pub fn bundle_gas_limit<'a, I>(iter_ops: I, chain_id: u64) -> U256
+where
+    I: Iterator<Item = &'a UserOperation>,
+{
+    let ov = GasOverheads::default();
```
alex-miao, line 102: somewhat unnecessary duplication since we are creating the `GasOverheads` object both here and within `user_operation_gas_limit`, but favoring this over needing to make `GasOverheads` public.

crates/builder/src/bundle_proposer.rs
```diff
@@ -786,7 +781,8 @@ mod tests {
             op.pre_verification_gas
                 + op.verification_gas_limit * 2
                 + op.call_gas_limit
-                + BUNDLE_TRANSACTION_GAS_OVERHEAD_BUFFER,
+                + U256::from(5_000)
+                + U256::from(21_000),
```
dancoombs, line 785: Is it possible to get these from some common place? Tests will break if we ever modify these values.

These are the sorts of things I want to move to the `ChainSpec` config class in the future.

alex-miao: Changed to exposing the relevant fields of `GasOverheads`, it's not great but just avoiding a major refactor for now.

Issue #362, opened 2023-09-12
builder: Remove polling of mined transactions


0xfourzerofour:
**Describe the feature**
Currently the transaction_tracker polls for mined transactions. We need to update this to be updated by the pool instead.

Remove the polling of mined transactions and us updates from the pool instead.

PR #445, opened 2023-10-05
feat(builder): add bundle gas stats to metrics

### LINKED ISSUE #379: builder: Metric for gas used throughput
**Describe the feature**
We should add a metric to the builder that allows us to graph gas used and gas limit throughputs in grafana.

alex-miao:
Closes #379 

## Proposed Changes

  - Log gas limit, gas used, and gas limit utilization for mined bundles

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -552,11 +552,8 @@ impl BuilderMetrics {
     }
 
     fn set_bundle_gas_stats(gas_limit: U256, gas_used: U256) {
-        let gas_limit = gas_limit.as_u128() as f64;
-        let gas_used = gas_used.as_u128() as f64;
-        metrics::gauge!("builder_bundle_gas_limit", gas_limit);
-        metrics::gauge!("builder_bundle_gas_used", gas_used);
-        metrics::gauge!("builder_bundle_gas_utilization", gas_limit / gas_used);
+        metrics::counter!("builder_bundle_gas_limit", gas_limit.as_u64());
+        metrics::counter!("builder_bundle_gas_used", gas_used.as_u64());
     }
 
     fn set_current_fees(fees: &GasFees) {
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -262,9 +262,12 @@ where
                 tx_hash,
                 block_number,
                 attempt_number,
+                gas_limit,
+                gas_used,
                 ..
             } => {
                 BuilderMetrics::increment_bundle_txns_success(self.builder_index);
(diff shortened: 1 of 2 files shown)


crates/builder/src/transaction_tracker.rs
```diff
@@ -379,6 +387,14 @@ where
             TransactionTrackerMetrics::set_current_fees(None);
         }
     }
+
+    async fn get_mined_tx_gas_info(&self, tx_hash: H256) -> anyhow::Result<(U256, U256)> {
+        let (tx, tx_receipt) = tokio::try_join!(
+            self.provider.get_transaction(tx_hash),
```
alex-miao, line 393: We technically already have info about the gas limit of the transaction, but it's a pretty messy refactor to pass it all the way through to where we can log it so choosing to just make this extra request instead.

0xfourzerofour (approved):

crates/builder/src/bundle_sender.rs
```diff
@@ -551,6 +557,14 @@ impl BuilderMetrics {
         metrics::increment_counter!("builder_bundle_fee_increases", "builder_index" => builder_index.to_string());
     }
 
+    fn set_bundle_gas_stats(gas_limit: U256, gas_used: U256) {
+        let gas_limit = gas_limit.as_u128() as f64;
+        let gas_used = gas_used.as_u128() as f64;
+        metrics::gauge!("builder_bundle_gas_limit", gas_limit);
+        metrics::gauge!("builder_bundle_gas_used", gas_used);
+        metrics::gauge!("builder_bundle_gas_utilization", gas_limit / gas_used);
```
dancoombs, line 565: Why did you decide to use a gauge here? These metrics might be hard to visualize.

I think these should likely be counters. `gas_limit` and `gas_used`. And then we can take the derivative across a minute for each to get gas rates, then divide the gas rates to get gas efficiency.

crates/builder/src/transaction_tracker.rs
dancoombs: This info is in the pool, yeah? Part of https://github.com/alchemyplatform/rundler/issues/362 will be to pipe more information about mined transactions from pool to builder, so lets improve this then.

crates/builder/src/bundle_sender.rs
alex-miao: Oh interesting, I didn't consider that approach. Are we able to do this pointwise division between two different time series in Grafana?

dancoombs: We aren't, but the idea is to get a general sense of what our trend is there, so I think aggregated metrics are fine. For pointwise stuff, we can use the event logs.

dancoombs (approved):

PR #456, opened 2023-10-05
feat(spectests): fix timeout and run in CI

### LINKED ISSUE #147: ci: incorporate bundler-spec-tests into CI
**Describe the feature**
Run the `bundler-spec-tests` in CI and fail PRs if we have regressions.

0xfourzerofour:
Closes #147

## Proposed Changes
  - Add spec tests to CI and fix timeout issues due to race condition
  - Update `debug_sendBundleNow` so that we wait for the pool to be updated to the same transaction block before returning

### DIFF
--- a/.github/workflows/compliance.yaml
+++ b/.github/workflows/compliance.yaml
@@ -1,0 +1,40 @@
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+name: compliance
+jobs:
+  compliance:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+           submodules: recursive
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'yarn'
+          cache-dependency-path: test/spec-tests/bundler-spec-tests/@account-abstraction/yarn.lock
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+      - name: Setup PDM
+        uses: pdm-project/setup-pdm@v3
+        with:
+          cache: true
+          cache-dependency-path: '**/pdm.lock'
+      - name: Run spec tests
+        run: ./test/spec-tests/ci/run-spec-tests.sh
+      - name: Move cache
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -11,12 +11,9 @@
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
 
-use std::{
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
-    time::Duration,
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
 };
 
 use anyhow::{bail, Context};
@@ -31,7 +28,6 @@ use rundler_utils::emit::WithEntryPoint;
 use tokio::{
     join,
     sync::{broadcast, mpsc, oneshot},
-    time,
 };
 use tracing::{error, info, trace, warn};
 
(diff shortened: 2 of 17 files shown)


.github/workflows/compliance.yaml
```diff
@@ -0,0 +1,40 @@
...
+        uses: pdm-project/setup-pdm@v3
+        with:
+          cache: true
+          cache-dependency-path: '**/pdm.lock'
+      - name: Run spec tests
+        run: ./test/spec-tests/ci/run-spec-tests.sh
+      - name: Move cache
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
```
dancoombs, line 40: Curious why this was needed?

0xfourzerofour: Its a bug in the actions/cache runner that will not auto cleanup so this makes sure the cache does not continually grow for 7 days. 10gb max

crates/rpc/src/debug.rs
```diff
@@ -95,10 +96,32 @@ where
...
+            .map_err(|e| rpc_err(INTERNAL_ERROR_CODE, e.to_string()))?;
+
+        let (tx, block_number) = self
+            .builder
             .debug_send_bundle_now()
             .await
-            .map_err(|e| rpc_err(INTERNAL_ERROR_CODE, e.to_string()))
+            .map_err(|e| rpc_err(INTERNAL_ERROR_CODE, e.to_string()))?;
+
+        loop {
```
dancoombs, line 111: Worth adding a comment here explaining why we're doing this.

crates/builder/src/transaction_tracker.rs
```diff
@@ -180,7 +180,7 @@ where
         let nonce = provider
             .get_transaction_count(sender.address())
             .await
-            .context("tracker should load initial nonce on construction")?;
+            .unwrap_or(U256::zero());
```
dancoombs, line 183: Did we determine the implications if this change? Do we eventually get the correct nonce before submitting the transaction?

crates/builder/src/bundle_sender.rs
```diff
@@ -141,6 +141,7 @@ where
 
         loop {
             let mut send_bundle_response: Option<oneshot::Sender<SendBundleResult>> = None;
+            let mut last_block = None;
 
             if self.manual_bundling_mode.load(Ordering::Relaxed) {
                 tokio::select! {
```
dancoombs, line 147: I think we can remove the `time::sleep(..)` from this (and remove the select). That was leftover from the pre-refactor and not needed

crates/rpc/src/debug.rs
0xfourzerofour: Yeh I can add a comment

crates/builder/src/bundle_sender.rs
```diff
@@ -192,7 +195,7 @@ where
                     } else {
                         info!("Bundle with hash {tx_hash:?} landed in block {block_number} after increasing gas fees {attempt_number} time(s)");
                     }
-                SendBundleResult::NoOperationsInitially => trace!("No ops to send at block {}", last_block.block_number),
+                SendBundleResult::NoOperationsInitially => trace!("No ops to send at block {}", last_block.unwrap().block_number),
```
dancoombs, line 198: Is this `.unwrap()` dangerous? `last_block` starts as `None` above. Probably want an `unwrap_or(0)`

crates/builder/src/transaction_tracker.rs
0xfourzerofour: In the transaction tracker logic there seems to be a final check on send transaction that rechecks the nonce if it is the reason for failure

.github/workflows/compliance.yaml
0xfourzerofour: Note, it might make sense to cache into s3 because its way faster instead of copying the data internally to gihub actions datastore

crates/builder/src/transaction_tracker.rs
0xfourzerofour: https://github.com/alchemyplatform/rundler/blob/main/crates/builder/src/transaction_tracker.rs#L240C26-L240C26

Here for reference

crates/builder/src/bundle_sender.rs
```diff
@@ -141,44 +136,44 @@ where
...
-                    Some(r) = self.send_bundle_receiver.recv() => {
-                        send_bundle_response = Some(r.responder);
-                    }
-                    _ = time::sleep(self.eth_poll_interval) => {
-                        continue;
-                    }
+                if let Some(r) = self.send_bundle_receiver.recv().await {
+                    send_bundle_response = Some(r.responder);
+                } else {
+                    continue;
```
dancoombs, line 145: This shouldn't happen. We can either panic here, or if we're trying to be defensive we can log an error and turn manual bundling mode off. I'm leaning towards panic.

0xfourzerofour: I added a bail that will cause the task to stop the application

dancoombs (approved):

PR #457, opened 2023-10-06
feat(builder): only include paymaster gas if there is a postOp

alex-miao:
#349

## Proposed Changes

  - In the gas limit calculation, only multiply the verification gas limit by 3 if we know the user operation uses a paymaster that has a postOp

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -30,9 +30,7 @@ use mockall::automock;
 use rundler_pool::{PoolOperation, PoolServer};
 use rundler_provider::{EntryPoint, HandleOpsOut, Provider};
 use rundler_sim::{
-    gas::{
-        self, user_operation_gas_limit, user_operation_pre_verification_gas_limit, GasOverheads,
-    },
+    gas::{self, GasOverheads},
     ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError, SimulationSuccess, Simulator,
 };
 use rundler_types::{Entity, EntityType, GasFees, Timestamp, UserOperation, UserOpsPerAggregator};
@@ -132,7 +130,9 @@ where
             self.builder_index,
             ops.len(),
         );
-        let (ops, gas_limit) = self.limit_gas_in_bundle(ops);
+
+        // Do an initial filtering of ops that we want to simulate.
+        let (ops, gas_limit) = self.limit_user_operations_for_simulation(ops);
         tracing::debug!(
             "Builder index: {}, bundle proposal after limit had {} ops and {:?} gas limit",
             self.builder_index,
@@ -189,10 +189,6 @@ where
             .await;
         while !context.is_empty() {
             let gas_estimate = self.estimate_gas_rejecting_failed_ops(&mut context).await?;
-            let mut expected_storage = ExpectedStorage::default();
-            for op in context.iter_ops_with_simulations() {
-                expected_storage.merge(&op.simulation.expected_storage)?;
-            }
             if let Some(gas_estimate) = gas_estimate {
                 tracing::debug!(
                     "Builder index: {}, bundle proposal succeeded with {} ops and {:?} gas limit",
@@ -200,6 +196,12 @@ where
                     context.iter_ops().count(),
                     gas_estimate
                 );
+
+                let mut expected_storage = ExpectedStorage::default();
+                for op in context.iter_ops_with_simulations() {
+                    expected_storage.merge(&op.simulation.expected_storage)?;
+                }
+
                 return Ok(Bundle {
                     ops_per_aggregator: context.to_ops_per_aggregator(),
                     gas_estimate,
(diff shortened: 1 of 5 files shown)


crates/sim/src/gas/gas.rs
```diff
@@ -329,68 +308,3 @@ const NON_EIP_1559_CHAIN_IDS: &[u64] = &[
...
-                Bytes::from(vec![0; 20])
-            } else {
-                Default::default()
-            },
-            ..Default::default()
-        }
-    }
-
-    #[tokio::test]
-    async fn test_bundle_gas_limit() {
```
dancoombs, line 359: Should these tests move too?

crates/builder/src/bundle_proposer.rs
```diff
@@ -736,8 +739,32 @@ impl ProposalContext {
...
+                + user_operation_pre_verification_gas_limit(op, chain_id, false)
+                + op.verification_gas_limit * 2
+                + op.call_gas_limit
+                + post_exec_req_gas;
+            max_gas = cmp::max(required_gas, max_gas);
+            gas_spent += user_operation_gas_limit(
+                op,
+                chain_id,
+                false,
+                Some(op_with_sim.simulation.requires_post_op),
```
dancoombs, line 763: Totally forgot that we run the simulation in the builder so this information is available to us after that. Won't be the case when we remove the simulation, but works for now!

crates/builder/src/bundle_proposer.rs
```diff
@@ -536,7 +539,7 @@ where
         let mut gas_left = U256::from(self.settings.max_bundle_gas);
         let mut ops_in_bundle = Vec::new();
         for op in ops {
-            let gas = gas::user_operation_gas_limit(&op.uo, self.settings.chain_id, false);
+            let gas = gas::user_operation_gas_limit(&op.uo, self.settings.chain_id, false, None);
```
dancoombs, line 542: Why is postop `None` here?

dancoombs: Seems like we need to account for that when we're limiting the gas in the bundle.

dancoombs: This is annoying, however, since we don't have this information here since we haven't simulated yet. Ideally we limit gas in bundle before simulating...

crates/sim/src/gas/gas.rs
alex-miao: Yep, just added another commit to move them over to `bundle_proposer.rs`

crates/builder/src/bundle_proposer.rs
alex-miao: Yep agree we need to account for it when limiting gas in the bundle, so the way it gets accounted for is that the `gas::user_operation_gas_limit` function takes in an `Option<bool>` representing whether we have a `postOp` or not. The `None` state represents when we have not done simulation yet, and the `gas::verification_gas_limit_multiplier` function defaults to the worst case of 3. We pass in `Some(true)` and `Some(false)` only post-simulation when we actually know if we use `postOp` or not.

dancoombs: Won't that mean that we're going to create bundles well below our limit if they all use postops?

alex-miao: Or rather the other way around, if they a lot of UOs use paymasters that don't have postOp, but this is the same behavior that the bundler has now. This PR only improves the gas limit that we submit the bundle with. To improve the actual UO limiting efficiency we would need to do something like do simulation before limiting the bundle size.

alex-miao: One approach we could do is to have an initial pre-sim bundle gas limiting phase like we have now. We can add UOs up until the gas limit, but for UOs that have a paymaster, we can compute an optimistic and worst-case gas usage (optimistic being that they all have no postOp, worst-case being they all use postOp). We then sim these UOs, but also a set of UOs which we deem as "backup" in the case that we end up having more space than we expected.

dancoombs: Hm, improving the gas limit we submit the bundle with doesn't really improve our ability to land more UOs on chain. We likely need to do more.

How about this?

* Limit bundle assuming that non of the UOs use post op
* Simulate
* Further limit after knowing if UOs use post op.

dancoombs: Looks like we had a similar idea. It must be good then...

alex-miao: Sounds good, since this represents a shift to limiting post sim, might also make sense here to take into account UOs that fail validation

alex-miao: Made #461 to address this issue, would be good to get this one in first since #461 depends on this.

dancoombs (commented):
Initial LGTM. Will wait to review once #461 merges

dancoombs (approved):
🚢

Issue #248, opened 2023-06-14
pool: Implement P2P mempool 


dancoombs:
**Describe the feature**
Implement https://github.com/eth-infinitism/bundler-spec/blob/main/p2p-specs/p2p-interface.md

PR #460, opened 2023-10-10
feat(network): add rpc messages and codec

dancoombs:
Starts #248 

## Proposed Changes

  - Starts the `rundler-network` crate
  - Defines all of the request/response methods from the [spec](https://github.com/eth-infinitism/bundler-spec/blob/main/p2p-specs/p2p-interface.md#messages)
  - Defines all of the supported libp2p protocols.
      -  Note: some of this is dead code at the moment, but will be used in an upcoming PR. Made sense to combine the definitions in this PR.
  - Implements Tokio [Decoder](https://docs.rs/tokio-util/latest/tokio_util/codec/trait.Decoder.html) and [Encoder](https://docs.rs/tokio-util/latest/tokio_util/codec/trait.Encoder.html) traits for both Inbound (requests from peers) and Outbound (requests too peers) directions.

Notes:

- Some of the TODOs left in the code are the result of ambiguous or missing wording in the spec. I'll be following up with fixes once discussed with the community.
- There are also some TODOs around the structure of the RPC messages that should be worked out. 

Next up:

- Use these messages to create the RPC libp2p behavior.
- Peer discovery and management

### DIFF
--- a/crates/network/src/rpc/message.rs
+++ b/crates/network/src/rpc/message.rs
@@ -1,0 +1,214 @@
+// This file is part of Rundler.
+//
+// Rundler is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Lesser General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version.
+//
+// Rundler is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+// without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along with Rundler.
+// If not, see https://www.gnu.org/licenses/.
+
+use ethers::types::H256;
+use rundler_types::UserOperation;
+use ssz_derive::{Decode, Encode};
+
+pub(crate) const MAX_ERROR_MESSAGE_LEN: usize = 256;
+pub(crate) const MAX_OPS_PER_REQUEST: usize = 4096;
+
+/// Request types
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum Request {
+    Status(Status),
+    Goodbye(Goodbye),
+    Ping(Ping),
+    Metadata,
+    PooledUserOpHashes(PooledUserOpHashesRequest),
+    PooledUserOpsByHash(PooledUserOpsByHashRequest),
+}
+
+impl Request {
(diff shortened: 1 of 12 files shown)


crates/network/src/rpc/protocol.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+impl AsRef<str> for ProtocolSchema {
+    fn as_ref(&self) -> &str {
+        match self {
+            ProtocolSchema::StatusV1 => "status/1",
+            ProtocolSchema::GoodbyeV1 => "goodbye/1",
+            ProtocolSchema::PingV1 => "ping/1",
+            ProtocolSchema::MetadataV1 => "metadata/1",
+            ProtocolSchema::PooledUserOpHashesV1 => "pooled_user_op_hashes/1",
```
alex-miao, line 90: Might be an inconsistency in the spec, but the protocol id in the spec adds an "s" and calls it `pooled_user_ops_hashes` 

https://github.com/eth-infinitism/bundler-spec/blob/main/p2p-specs/p2p-interface.md#pooleduserophashes

crates/network/src/rpc/message.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// the last one received. If so, request new metadata from the peer.
+    pub(crate) metadata_seq_number: u64,
+}
+
+// Metadata response
+#[derive(Clone, Debug, PartialEq, Encode, Decode)]
+pub(crate) struct Metadata {
+    /// Metadata sequence number
+    pub(crate) seq_number: u64,
+    // TODO: spec has an invalid field here, add if needed
```
alex-miao, line 136: Is this referring to `mempool_nets`? Why is it invalid?

crates/network/src/rpc/protocol.rs
dancoombs: Good eye, I have this on my list of changes to request to make in the spec. The protocol name should match the message name, which doesn't have the "s".

crates/network/src/rpc/message.rs
dancoombs: This is another thing I wanted to discuss with the spec writers, I don't believe the mempool nets field is valid as defined. There currently is no definition for a "mempool id gossip subnets". We use the gossip topic to signify different  domains for mempool ids.

0xfourzerofour (approved):
looks  good from what I can see!

alex-miao (approved):

PR #461, opened 2023-10-10
feat(builder): add two phase UO gas limit filtering

alex-miao:
#349

## Proposed Changes

Add a two phase bundle gas filtering process. In the first phase, we filter the UOs in the pool down to a subset that we want to simulate. This phase intentionally includes more UOs than necessary in order to account for any UOs that might be dropped later on. The second phase of filtering happens after simulation. At this point we are able to get a better gas limit estimate for each UO because we know for sure if the UO uses a `postOp` or not, and also drop UOs that fail validation. This filtering is when we ensure that all the UOs we have selected satisfy the maximum bundle gas limit.

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -30,9 +30,7 @@ use mockall::automock;
 use rundler_pool::{PoolOperation, PoolServer};
 use rundler_provider::{EntryPoint, HandleOpsOut, Provider};
 use rundler_sim::{
-    gas::{
-        self, user_operation_gas_limit, user_operation_pre_verification_gas_limit, GasOverheads,
-    },
+    gas::{self, GasOverheads},
     ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError, SimulationSuccess, Simulator,
 };
 use rundler_types::{Entity, EntityType, GasFees, Timestamp, UserOperation, UserOpsPerAggregator};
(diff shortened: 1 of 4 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -311,6 +321,20 @@ where
...
+            let required_gas = get_gas_required_for_op(
+                gas_spent,
+                self.settings.chain_id,
+                ov,
+                &op,
+                simulation.requires_post_op,
+            );
+            if required_gas > self.settings.max_bundle_gas.into() {
+                println!("max gas: {}", self.settings.max_bundle_gas);
+                println!("required gas: {}, op: {}", required_gas, op.sender);
```
dancoombs, line 334: Remove prints

crates/builder/src/bundle_proposer.rs
```diff
@@ -340,6 +364,15 @@ where
                     *balance -= max_cost;
                 }
             }
+
+            // Update the running gas that would need to be be spent to execute the bundle so far.
+            gas_spent += user_operation_gas_limit(
```
dancoombs, line 369: nit: I like to keep functions scoped to the module they're defined in. So here `gas::user_operation_gas_limit`

crates/builder/src/bundle_proposer.rs
```diff
@@ -409,7 +442,7 @@ where
         // sum up the gas needed for all the ops in the bundle
         // and apply an overhead multiplier
         let gas = math::increase_by_percent(
-            context.get_total_gas_limit(self.settings.chain_id),
+            context.get_bundle_gas_limit(self.settings.chain_id),
```
dancoombs, line 445: Is this different than the `gas_spent` variable above? Could we just save that with the context?

dancoombs (approved):
LGTM. Great work!

crates/builder/src/bundle_proposer.rs
alex-miao: fixed, also did the same for `gas::user_operation_pre_verification_gas_limit`

PR #464, opened 2023-10-12
chore(docker): do not overide cache with COPY

0xfourzerofour:
I was looking into the cargo chef build speeds in CI and found a few interesting things that have been implemented and recommended by the chef project.

this was first proposed in https://github.com/LukeMathWalker/cargo-chef/issues/180 and eventually zcach found a workaround with rzync in the following PR https://github.com/ZcashFoundation/zebra/pull/6934

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN curl -L https://foundry.paradigm.xyz | bash
 ENV PATH="/root/.foundry/bin:${PATH}"
 RUN foundryup
 
-RUN cargo install cargo-chef 
+RUN cargo install cargo-chef --locked
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -36,6 +36,11 @@ RUN cargo chef cook --profile $BUILD_PROFILE --recipe-path recipe.json
 
 # Build application
 COPY . .
+
+COPY --from=planner /app recipe-original
+RUN rsync --recursive --checksum --itemize-changes --verbose recipe-original/ .
+RUN rm -r recipe-original
```
dancoombs, line 42: Can you leave a comment explaining what this is doing and why

dancoombs:
What is the impact here?

0xfourzerofour:
> What is the impact here?

No impact in application, however this should in theory speed up deployments and docker builds in CI and locally. I will test it with a dummy staging deploy before this is merged

dancoombs:
> > What is the impact here?
> 
> No impact in application, however this should in theory speed up deployments and docker builds in CI and locally. I will test it with a dummy staging deploy before this is merged

Yeah thats what I mean, what is the speedup to build times?

0xfourzerofour:
> > > What is the impact here?
> > 
> > 
> > No impact in application, however this should in theory speed up deployments and docker builds in CI and locally. I will test it with a dummy staging deploy before this is merged
> 
> Yeah thats what I mean, what is the speedup to build times?

Worst case before was around 2200s now its around 1500 without being cached. Note: this is worst case due to deployments from local machine that requires QEMU as we are using apple silicon. On CI it should be faster than that due to using an x86 image

alex-miao (approved):
Great find!

dancoombs (approved):
🚀

PR #467, opened 2023-10-14
fix(pool): add back already known check after refactor

dancoombs:


crates/pool/src/mempool/pool.rs
```diff
@@ -226,35 +234,8 @@ impl PoolInner {
...
-                self.get_min_replacement_fees(pool_op.uo());
-
-            // replace only if higher gas
-            if op.uo.max_priority_fee_per_gas >= replacement_priority_fee
-                && op.uo.max_fee_per_gas >= replacement_fee
-            {
-                self.remove_operation_by_hash(
-                    pool_op
-                        .uo()
-                        .op_hash(self.config.entry_point, self.config.chain_id),
```
dancoombs, line 249: Hm this logic is missing now.

0xfourzerofour (approved):
LGTM

PR #471, opened 2023-10-17
fix(builder): attempt to re-lock lost key locks

dancoombs:
## Proposed Changes

  - Attempts to re-lock lost locks, instead of looping trying to extend them

### DIFF
--- a/crates/builder/src/signer/aws.rs
+++ b/crates/builder/src/signer/aws.rs
@@ -16,7 +16,7 @@ use std::{sync::Arc, time::Duration};
 use anyhow::Context;
 use ethers::providers::Middleware;
 use ethers_signers::{AwsSigner, Signer};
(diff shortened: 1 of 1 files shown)


crates/builder/src/signer/aws.rs
```diff
@@ -73,43 +73,58 @@ impl KmsSigner {
...
-                    continue;
-                }
+            if let Some(l) = try_lock(&lm, lock_id, ttl_millis as usize).await {
+                lock = Some(l);
+                kid = Some(key_id.clone());
+                locked_id = Some(lock_id.clone());
+                break;
             }
         }
         if lock.is_none() {
```
0xfourzerofour, line 90: Nit: these could use a `if let Some(lock)` or match to get rid of unwraps but just preference

0xfourzerofour (approved):

Issue #455, opened 2023-10-05
[spec] UREP-030


dancoombs:
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L255

> * **[UREP-030]** If an unstaked entity causes an invalidation of a bundle, its `opsSeen` is set to `1000`, effectively blocking it from inclusion for 24 hours.

We don't do this, we just remove from the pool.

Issue #454, opened 2023-10-05
[spec] UREP-20


dancoombs:
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L252C39-L252C39

> * **[UREP-020]** For other entities: \
    `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
    * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for new entity

We don't track the reputation of unstaked entities and we don't apply this limit on entity inclusion, only to unstaked senders.

dancoombs:
Should do at the same time as #354

Issue #451, opened 2023-10-05
[spec] SREP-050


dancoombs:
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L230C1-L231C1

> * **[SREP-050]** If a staked entity fails the second validation or fails bundle creation, its `opsSeen` is incremented by `10000`, causing it to be `BANNED`.

We currently just drop the entity, we need to also cause it to be `BANNED`.

Issue #449, opened 2023-10-05
[spec] SREP-020


dancoombs:
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L221

> * **[SREP-020]** A `BANNED` address is not allowed into the mempool.\
  Also, all existing `UserOperations` referencing this address are removed from the mempool.

We need to remove all UOs referencing an address when that entity is banned.

PR #478, opened 2023-10-31
feat(pool): add basic structure for update entities pool operation

alex-miao:
Progress towards #451, #455, #454, #449

## Proposed Changes

  - expose the update_entities call in the mempool that is mean't to replace remove_entities to allow for more flexibility in the types of updates the pool can handle


crates/pool/src/mempool/reputation.rs
```diff
@@ -169,19 +186,31 @@ impl ReputationManager for HourlyMovingAverageReputation {
...
+    bundle_invalidation_ops_seen_unstaked_penalty: u64,
+    same_unstaked_entity_mempool_count: u64,
     min_inclusion_rate_denominator: u64,
+    inclusion_rate_factor: u64,
     throttling_slack: u64,
     ban_slack: u64,
 }
 
 impl ReputationParams {
     pub(crate) fn bundler_default() -> Self {
```
0xfourzerofour, line 207: We could probably implement the Default trait here and then on the `bundler_default` and `client_default` we could just have the unique field values followed by ..Self::default(). Just removes some duplicate code

crates/pool/src/mempool/reputation.rs
```diff
@@ -268,6 +312,20 @@ impl AddressReputation {
         count.ops_included = ops_included;
     }
 
+    fn get_ops_allowed(&self, address: Address) -> u64 {
+        let count = self.counts.get(&address).unwrap();
```
0xfourzerofour, line 316: might want to have a check here or or unwrap_or default just to make sure its never called when the address has no count. I know it most likely never will but just to be sure there is no panic

crates/types/src/entity.rs
```diff
@@ -63,7 +63,7 @@ impl FromStr for EntityType {
 /// An entity associated with a user operation
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
 pub struct Entity {
-    /// The tpe of entity
+    /// The the of entity
```
0xfourzerofour, line 66: shoudnt this be type?

0xfourzerofour (commented):
Looks good, just a couple comments

crates/pool/src/mempool/reputation.rs
alex-miao: seems reasonable to just default to a `seen` of 0 since the address has never been seen before.

0xfourzerofour (approved):

PR #481, opened 2023-11-02
feat(error): add entity kind to error message

### LINKED ISSUE #480: Non-compliant Storage Access Restrictions
Sorry if this is already captured in an existing issue. Also note that I am using `runder` over Alchemy and not running it locally.

I believe there may be an issue with slot restrictions in the `validateUserOp` simulation. In particular, ERC-4337 specifies a "slot associated with an address `A`" to be:

```
Slots of type `keccak256(A || X) + n` on any other address. (to cover `mapping(address => value)`, which is usually used for balance in ERC-20 tokens). `n` is an offset value up to 128, to allow accessing fields in the format `mapping(address => struct)`
```

From my understanding, all slots associated with the account (user operation `sender` field) are allowed to be accessed as part of `validateUserOp`, regardless of which contract they are on.

Additionally, the spec puts **no restriction on `X`**, which notably means that things like `mapping(key => mapping(address => value))` should also be supported as the accessed slot is:
```
keccak256(address || keccak256(key || mapping_slot))
  <=> keccak256(A || X) + n
      where A = address
            X = keccak256(key || mapping_slot)
            n = 0
```

However, I noticed that when submitting user operations that access such slots, that the following error is returned (user operation request included 👇):

```ts
{
  code: -32505,
  message: 'entity stake/unstake delay too low',
  data: {
    entity: { account: '0x0C13AF2986b1bf9E082cd66fA49229D64D73E21A' },
    minimumStake: '0x16345785d8a0000',
    minimumUnstakeDelay: '0x15180'
  }
}
```

I would expect such a user operation to be allowed. Removing the [storage access](https://goerli.etherscan.io/address/0x93aa2f6d83D668Cc50e2Af1a37EB049bee6A69b2#code#F26#L76) (commenting it out for example) causes the bundler to accept the user operation once again.

### Sample `eth_sendUserOperation` Request

The following `eth_sendUserOperation` demonstrates the issue. It was done on **Görli**.

<details><summary><code>request.json</code></summary>

```json
{
  "method": "eth_sendUserOperation",
  "params": [
    {
      "nonce": "0x00",
      "callData": "0x7bb3742800000000000000000000000002270bd144e70ce6963ba02f575776a16184e1e600000000000000000000000000000000000000000000000000005af3107a4000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "verificationGasLimit": "0x076ca6",
      "preVerificationGas": "0xc820",
      "callGasLimit": "0x6000",
      "maxFeePerGas": "0x3b9aca20",
      "maxPriorityFeePerGas": "0x3b9aca00",
      "initCode": "0x4e1dcf7ad4e460cfd30791ccc4f9c8a4f820ec671688f0b900000000000000000000000041675c099f32341bf84bfc5382af534df5c7461a0000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001e4b63e800d000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000010000000000000000000000008483eb040e5c0a24a34281f5fe7cb662c800aabb000000000000000000000000000000000000000000000000000000000000014000000000000000000000000093aa2f6d83d668cc50e2af1a37eb049bee6a69b20000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000feb6565bdd4aec9bd836f3ec7c7c0ffa79d025a400000000000000000000000000000000000000000000000000000000000000648d0dc49f0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000093aa2f6d83d668cc50e2af1a37eb049bee6a69b20000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "paymasterAndData": "0x",
      "sender": "0x0C13AF2986b1bf9E082cd66fA49229D64D73E21A",
      "signature": "0x8e8087941cc796b08d52ee65e1569d0b284308b5ef1bad4c2d41c454fb09e7f41878724b598f5898017dfbea533b0d17cde7d7805a848d555eb767cd0106ea1b1c"
    },
    "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789"
  ],
  "id": 4337,
  "jsonrpc": "2.0"
}
```

</details>

0xfourzerofour:
Fixes #480 

## Proposed Changes

  - Update error message to show the actual entity kind to remove confusion

### DIFF
--- a/crates/sim/src/simulation/simulation.rs
+++ b/crates/sim/src/simulation/simulation.rs
@@ -313,7 +313,7 @@ where
                     });
                     match restriction {
                         StorageRestriction::Allowed => {}
-                        StorageRestriction::NeedsStake => needs_stake = true,
+                        StorageRestriction::NeedsStake(_) => needs_stake = true,
                         StorageRestriction::Banned => {
                             banned_slots_accessed.insert(StorageSlot {
                                 address,
@@ -684,7 +684,7 @@ fn is_staked(info: StakeInfo, sim_settings: Settings) -> bool {
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum StorageRestriction {
     Allowed,
-    NeedsStake,
+    NeedsStake(EntityType),
     Banned,
 }
 
@@ -715,15 +715,18 @@ fn get_storage_restriction(args: GetStorageRestrictionArgs<'_>) -> StorageRestri
     } else if slots_by_address.is_associated_slot(sender_address, slot) {
         // Allow entities to access the sender's associated storage unless its during an unstaked wallet creation
         // Can always access the entry point's associated storage (note only depositTo is allowed to be called)
-        if accessed_address == entry_point_address || !is_unstaked_wallet_creation {
+        if entity_address == sender_address
+            || accessed_address == entry_point_address
+            || !is_unstaked_wallet_creation
+        {
             StorageRestriction::Allowed
         } else {
-            StorageRestriction::NeedsStake
+            StorageRestriction::NeedsStake(EntityType::Factory)
         }
     } else if accessed_address == entity_address
         || slots_by_address.is_associated_slot(entity_address, slot)
     {
-        StorageRestriction::NeedsStake
+        StorageRestriction::NeedsStake(EntityType::Account)
     } else {
         StorageRestriction::Banned
     }
--- a/crates/rpc/src/eth/error.rs
+++ b/crates/rpc/src/eth/error.rs
@@ -73,7 +73,7 @@ pub enum EthRpcError {
     #[error("entity throttled or banned")]
     ThrottledOrBanned(Entity),
     /// Entity stake/unstake delay too low
-    #[error("entity stake/unstake delay too low")]
+    #[error("{} stake/unstake delay too low", .0.entity.kind)]
     StakeTooLow(StakeTooLowData),
     /// Unsupported aggregator
     #[error("unsupported aggregator")]


crates/sim/src/simulation/simulation.rs
```diff
@@ -715,15 +715,18 @@ fn get_storage_restriction(args: GetStorageRestrictionArgs<'_>) -> StorageRestri
...
             StorageRestriction::Allowed
         } else {
-            StorageRestriction::NeedsStake
+            StorageRestriction::NeedsStake(EntityType::Factory)
         }
     } else if accessed_address == entity_address
         || slots_by_address.is_associated_slot(entity_address, slot)
     {
-        StorageRestriction::NeedsStake
+        StorageRestriction::NeedsStake(EntityType::Account)
```
dphilipson, line 729: This should be the type of the currently executing entity, not necessarily the account.

dphilipson, line 729: This should be the type of the currently executing entity, not necessarily the account.

PR #484, opened 2023-11-06
feat(pool): update throttling logic

### LINKED ISSUE #450: [spec] SREP-030
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L223

> * **[SREP-030]** A `THROTTLED` address is limited to:
    * `THROTTLED_ENTITY_MEMPOOL_COUNT` entries in the mempool.
    * `THROTTLED_ENTITY_BUNDLE_COUNT` `UserOperations` in a bundle.
    * Can remain in the mempool only for `THROTTLED_ENTITY_LIVE_BLOCKS`.

We don't enforce `THROTTLED_ENTITY_BUNDLE_COUNT`

alex-miao:
Fixes #450 

## Proposed Changes

  - add constants related to throttling and update throttling logic to use the constants

### DIFF
--- a/bin/rundler/src/cli/mod.rs
+++ b/bin/rundler/src/cli/mod.rs
@@ -45,7 +45,7 @@ pub async fn run() -> anyhow::Result<()> {
         .context("metrics server should start")?;
 
     match opt.command {
-        Command::Node(args) => node::run(args, opt.common).await?,
+        Command::Node(args) => node::run(*args, opt.common).await?,
         Command::Pool(args) => pool::run(args, opt.common).await?,
         Command::Rpc(args) => rpc::run(args, opt.common).await?,
         Command::Builder(args) => builder::run(args, opt.common).await?,
@@ -62,7 +62,7 @@ enum Command {
     ///
     /// Runs the Pool, Builder, and RPC servers in a single process.
     #[command(name = "node")]
-    Node(NodeCliArgs),
+    Node(Box<NodeCliArgs>),
 
     /// Rpc command
     ///
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -101,6 +101,22 @@ pub struct PoolArgs {
         env = "POOL_CHAIN_UPDATE_CHANNEL_CAPACITY"
     )]
     pub chain_update_channel_capacity: Option<usize>,
+
+    #[arg(
+        long = "pool.throttled_entity_mempool_count",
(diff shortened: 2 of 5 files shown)


bin/rundler/src/cli/mod.rs
```diff
@@ -62,7 +62,7 @@ enum Command {
     ///
     /// Runs the Pool, Builder, and RPC servers in a single process.
     #[command(name = "node")]
-    Node(NodeCliArgs),
+    Node(Box<NodeCliArgs>),
```
alex-miao, line 65: Needed to fix this clippy error: https://rust-lang.github.io/rust-clippy/master/index.html#/large_enum_variant

bin/rundler/src/cli/mod.rs
```diff
@@ -45,7 +45,7 @@ pub async fn run() -> anyhow::Result<()> {
         .context("metrics server should start")?;
 
     match opt.command {
-        Command::Node(args) => node::run(args, opt.common).await?,
+        Command::Node(args) => node::run(*args, opt.common).await?,
```
0xfourzerofour, line 48: Why does this need to be dereferenced? Not a needed change but would like to know

alex-miao: This is from changing the enum to use `Box<NodeCliArgs>`

bin/rundler/src/cli/mod.rs
0xfourzerofour: Ah right yeh this makes sense

bin/rundler/src/cli/pool.rs
```diff
@@ -101,6 +101,22 @@ pub struct PoolArgs {
...
+    )]
+    pub throttled_entity_mempool_count: u64,
+
+    #[arg(
+        long = "pool.throttled_entity_live_blocks",
+        name = "pool.throttled_entity_live_blocks",
+        env = "POOL_THROTTLED_ENTITY_LIVE_BLOCKS",
+        default_value = "4"
+    )]
+    pub throttled_entity_live_blocks: u64,
```
dancoombs, line 119: These are the sort of things I want to move to `ChainSpec` in the future. Linking https://github.com/alchemyplatform/rundler/issues/486 to track.

dancoombs (approved):

PR #487, opened 2023-11-07
feat(cross-arch): add cross arch release pipeline

### LINKED ISSUE #421: ci: release pipeline with cross-arch builds
**Describe the feature**
Build a release pipeline that:

1. Runs release level testing
2. Increments the version
3. Publishes a new CHANGELOG
4. Builds docker containers for cross-arch
5. Uploads docker containers

This should be done automatically from the `release/*` branches upon branch creation (major/minor versions) or new merges (only fix versions) or manually.

0xfourzerofour:
Closes #421 

## Proposed Changes

  - Cross arch docker and binary releases
  - Remove git submodule update from build.rs due to release pipeline incompatibility (https://github.com/cross-rs/cross/issues/1358)
  - Update docker container to allow users to run app in different modes


.github/workflows/docker-release.yaml
```diff
@@ -0,0 +1,53 @@
+name: Docker release
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - v*
```
dancoombs, line 7: Is this a regex? Maybe something more specific like `v(\d+\.)?(\d+\.)?(\*|\d+)` so that we only tag valid semantic version tags?

.github/workflows/docker-release.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: alchemyplatform
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up Docker builder
+        run: |
+          docker run --privileged --rm tonistiigi/binfmt --install arm64,amd64
```
dancoombs, line 47: What is the `--rm tonistiigi/binfmt` for?

0xfourzerofour: it makes sure you ahve the right platform emulator in the github action for cross arch builds

.github/workflows/release.yaml
```diff
@@ -0,0 +1,216 @@
+## This release action is inpired from https://githug.com/paradigmxyz/reth
```
dancoombs, line 1: nit spelling: `inspired`

.github/workflows/docker-release.yaml
0xfourzerofour: it is not regex unfortunately, I am pretty sure it only accepts wildcards but we might be able to go like `v*.*.*`

Cargo.toml
```diff
@@ -43,3 +43,4 @@ tower = "0.4.13"
 tracing = "0.1.37"
 strum = "0.25.0"
 url = "2.3.1"
+
```
dancoombs, line 46: nit: remove whitespace change

crates/types/build.rs
```diff
@@ -19,7 +19,6 @@ fn main() -> Result<(), Box<dyn error::Error>> {
     println!("cargo:rerun-if-changed=contracts/lib");
     println!("cargo:rerun-if-changed=contracts/src");
     println!("cargo:rerun-if-changed=contracts/foundry.toml");
-    update_submodules()?;
```
dancoombs, line 22: What was the reason for removing this?

Makefile
```diff
@@ -28,4 +33,43 @@ test-spec-integrated: ## Run spec tests in integrated mode
...
+	docker buildx build --file ./Dockerfile.cross . \
+		--platform linux/arm64,linux/amd64 \
+		--tag $(DOCKER_IMAGE_NAME):$(1) \
+		--tag $(DOCKER_IMAGE_NAME):$(2) \
+		--provenance=false 
+endef
+
+
+
+# docker buildx build --file ./Dockerfile.cross . \
```
dancoombs, line 70: nit: remove commented code, or comment why its commented out

Makefile
```diff
@@ -28,4 +33,43 @@ test-spec-integrated: ## Run spec tests in integrated mode
 .PHONY: test-spec-modular
 test-spec-modular: ## Run spec tests in modular mode
 	test/spec-tests/remote/run-spec-tests.sh
+
+.PHONY: submodule-update
+submodule-update: ## Run spec tests in modular mode
```
dancoombs, line 38: nit: fix comment

crates/types/build.rs
0xfourzerofour: https://github.com/actions/checkout/issues/766

it is to do with this, I added the conversion with cross-rs team above

Makefile
```diff
@@ -28,4 +33,43 @@ test-spec-integrated: ## Run spec tests in integrated mode
...
+	git submodule update
+
+build-%:
+	cross build --bin rundler --target $* --profile "$(PROFILE)"
 	
+# Note: This requires a buildx builder with emulation support. For example:
```
dancoombs, line 44: nit: `For example:` doesn't have the example

Makefile
```diff
@@ -28,4 +33,43 @@ test-spec-integrated: ## Run spec tests in integrated mode
...
+.PHONY: docker-build-latest
+docker-build-latest: ## Build and push a cross-arch Docker image tagged with the latest git tag and `latest`.
+	$(call build_docker_image,$(GIT_TAG),latest)
+
+# Create a cross-arch Docker image with the given tags and push it
+define build_docker_image
+	echo $(GIT_TAG)
```
dancoombs, line 51: Do we want to leave this in? Seems like just echoing a git tag isn't a lot of info, maybe add some context?

Dockerfile.cross
```diff
@@ -0,0 +1,15 @@
+# This image is meant to enable cross-architecture builds.
+# It assumes the reth binary has already been compiled for `$TARGETPLATFORM` and is
```
dancoombs, line 2: reth -> rundler. Add a comment that this was taken from reth

Makefile
0xfourzerofour: oh yeh that was for testing I can remove

Makefile
```diff
@@ -28,4 +33,43 @@ test-spec-integrated: ## Run spec tests in integrated mode
...
+submodule-update: ## Run spec tests in modular mode
+	git submodule update
+
+build-%:
+	cross build --bin rundler --target $* --profile "$(PROFILE)"
 	
+# Note: This requires a buildx builder with emulation support. For example:
+.PHONY: docker-build-latest
+docker-build-latest: ## Build and push a cross-arch Docker image tagged with the latest git tag and `latest`.
+	$(call build_docker_image,$(GIT_TAG),latest)
```
dancoombs, line 50: Any reason why this is broken out as a function? Seems like we only use it here

dancoombs: Might make sense to have a version of this where we don't tag with `latest` for releasing patch versions

.github/workflows/release.yaml
```diff
@@ -0,0 +1,216 @@
...
+            | <img src="https://simpleicons.org/icons/linux.svg" style="width: 32px;"/> | aarch64 | [rundler-${{ env.VERSION }}-aarch64-unknown-linux-gnu.tar.gz](https://github.com/${{ env.REPO_NAME }}/releases/download/${{ env.VERSION }}/rundler-${{ env.VERSION }}-aarch64-unknown-linux-gnu.tar.gz) | [PGP Signature](https://github.com/${{ env.REPO_NAME }}/releases/download/${{ env.VERSION }}/rundler-${{ env.VERSION }}-aarch64-unknown-linux-gnu.tar.gz.asc) |
+            | <img src="https://simpleicons.org/icons/windows.svg" style="width: 32px;"/> | x86_64 | [rundler-${{ env.VERSION }}-x86_64-pc-windows-gnu.tar.gz](https://github.com/${{ env.REPO_NAME }}/releases/download/${{ env.VERSION }}/rundler-${{ env.VERSION }}-x86_64-pc-windows-gnu.tar.gz) | [PGP Signature](https://github.com/${{ env.REPO_NAME }}/releases/download/${{ env.VERSION }}/rundler-${{ env.VERSION }}-x86_64-pc-windows-gnu.tar.gz.asc) |
```
dancoombs, line 202: Do we build for windows?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,216 @@
...
+          name: rundler-${{ needs.extract-version.outputs.VERSION }}-${{ matrix.arch }}.tar.gz
+          path: rundler-${{ needs.extract-version.outputs.VERSION }}-${{ matrix.arch }}.tar.gz
+
+      - name: Upload signature
+        uses: actions/upload-artifact@v3
+        with:
+          name: rundler-${{ needs.extract-version.outputs.VERSION }}-${{ matrix.arch }}.tar.gz.asc
+          path: rundler-${{ needs.extract-version.outputs.VERSION }}-${{ matrix.arch }}.tar.gz.asc
+
+  draft-release:
```
dancoombs, line 127: I'm not familiar with the "draft release" flow. What is the order of operations here for getting to a release?

0xfourzerofour: it will create a draft release that we would just have to double check and then approve it on the release page on github

.github/workflows/release.yaml
```diff
@@ -0,0 +1,216 @@
...
+      - name: Get latest version of stable Rust
+        run: rustup update stable
+      - name: Install target
+        run: rustup target add ${{ matrix.arch }}
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+
+      # ==============================
+      # Apple M1 SDK setup
```
dancoombs, line 64: Nit: maybe remove the M1 from this. I think mac is on M3 now? Maybe we just call this apple setup

.github/workflows/release.yaml
```diff
@@ -0,0 +1,216 @@
...
+    name: build release
+    strategy:
+      matrix:
+        arch:
+          [
+            aarch64-unknown-linux-gnu,
+            x86_64-unknown-linux-gnu,
+            x86_64-apple-darwin,
+            aarch64-apple-darwin,
+            x86_64-pc-windows-gnu,
```
dancoombs, line 31: I don't see these 3 getting built in the makefile, is there a different pathway here?

.github/workflows/docker-release.yaml
dancoombs: Yeah I'd prefer that if it works

0xfourzerofour:
Add workflow dispatch to test this before deployment

.github/workflows/docker-release.yaml
```diff
@@ -0,0 +1,53 @@
...
+          docker buildx create --use --name cross-builder
+
+      - name: Build and push image
+        run: |
+          cargo install cross --git https://github.com/cross-rs/cross 
+          sudo -E env "PATH=$PATH" make docker-build-latest
```
dancoombs, line 53: This will cause every build to be tagged with `latest`, correct? Concerned about patch versions squashing latest. Lets add a ticket to follow up on that in 0.2.

dancoombs (commented):
Code LGTM. Final couple requests here:

1. How much of this is tested? What is the testing plan? Do we need to merge first and then iterate?

2. Can you write 2 docs to support this code:
  a. release.md explaining the release process
  b. Update [docker.md](https://github.com/alchemyplatform/rundler/blob/main/docs/docker.md) explaining the use of the 3 docker files, an overview of how to do builds, note anything that we had to do what wasn't intuitive, etc.

0xfourzerofour:
The annoying thing about github is you cannot test updated workflows unless they are first in the default branch. We will need to merge first and then I will manually run each of them. I have added an override for the git tag in the workflow dispatch for patches and inital testing deploying to dockerhub

.github/workflows/docker-release.yaml
```diff
@@ -0,0 +1,62 @@
...
+          docker buildx create --use --name cross-builder
+
+      - name: Build and push image
+        run: |
+          cargo install cross --git https://github.com/cross-rs/cross 
+
+          if [ -n "${{ github.event.inputs.version }}" ]; then
+            make GIT_TAG="${{ github.event.inputs.version }}" docker-build 
+          else
+            sudo -E env "PATH=$PATH" make docker-build-latest
```
dancoombs, line 61: Why are these commands different? Why does one need sudo?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,226 @@
...
+          ]
+        include:
+          - arch: aarch64-unknown-linux-gnu
+            platform: ubuntu-latest
+            profile: maxperf
```
dancoombs, line 46: Is `maxperf` defined as a profile in `Cargo.toml`? We need to define it: https://github.com/paradigmxyz/reth/blob/4e5ae6c540fb8085ee102586c5caf1cd20379e4b/Cargo.toml#L106

I think we can stick with release for now though

docs/docker.md
```diff
@@ -31,3 +31,49 @@ services:
...
+          docker run --privileged --rm tonistiigi/binfmt --install arm64,amd64
+          docker buildx create --use --name cross-builder
+
+      - name: Build and push image
+        run: |
+          cargo install cross --git https://github.com/cross-rs/cross 
+          sudo -E env "PATH=$PATH" make docker-build-latest
+```
+
+**Local Builds**
```
dancoombs, line 72: We should note that this is only needed if you're trying to do cross-platform builds locally. Else they should just use the simple docker command above.

Dockerfile.build
```diff
@@ -0,0 +1,18 @@
...
+
+RUN apt-get update && apt-get install -y gnupg2 apt-transport-https ca-certificates software-properties-common
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN mkdir -p /etc/apt/keyrings
+RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list
+RUN apt-get update && apt-get -y upgrade && apt-get install -y libclang-dev pkg-config protobuf-compiler nodejs yarn
+RUN add-apt-repository ppa:ethereum/ethereum
+RUN apt-get update
+RUN apt-get install -y solc
```
dancoombs, line 18: Is it possible to share this code with `Dockerfile`? If complicated, no worries, just not ideal that its in 2 places.

0xfourzerofour: We could potentially create a base docker image and use that as the base for both but I am not sure if that would work so I will have to test

.github/workflows/docker-release.yaml
0xfourzerofour: I was using sudo during test to make sure the path was correct but can delete now

dancoombs (approved):
:shipit:

PR #488, opened 2023-11-07
fix(storage-rules): check storage rules for the correct entity

0xfourzerofour:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/sim/src/simulation/simulation.rs
+++ b/crates/sim/src/simulation/simulation.rs
@@ -295,8 +295,12 @@ where
                     entity, contract, precompile,
                 ));
             }
+
             let mut needs_stake = entity.kind == EntityType::Paymaster
                 && !entry_point_out.return_info.paymaster_context.is_empty();
+
+            let mut needs_stake_entity_type = None;
+
             let mut banned_slots_accessed = IndexSet::<StorageSlot>::new();
             for StorageAccess { address, slots } in &phase.storage_accesses {
                 let address = *address;
@@ -306,14 +310,17 @@ where
                         slots_by_address: &tracer_out.associated_slots_by_address,
                         is_unstaked_wallet_creation,
                         entry_point_address: self.entry_point_address,
-                        entity_address: entity_info.address,
+                        entity,
                         sender_address,
                         accessed_address: address,
                         slot: *slot,
                     });
                     match restriction {
                         StorageRestriction::Allowed => {}
-                        StorageRestriction::NeedsStake => needs_stake = true,
+                        StorageRestriction::NeedsStake(e) => {
+                            needs_stake = true;
+                            needs_stake_entity_type = Some(e);
+                        }
                         StorageRestriction::Banned => {
                             banned_slots_accessed.insert(StorageSlot {
                                 address,
@@ -324,10 +331,17 @@ where
                 }
             }
             if needs_stake {
-                entities_needing_stake.push(entity.kind);
+                let entity_type = needs_stake_entity_type.unwrap();
+                entities_needing_stake.push(entity_type);
+
+                let entity_info = entity_infos.get(entity_type).unwrap();
+
                 if !entity_info.is_staked {
                     violations.push(SimulationViolation::NotStaked(
-                        entity,
+                        Entity {
+                            kind: entity_type,
+                            address: entity_info.address,
+                        },
                         self.sim_settings.min_stake_value.into(),
                         self.sim_settings.min_unstake_delay.into(),
                     ));
@@ -684,7 +698,7 @@ fn is_staked(info: StakeInfo, sim_settings: Settings) -> bool {
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum StorageRestriction {
     Allowed,
-    NeedsStake,
+    NeedsStake(EntityType),
     Banned,
 }
 
@@ -693,7 +707,7 @@ struct GetStorageRestrictionArgs<'a> {
     slots_by_address: &'a AssociatedSlotsByAddress,
     is_unstaked_wallet_creation: bool,
     entry_point_address: Address,
-    entity_address: Address,
+    entity: Entity,
     sender_address: Address,
     accessed_address: Address,
     slot: U256,
@@ -704,7 +718,7 @@ fn get_storage_restriction(args: GetStorageRestrictionArgs<'_>) -> StorageRestri
         slots_by_address,
         is_unstaked_wallet_creation,
         entry_point_address,
-        entity_address,
+        entity,
         sender_address,
         accessed_address,
         slot,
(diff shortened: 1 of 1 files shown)


crates/sim/src/simulation/simulation.rs
```diff
@@ -295,8 +295,12 @@ where
                     entity, contract, precompile,
                 ));
             }
+
             let mut needs_stake = entity.kind == EntityType::Paymaster
                 && !entry_point_out.return_info.paymaster_context.is_empty();
+
+            let mut needs_stake_entity_type = None;
```
dphilipson, line 302: Later, you do
```rust
if needs_state {
    let entity_type = needs_stake_entity_type.unwrap();
    // ...
}
```
which will panic if `needs_stake` is initialized to `true` and this isn't updated further.

I think you don't actually need the boolean, you can just use the option by itself, which would remove the need for the `.unwrap()`, e.g.
```rust
let needs_stake_entity_type = if ... {
    Some(EntityType::Paymaster)
} else {
    None
}
```

crates/sim/src/simulation/simulation.rs
```diff
@@ -306,14 +310,17 @@ where
...
                         sender_address,
                         accessed_address: address,
                         slot: *slot,
                     });
                     match restriction {
                         StorageRestriction::Allowed => {}
-                        StorageRestriction::NeedsStake => needs_stake = true,
+                        StorageRestriction::NeedsStake(e) => {
+                            needs_stake = true;
+                            needs_stake_entity_type = Some(e);
```
dphilipson, line 322: This gets the case where both the paymaster and factory need stake. If above we determined that the paymaster requires stake and then we find out here that the factory requires stake, the factory will replace the value of `needs_stake_entity_type` and we won't end up verifying that the paymaster has stake.

crates/sim/src/simulation/simulation.rs
```diff
@@ -715,15 +729,18 @@ fn get_storage_restriction(args: GetStorageRestrictionArgs<'_>) -> StorageRestri
...
         // Can always access the entry point's associated storage (note only depositTo is allowed to be called)
-        if accessed_address == entry_point_address || !is_unstaked_wallet_creation {
+        if entity.address == sender_address
+            || accessed_address == entry_point_address
+            || !is_unstaked_wallet_creation
+        {
             StorageRestriction::Allowed
         } else {
-            StorageRestriction::NeedsStake
+            StorageRestriction::NeedsStake(EntityType::Factory)
```
dphilipson, line 738: Similar to the above comment, this also misses the case where both the factory and currently executing entity need stake- we'll only end up checking one or the other, depending on which storage is accessed first.

dphilipson (commented):
Looks good! A few comments inline.

crates/sim/src/simulation/simulation.rs
0xfourzerofour: yeh this is why I was like its messy haha I will defs clean up

crates/sim/src/simulation/simulation.rs
0xfourzerofour: do you think we should check if paymaster needs stake and then if its overriding we add the paymaster to entities_needing_stake vec before overriding ?

dphilipson: Hm, yea this is awkward. I also just realized that now that we can have one entity (the factory) needing stake during the execution of another entity, that with the current code structure we'll end up adding the same violation for the factory not having stake multiple times. I'll think about this.

dphilipson: What we may want is to collect a set of entities-needing-stake across all entities, then compute the violations once we've processed all entities.

0xfourzerofour: @dphilipson I added a hashmap so that I did not need to implement Hasher for Enum types and just used the existing to_str implementation for the key. This checks for storage restrictions across all the entities and then I loop through to add the entity and the stake information

dphilipson: Sounds good, but not sure I understand why not to implement hash for enum types. Is there a drawback to slapping a `#[derive(Hash, Eq)]` on it?

0xfourzerofour: oh thats handy I though when implementing hashing you needed to implement with a random seed and there was no macro but this is clean I  just added

0xfourzerofour: looks like the spec tests were failing with the extra condition `entity_address == sender_address` in the get_storage_restriction function

0xfourzerofour:
```
    StorageTestCase(
        "account_reference_storage_init_code",
        UNSTAKED,
        SENDER,
        with_initcode(build_userop_for_sender),
        assert_unstaked_error,
    ),


```

This is the spec test in question

crates/sim/src/simulation/simulation.rs
```diff
@@ -295,8 +297,17 @@ where
...
                 ));
             }
-            let mut needs_stake = entity.kind == EntityType::Paymaster
-                && !entry_point_out.return_info.paymaster_context.is_empty();
+
+            if entity.kind == EntityType::Paymaster
+                && !entry_point_out.return_info.paymaster_context.is_empty()
+                && !entity_info.is_staked
+            {
+                entity_infos.add_to_storage_restrictions(
```
dancoombs, line 305: I was confused by this being called "storage_restrictions" since in this case its being added because the paymaster returns a context. Maybe just call it `entity_type_requires_stake` or something?

crates/sim/src/simulation/simulation.rs
```diff
@@ -671,6 +689,24 @@ impl EntityInfos {
         }
     }
 
+    fn add_to_storage_restrictions(
```
dancoombs, line 692: From a code organization perspective its confusing to me that this function is on `EntityInfos`. Maybe better just as a static method on the simulator?

crates/sim/src/simulation/simulation.rs
```diff
@@ -358,6 +367,15 @@ where
...
+        for ent in entity_type_storage_restrictions.values() {
+            entities_needing_stake.push(ent.kind);
+            violations.push(SimulationViolation::NotStaked(
+                *ent,
+                self.sim_settings.min_stake_value.into(),
+                self.sim_settings.min_unstake_delay.into(),
+            ));
+        }
+
         if let Some(aggregator_info) = entry_point_out.aggregator_info {
```
dancoombs, line 379: Could we use the `entity_type_storage_restrictions` map to handle the aggregator here and then move the loop above to below this function? So that we collect all of the entities that need stake, and then add all the violations at once?

In general I'm having a hard time reasoning about the  `entities_needing_stake` `entity_type_storage_restrictions` and `violations` (with NotStaked) collections. Can we simplify?

crates/sim/src/simulation/simulation.rs
```diff
@@ -718,12 +754,12 @@ fn get_storage_restriction(args: GetStorageRestrictionArgs<'_>) -> StorageRestri
         if accessed_address == entry_point_address || !is_unstaked_wallet_creation {
             StorageRestriction::Allowed
         } else {
-            StorageRestriction::NeedsStake
+            StorageRestriction::NeedsStake(EntityType::Factory)
```
dancoombs, line 757: For my understanding, this is the bug fix for #480, correct?

crates/sim/src/simulation/simulation.rs
0xfourzerofour: Yeh I was not sure about this, as I thought passing in the reference of both might seems a little weird but I can reformat to a static method

crates/sim/src/simulation/simulation.rs
0xfourzerofour: Yes that what I understand, Dphil seemed to agree that it should be fine

crates/sim/src/simulation/simulation.rs
0xfourzerofour: Yeh I agree its a little verbose and could be simplified

crates/sim/src/simulation/simulation.rs
dancoombs: Is there a spec test for this?

crates/sim/src/simulation/simulation.rs
0xfourzerofour: updated

dancoombs (approved):

PR #490, opened 2023-11-08
fix(builder): allow for none transaction receipt

### LINKED ISSUE #489: Debug tokio runtime panics
```
{
  "timestamp": "2023-11-01T18:41:21.332409Z",
  "level": "ERROR",
  "fields": {
    "message": "Builder server error: task 230 panicked"
  },
  "target": "rundler_builder::task"
}

```

This error has happened periodically and we would like to see the stack trace

alex-miao:
Closes #489

### DIFF
--- a/crates/builder/src/transaction_tracker.rs
+++ b/crates/builder/src/transaction_tracker.rs
@@ -20,7 +20,7 @@ use rundler_provider::Provider;
 use rundler_sim::ExpectedStorage;
 use rundler_types::GasFees;
 use tokio::time;
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::sender::{TransactionSender, TxStatus};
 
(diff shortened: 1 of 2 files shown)


crates/builder/src/transaction_tracker.rs
```diff
@@ -388,12 +388,16 @@ where
         }
     }
 
-    async fn get_mined_tx_gas_info(&self, tx_hash: H256) -> anyhow::Result<(U256, U256)> {
+    async fn get_mined_tx_gas_info(&self, tx_hash: H256) -> anyhow::Result<(U256, Option<U256>)> {
```
0xfourzerofour, line 391: we might want to log this for our own research so that we can get to the bottom of what is happening with the nodes

0xfourzerofour (approved):

PR #493, opened 2023-11-10
feat(mempool): check for same nonce and sender 

### LINKED ISSUE #278: pool: Remove ops by sender/nonce when mined
**Describe the feature**
If an op with the same sender/nonce is mined, we should remove the op from the mempool. We currently only remove ops by hash.

0xfourzerofour:
Closes #278 

## Proposed Changes

  - Use the useroperation id to get the hash instead so we can make sure we are removing the correct item from the hashmap

### DIFF
--- a/crates/pool/src/mempool/pool.rs
+++ b/crates/pool/src/mempool/pool.rs
@@ -31,6 +31,7 @@ use super::{
     size::SizeTracker,
     PoolConfig, PoolOperation,
 };
+use crate::chain::MinedOp;
 
 #[derive(Debug, Clone)]
 pub(crate) struct PoolInnerConfig {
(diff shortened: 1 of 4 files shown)


dancoombs (commented):
LGTM! Worth adding a test or two before merging

0xfourzerofour:
> LGTM! Worth adding a test or two before merging

Yeh I can add a couple tests

crates/pool/src/mempool/pool.rs
```diff
@@ -152,9 +153,15 @@ impl PoolInner {
 
     pub(crate) fn mine_operation(
         &mut self,
-        hash: H256,
+        mined_op: &MinedOp,
         block_number: u64,
     ) -> Option<Arc<PoolOperation>> {
+        let tx_in_pool = self.by_id.get(&mined_op.id())?;
```
alex-miao, line 159: Just to make sure I understand the edge cases with replacements, in the case when a replacement happens but the replaced operation gets mined anyways, what happens here is that we end up mining the old operation, we look up in `by_id` the id of the old operation which coincides with the new replacement operation, and we delete the replacement operation here, which is fine since the old operation would have gotten removed in the replacement? Would we need to make the removal step atomic in the edge case that we have not yet updated the operation in the `by_hash` map in the replacement process?

dancoombs (approved):

Issue #453, opened 2023-10-05
[spec] EREP-020/030


dancoombs:
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L237C3-L237C3

> * **[EREP-020]** A staked factory is "accountable" for account breaking the rules. \
That is, if the `validateUserOp()` is rejected for any reason, it is treated as if the factory caused this failure, and thus this affects its reputation.
> * **[EREP-030]** A Staked Account is accountable for failures in other entities (`paymaster`, `aggregator`) even if they are staked. \
(see [https://github.com/eth-infinitism/account-abstraction/issues/274](https://github.com/eth-infinitism/account-abstraction/issues/274))

Currently we apply failures to all entities involved with a UO. We need to do more failure attribution work.

Issue #419, opened 2023-09-25
refactor: simulation and alt-mempools


dancoombs:
**Describe the feature**
Decouple the simulation and alt-mempool matching into two separate classes.

Simulation should return all of the violations.

Alt-mempool matching should attempt to find a mempool that supports all of the violations.

This can be done in two steps.

PR #494, opened 2023-11-12
feat(builder): send entity updates to pool

### LINKED ISSUE #451: [spec] SREP-050

alex-miao:
Closes #451, #455, and #453

## Proposed Changes

  - Changes the error return type of simulation to also include `EntityInfos` to get the staking information of every single entity
  - Also pass through `EntityInfos` in `SimulationSuccess` so that we are able to fetch the stake info of all the entities involved in an op during the first simulation in case the second simulation fails and it is not clear what the entity stake is from the output of the second simulation
  - Have the bundler pass back to the pool all the entity updates, depending on if an entity was staked or not during the simulation to satisfy #451 and #455
  - Do proper validation error attribution to follow the rules defined in #453

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -33,9 +33,13 @@ use rundler_pool::{PoolOperation, PoolServer};
 use rundler_provider::{EntryPoint, HandleOpsOut, Provider};
 use rundler_sim::{
     gas::{self, GasOverheads},
-    ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError, SimulationSuccess, Simulator,
+    EntityInfos, ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError,
+    SimulationSuccess, SimulationViolation, Simulator, ViolationError,
 };
-use rundler_types::{Entity, EntityType, GasFees, Timestamp, UserOperation, UserOpsPerAggregator};
+use rundler_types::{
+    Entity, EntityType, EntityUpdate, EntityUpdateType, GasFees, Timestamp, UserOperation,
+    UserOpsPerAggregator,
+};
 use rundler_utils::{emit::WithEntryPoint, math};
 use tokio::{sync::broadcast, try_join};
 use tracing::{error, info, warn};
@@ -54,7 +58,7 @@ pub(crate) struct Bundle {
     pub(crate) gas_fees: GasFees,
     pub(crate) expected_storage: ExpectedStorage,
     pub(crate) rejected_ops: Vec<UserOperation>,
-    pub(crate) rejected_entities: Vec<Entity>,
+    pub(crate) entity_updates: Vec<EntityUpdate>,
 }
 
 impl Bundle {
@@ -209,14 +213,13 @@ where
                     gas_fees: bundle_fees,
                     expected_storage,
                     rejected_ops: context.rejected_ops,
-                    rejected_entities: context.rejected_entities,
+                    entity_updates: context.entity_updates,
                 });
             }
             info!("Bundle gas estimation failed. Retrying after removing rejected op(s).");
         }
         Ok(Bundle {
             rejected_ops: context.rejected_ops,
-            rejected_entities: context.rejected_entities,
             gas_fees: bundle_fees,
             ..Default::default()
         })
@@ -268,8 +271,8 @@ where
         match result {
             Ok(success) => Ok((op.uo, Ok(success))),
             Err(error) => match error {
-                SimulationError::Violations(_) => Ok((op.uo, Err(error))),
-                SimulationError::Other(error) => Err(error),
+                (ViolationError::Violations(_), _) => Ok((op.uo, Err(error))),
+                (ViolationError::Other(error), _) => Err(error),
             },
         }
     }
@@ -285,6 +288,7 @@ where
             .collect();
         let mut groups_by_aggregator = LinkedHashMap::<Option<Address>, AggregatorGroup>::new();
         let mut rejected_ops = Vec::<UserOperation>::new();
+        let mut entity_updates = Vec::<EntityUpdate>::new();
         let mut paymasters_to_reject = Vec::<Address>::new();
 
         let ov = GasOverheads::default();
@@ -296,8 +300,13 @@ where
                     self.emit(BuilderEvent::rejected_op(
                         self.builder_index,
                         self.op_hash(&op),
-                        OpRejectionReason::FailedRevalidation { error },
+                        OpRejectionReason::FailedRevalidation {
+                            error: error.clone(),
+                        },
                     ));
+                    if let (ViolationError::Violations(violations), entity_infos) = error {
+                        self.add_entity_updates(violations, entity_infos, &mut entity_updates);
+                    }
                     rejected_ops.push(op);
                     continue;
                 }
@@ -378,7 +387,7 @@ where
         let mut context = ProposalContext {
             groups_by_aggregator,
             rejected_ops,
-            rejected_entities: Vec::new(),
+            entity_updates: Vec::new(),
         };
         for paymaster in paymasters_to_reject {
             // No need to update aggregator signatures because we haven't computed them yet.
@@ -740,6 +749,125 @@ where
     fn op_hash(&self, op: &UserOperation) -> H256 {
         op.op_hash(self.entry_point.address(), self.settings.chain_id)
     }
+
+    fn get_entity_update_type(
+        &self,
+        entity: Entity,
+        entity_infos: Option<EntityInfos>,
+    ) -> EntityUpdateType {
+        match entity_infos {
+            Some(infos) => match entity.kind {
+                EntityType::Account => {
+                    if infos.sender.is_staked {
+                        EntityUpdateType::StakedInvalidation
+                    } else {
+                        EntityUpdateType::UnstakedInvalidation
+                    }
+                }
+                EntityType::Paymaster => {
+                    if let Some(paymaster) = infos.paymaster {
+                        if paymaster.is_staked {
+                            EntityUpdateType::StakedInvalidation
+                        } else {
+                            EntityUpdateType::UnstakedInvalidation
+                        }
+                    } else {
+                        EntityUpdateType::StakedInvalidation
+                    }
+                }
+                EntityType::Factory => {
+                    if let Some(factory) = infos.factory {
+                        if factory.is_staked {
+                            EntityUpdateType::StakedInvalidation
+                        } else {
+                            EntityUpdateType::UnstakedInvalidation
+                        }
+                    } else {
+                        EntityUpdateType::StakedInvalidation
+                    }
+                }
+                _ => EntityUpdateType::StakedInvalidation,
+            },
+            None => EntityUpdateType::StakedInvalidation,
+        }
+    }
+
+    fn add_entity_updates(
+        &self,
+        violations: Vec<SimulationViolation>,
+        entity_infos: Option<EntityInfos>,
+        entity_updates: &mut Vec<EntityUpdate>,
+    ) {
+        for violation in violations {
+            match violation {
+                SimulationViolation::UsedForbiddenOpcode(entity, _, _) => {
+                    entity_updates.push(EntityUpdate {
+                        entity,
+                        update_type: self.get_entity_update_type(entity, entity_infos),
+                    })
+                }
+                SimulationViolation::UsedForbiddenPrecompile(entity, _, _) => {
+                    entity_updates.push(EntityUpdate {
+                        entity,
+                        update_type: self.get_entity_update_type(entity, entity_infos),
+                    })
+                }
+                SimulationViolation::AccessedUndeployedContract(entity, _) => {
+                    entity_updates.push(EntityUpdate {
+                        entity,
+                        update_type: self.get_entity_update_type(entity, entity_infos),
+                    })
+                }
+                SimulationViolation::InvalidStorageAccess(entity, _) => {
+                    entity_updates.push(EntityUpdate {
+                        entity,
+                        update_type: self.get_entity_update_type(entity, entity_infos),
+                    })
+                }
+                SimulationViolation::CalledBannedEntryPointMethod(entity) => {
+                    entity_updates.push(EntityUpdate {
+                        entity,
+                        update_type: self.get_entity_update_type(entity, entity_infos),
+                    })
+                }
+                SimulationViolation::CallHadValue(entity) => entity_updates.push(EntityUpdate {
+                    entity,
+                    update_type: self.get_entity_update_type(entity, entity_infos),
+                }),
+                SimulationViolation::NotStaked(entity, _, _) => entity_updates.push(EntityUpdate {
+                    entity,
+                    update_type: self.get_entity_update_type(entity, entity_infos),
+                }),
+                SimulationViolation::UnintendedRevertWithMessage(entity_type, _, address) => {
+                    if let Some(entity_address) = address {
+                        entity_updates.push(EntityUpdate {
+                            entity: Entity {
+                                address: entity_address,
+                                kind: entity_type,
+                            },
+                            update_type: EntityUpdateType::StakedInvalidation,
+                        })
+                    }
+                }
+                SimulationViolation::UnintendedRevert(entity_type, address) => {
+                    if let Some(entity_address) = address {
+                        entity_updates.push(EntityUpdate {
+                            entity: Entity {
+                                address: entity_address,
+                                kind: entity_type,
+                            },
+                            update_type: EntityUpdateType::StakedInvalidation,
+                        })
+                    }
+                }
+                SimulationViolation::OutOfGas(entity) => entity_updates.push(EntityUpdate {
+                    entity,
+                    update_type: EntityUpdateType::StakedInvalidation,
+                }),
+                _ => continue,
+            }
+        }
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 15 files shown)


crates/builder/src/bundle_sender.rs
```diff
@@ -499,9 +497,9 @@ where
             .context("builder should remove rejected ops from pool")
     }
 
-    async fn remove_entities_from_pool(&self, entities: &[Entity]) -> anyhow::Result<()> {
+    async fn update_entities_in_pool(&self, entity_updates: &[EntityUpdate]) -> anyhow::Result<()> {
         self.pool
-            .remove_entities(self.entry_point.address(), entities.to_vec())
+            .update_entities(self.entry_point.address(), entity_updates.to_vec())
             .await
             .context("builder should remove rejected entities from pool")
```
0xfourzerofour, line 504: nit: the context should reflect the method name

crates/builder/src/bundle_proposer.rs
```diff
@@ -53,6 +57,7 @@ pub(crate) struct Bundle {
     pub(crate) expected_storage: ExpectedStorage,
     pub(crate) rejected_ops: Vec<UserOperation>,
     pub(crate) rejected_entities: Vec<Entity>,
+    pub(crate) entity_updates: Vec<EntityUpdate>,
```
dancoombs, line 60: Any reason why these two couldn't be combined?

crates/builder/src/bundle_proposer.rs
```diff
@@ -285,6 +291,7 @@ where
             .collect();
         let mut groups_by_aggregator = LinkedHashMap::<Option<Address>, AggregatorGroup>::new();
         let mut rejected_ops = Vec::<UserOperation>::new();
+        let mut entity_updates = Vec::<EntityUpdate>::new();
         let mut paymasters_to_reject = Vec::<Address>::new();
```
dancoombs, line 295: Same question here, could this be rolled into `entity_updates`?

crates/builder/src/bundle_proposer.rs
```diff
@@ -296,8 +303,97 @@ where
                     self.emit(BuilderEvent::rejected_op(
                         self.builder_index,
                         self.op_hash(&op),
-                        OpRejectionReason::FailedRevalidation { error },
+                        OpRejectionReason::FailedRevalidation {
+                            error: error.clone(),
+                        },
                     ));
+                    if let (ViolationError::Violations(violations), entity_infos) = error {
```
dancoombs, line 310: nit: can we move this into its own function for readability? Having this big match in the middle of the proposal logic is hard to follow.

crates/builder/src/bundle_proposer.rs
```diff
@@ -296,8 +303,97 @@ where
...
+                                    }
+                                }
+                                SimulationViolation::UnintendedRevert(entity_type, address) => {
+                                    if let Some(entity_address) = address {
+                                        entity_updates.push(EntityUpdate {
+                                            entity: Entity {
+                                                address: entity_address,
+                                                kind: entity_type,
+                                            },
+                                            update_type: EntityUpdateType::StakedInvalidation,
```
dancoombs, line 383: Why is an `UnintendedRevert` always a `StakedInvalidation`? Same question with oog below.

crates/builder/src/bundle_proposer.rs
```diff
@@ -601,6 +698,48 @@ where
...
+                        } else {
+                            EntityUpdateType::UnstakedInvalidation
+                        }
+                    } else {
+                        EntityUpdateType::StakedInvalidation
+                    }
+                }
+                _ => EntityUpdateType::StakedInvalidation,
+            },
+            None => EntityUpdateType::StakedInvalidation,
```
dancoombs, line 740: Why is `StakedInvalidation` the default?

crates/builder/src/bundle_proposer.rs
```diff
@@ -713,6 +853,10 @@ impl ProposalContext {
             _ => vec![],
         };
         self.rejected_entities.push(entity);
+        self.entity_updates.push(EntityUpdate {
+            entity,
+            update_type: EntityUpdateType::StakedInvalidation,
```
dancoombs, line 858: Why is this a `StakedInvalidation`

crates/provider/src/ethers/entry_point.rs
```diff
@@ -99,6 +99,18 @@ where
             .context("entry point should return balance")
     }
 
+    async fn get_deposit_info(
```
dancoombs, line 102: Where/how is this used?

dancoombs (commented):
I'm having a pretty hard time tracking the bundle proposer logic. Are there any opportunities to refactor this to make it easier to track? If not, lets at least add some comments explaining what is going on. I like the idea of adding comments with the actual  erc-aa-rules IDs for easy tracking.

crates/builder/src/bundle_proposer.rs
alex-miao: Ah yeah will delete references to `rejected_entities`, the goal is to remove the reject entities endpoint in the pool and any references in the builder

crates/builder/src/bundle_proposer.rs
alex-miao: So in the simulation we actually are unable to get the staking state of an entity at the point of throwing an `UnintendedRevert`. I was thinking we can potentially just make an `eth_call` to fetch the staking staking state right at the start of simulation but risk that we run this at a different block than the actual simulation tracer call.

crates/sim/src/simulation/simulation.rs
```diff
@@ -82,8 +83,15 @@ impl SimulationSuccess {
     }
 }
 
+pub struct ErrorWrapper(Error);
+
 /// The result of a failed simulation
-pub type SimulationError = ViolationError<SimulationViolation>;
+pub type SimulationError = (ViolationError<SimulationViolation>, Option<EntityInfos>);
+impl From<ErrorWrapper> for SimulationError {
```
alex-miao, line 90: Not sure if there is a cleaner way to do this in Rust. Needed to add this `ErrorWrapper` struct because without it and trying to convert from `Error` directly I get `only traits defined in the current crate can be implemented for arbitrary types`

crates/provider/src/ethers/entry_point.rs
alex-miao: Deleted, was left over from when I was still trying to figure out how to get entity staking information before simulation, but now what we do is by adding a field in PoolOperation so that we can pass this staking information in from the first simulation.

crates/pool/src/mempool/mod.rs
```diff
@@ -180,7 +182,7 @@ impl PoolOperation {
     /// For staked accounts, this function will always return true. Staked accounts
     /// are able to circumvent the mempool operation limits always need their reputation
     /// checked to prevent them from filling the pool.
-    pub fn is_staked(&self, entity: EntityType) -> bool {
+    pub fn requires_stake(&self, entity: EntityType) -> bool {
```
alex-miao, line 185: Renaming this to make it more clear that it is only about whether an entity requires stake for an op. It is possible that an entity is staked but does not require stake for this particular op, in which case this returns `false`

crates/builder/src/bundle_proposer.rs
```diff
@@ -207,15 +211,14 @@ where
                     gas_estimate,
                     gas_fees: bundle_fees,
                     expected_storage,
-                    rejected_ops: context.rejected_ops,
-                    rejected_entities: context.rejected_entities,
+                    rejected_ops: context.rejected_ops.iter().map(|po| po.0.clone()).collect(),
+                    entity_updates: context.entity_updates.into_values().collect(),
```
alex-miao, line 215: context.entity_updates is a `BTreeMap` to make sure that this result is deterministic (otherwise it can break the `test_reject_entities` test)

crates/builder/src/bundle_proposer.rs
```diff
@@ -661,11 +852,11 @@ impl ProposalContext {
         }
     }
 
-    fn get_op_at(&self, index: usize) -> anyhow::Result<&UserOperation> {
+    fn get_op_at(&self, index: usize) -> anyhow::Result<&OpWithSimulation> {
```
alex-miao, line 855: Changing this to return `OpWithSimulation` instead so that we are able to get the `EntityInfos` from `SimulationSuccess` to help with doing entity updates for the `factory` and `paymaster` in `process_failed_op`

crates/builder/src/bundle_proposer.rs
```diff
@@ -465,7 +478,7 @@ where
             }
             HandleOpsOut::SignatureValidationFailed(aggregator) => {
                 info!("Rejected aggregator {aggregator:?} because its signature validation failed during gas estimation.");
-                self.reject_entity(context, Entity::aggregator(aggregator))
+                self.reject_entity(context, Entity::aggregator(aggregator), false)
```
alex-miao, line 481: Unfortunately is fairly difficult to get the staking state of the aggregator at this point in the code so defaulting to `false` for now.

crates/builder/src/bundle_proposer.rs
```diff
@@ -207,15 +211,14 @@ where
...
-                    rejected_entities: context.rejected_entities,
+                    rejected_ops: context.rejected_ops.iter().map(|po| po.0.clone()).collect(),
+                    entity_updates: context.entity_updates.into_values().collect(),
                 });
             }
             info!("Bundle gas estimation failed. Retrying after removing rejected op(s).");
         }
         Ok(Bundle {
-            rejected_ops: context.rejected_ops,
-            rejected_entities: context.rejected_entities,
```
dancoombs, line 218: Why are there no `entity_updates` in this bundle?

crates/builder/src/bundle_proposer.rs
```diff
@@ -631,8 +822,8 @@ impl OpWithSimulation {
 #[derive(Debug)]
 struct ProposalContext {
     groups_by_aggregator: LinkedHashMap<Option<Address>, AggregatorGroup>,
-    rejected_ops: Vec<UserOperation>,
-    rejected_entities: Vec<Entity>,
+    rejected_ops: Vec<(UserOperation, EntityInfos)>,
+    entity_updates: BTreeMap<EntityType, EntityUpdate>,
```
dancoombs, line 826: What if there are 2 entities of the same type that have updates? I.e. 2 different paymasters that need to be banned in the same bundle? I'd expect this map to be keyed by address.

crates/builder/src/bundle_proposer.rs
```diff
@@ -606,6 +636,167 @@ where
     fn op_hash(&self, op: &UserOperation) -> H256 {
         op.op_hash(self.entry_point.address(), self.settings.chain_id)
     }
+
+    /// For a given entity that has caused a revert during simulation, determine if the entity was staked or not.
+    /// If it is staked, we push a StakedInvalidation entity update to enforce [SREP-050]
+    /// If it is not staked, we push an UnstakedInvalidation entity update to enforce [UREP-030]
```
dancoombs, line 642: nit: Use `//` since this is a private function

crates/builder/src/bundle_proposer.rs
```diff
@@ -606,6 +636,167 @@ where
...
+    fn get_entity_update_type(
+        &self,
+        entity_type: EntityType,
+        entity_infos: EntityInfos,
+    ) -> EntityUpdateType {
```
dancoombs, line 647: Does this use `&self`?

crates/builder/src/bundle_proposer.rs
```diff
@@ -606,6 +636,167 @@ where
...
+                    self.add_entity_update(entity, entity_infos, entity_updates)
+                }
+                _ => continue,
+            }
+        }
+    }
+
+    // Add an entity update for the entity
+    fn add_entity_update(
+        &self,
```
dancoombs, line 789: Same question about `ProposalContext` here

crates/builder/src/bundle_proposer.rs
```diff
@@ -707,7 +899,7 @@ impl ProposalContext {
     /// Returns the addresses of any aggregators whose signature may need to be
     /// recomputed.
     #[must_use = "rejected entity but did not update aggregator signatures"]
-    fn reject_entity(&mut self, entity: Entity) -> Vec<Address> {
+    fn reject_entity(&mut self, entity: Entity, is_staked: bool) -> Vec<Address> {
```
dancoombs, line 902: When should this be used vs `add_entity_updates`? Can we consolidate this logic?

crates/builder/src/bundle_proposer.rs
dancoombs: Can you add a ticket to track this?

crates/sim/src/simulation/simulation.rs
```diff
@@ -622,17 +673,26 @@ struct ValidationContext {
...
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+/// additional context for all the entities used in an op
+pub struct EntityInfos {
+    /// The entity info for the factory
+    pub factory: Option<EntityInfo>,
+    /// The entity info for the op sender
+    pub sender: EntityInfo,
+    /// The entity info for the paymaster
+    pub paymaster: Option<EntityInfo>,
+    /// The entity infor for the aggregator
```
dancoombs, line 694: nit: `infor` -> `info`

crates/builder/src/bundle_proposer.rs
dancoombs: I'd add this comment to `context.entity_updates` (if its not there already)

crates/builder/src/bundle_proposer.rs
alex-miao: created here: https://github.com/alchemyplatform/rundler/issues/515

crates/builder/src/bundle_proposer.rs
alex-miao: I am using `reject_entity` for any entity rejections that happen in bundle proposer as opposed to those that happen in simulation and get passed to the bundle proposer. For `add_entity_updates` we don't need to be making the `reject_aggregator`/`reject_paymaster`/`reject_factory` calls that reject_entity does.

crates/builder/src/bundle_proposer.rs
alex-miao: added

crates/builder/src/bundle_proposer.rs
alex-miao: Ah yeah messed this up, fixed now

crates/builder/src/bundle_proposer.rs
alex-miao: no, turned this into an associated function

crates/sim/src/simulation/simulation.rs
```diff
@@ -82,8 +85,17 @@ impl SimulationSuccess {
     }
 }
 
-/// The result of a failed simulation
-pub type SimulationError = ViolationError<SimulationViolation>;
+struct ErrorWrapper(Error);
+
+/// The result of a failed simulation. We return a list of the violations that ocurred during the failed simulation
+/// and also information about all the entities used in the op to handle entity penalties
+pub type SimulationError = (ViolationError<SimulationViolation>, Option<EntityInfos>);
```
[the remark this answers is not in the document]
alex-miao: changed to a struct

crates/sim/src/simulation/simulation.rs
alex-miao: Wasn't even working yesterday but somehow it just started working today, not sure what was causing this earlier.

crates/builder/src/bundle_proposer.rs
dancoombs: Got it. Maybe we rename `add_entity_updates` to be more descriptive of what it does. Maybe `process_violations` or something?

crates/builder/src/bundle_proposer.rs
```diff
@@ -260,45 +264,64 @@ where
...
                     ));
-                    rejected_ops.push(op);
+                    if let SimulationError {
+                        violation_error: ViolationError::Violations(violations),
+                        entity_infos,
+                    } = error
+                    {
+                        // try to use EntityInfos from the latest simulation, but if it doesn't exist use the EntityInfos from the previous simulation
+                        let infos = entity_infos.map_or(po.entity_infos, |e| e);
+                        ProposalContext::add_entity_updates(violations, infos, &mut entity_updates);
```
dancoombs, line 322: nit: Any reason why this is a static method? Could we create a proposal context above and update it here?

crates/sim/src/simulation/simulation.rs
```diff
@@ -171,28 +190,55 @@ where
         // mean the entry point is fine if one of the phases fails and it
         // doesn't reach the end of execution.
         if num_phases > 3 {
-            Err(vec![SimulationViolation::WrongNumberOfPhases(num_phases)])?
+            Err(SimulationError {
```
dancoombs, line 193: Not for this PR.

I think we should refactor this code such that the error type is only for handling errors in simulation (i.e. the trace call) and the result type contains the context as well as violations. We then remove the mempool matching logic from this function and place that at the caller. We then pass along any non-matched violations at that point. Trying to jam all this info into the error type smells.

Linking: https://github.com/alchemyplatform/rundler/issues/419 for tracking.

crates/builder/src/bundle_proposer.rs
alex-miao: Accidentally made this change in #519 instead of here, gonna leave it there

crates/sim/src/simulation/simulation.rs
alex-miao: Yeah agree that this current error structure isn't ideal, sounds good

PR #503, opened 2023-11-16
feat(compliance): add compliance for STO-040/041

### LINKED ISSUE #448: [spec] STO-040/041
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L214

> * **[STO-040]** `UserOperation` may not use an entity address (`factory`/`paymaster`/`aggregator`) that is used as an "account" in another `UserOperation` in the mempool. \
This means that a `Paymaster` and `Factory` contracts cannot practically be an "account" contract as well.
> * **[STO-041]** A contract whose Associated Storage slot is accessed during a `UserOperation` may not be an address of a "sender" in another `UserOperation` in the mempool.

0xfourzerofour:
Closes #448 

## Proposed Changes

  - Add compliance for sto-041 and sto-040

## Discussion

I added a simple  struct to be used within the count_by_address hashmap so that we can keep track of the different entity types within the mempool with the same address.

https://github.com/eth-infinitism/bundler/blob/main/packages/bundler/src/modules/MempoolManager.ts#L146C11-L146C38

Reference bundler logic


crates/pool/src/mempool/pool.rs
```diff
@@ -151,6 +156,29 @@ impl PoolInner {
...
+    pub(crate) fn check_multiple_roles_violation(&self, uo: &UserOperation) -> MempoolResult<()> {
+        if let Some(ec) = self.count_by_address.get(&uo.sender) {
+            if ec.includes_non_sender() {
+                return Err(MempoolError::SenderAddressUsedAsAlternateEntity(uo.sender));
+            }
+        }
+
+        for e in uo.entities() {
+            match e.kind {
+                EntityType::Factory | EntityType::Paymaster => {
```
alex-miao, line 168: Should we add `EntityType::Aggregator` here

crates/pool/src/mempool/error.rs
```diff
@@ -36,6 +36,12 @@ pub enum MempoolError {
     /// Max operations reached for this sender
     #[error("Max operations ({0}) reached for sender {1}")]
     MaxOperationsReached(usize, Address),
+    /// Multiple roles violation
+    #[error("A {} at {} in this UserOperation is used as a sender entity in another UserOperation currently in mempool.", .0.kind, .0.address)]
+    MultipleRolesViolation(Entity),
+    /// Sender address used as different entity in another UserOperation currently in the mempool.
+    #[error("The sender address {0} is used as a different entity in another UserOperation currently in mempool")]
+    SenderAddressUsedAsAlternateEntity(Address),
```
alex-miao, line 44: More of an organizational question but would it make sense to try to have a single error per spec rule or allow multiple errors per spec rule to be more detailed

crates/pool/src/mempool/pool.rs
```diff
@@ -256,8 +284,7 @@ impl PoolInner {
         }
 
         // Check sender count in mempool. If sender has too many operations, must be staked
-        if *self.count_by_address.get(&op.uo.sender).unwrap_or(&0)
-            >= self.config.max_userops_per_sender
+        if self.address_count(&op.uo.sender) >= self.config.max_userops_per_sender
```
alex-miao, line 287: Not related to this PR but I see that in the new spec this config variable is now named `SAME_SENDER_MEMPOOL_COUNT`, might be worth an update

crates/pool/src/mempool/error.rs
0xfourzerofour: Yeh this brings up a good point, I think what we could do is have a single RPC error in spec rule and then at the crate level we could be more specific.

@dancoombs what do you think?

0xfourzerofour: Seems like we are going to keep it for now after asking @dancoombs

PR #508, opened 2023-11-21
fix(builder): handle the post op revert 0.6 entry point bug

dancoombs:
Handle the bug in the 0.6 entry point that was fixed here https://github.com/eth-infinitism/account-abstraction/pull/325

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -14,7 +14,9 @@
 use std::{
     cmp,
     collections::{HashMap, HashSet},
+    future::Future,
     mem,
+    pin::Pin,
     sync::Arc,
     time::Duration,
 };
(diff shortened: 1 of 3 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -559,6 +568,133 @@ where
...
+                    message: Arc::new("post op reverted leading to entry point revert".to_owned()),
+                },
+            ));
+            self.reject_index(context, index).await;
+        }
+
+        Ok(())
+    }
+
+    async fn check_for_post_op_revert_single_op(
```
0xfourzerofour, line 629: Why return vec of op index if its either going to be empty or return on item? 

Could we just use an option of usize?

0xfourzerofour: Oh actually, its for the futures to be the same datatype. dont mind me

0xfourzerofour (approved):

Issue #498, opened 2023-11-13
[pool] Reputation for replacement operations


dancoombs:
**Describe the feature**
When a UO is replaced in the mempool we should update the reputation tracker and reduce the `OPS_SEEN` count for each entity involved in the replaced op by 1.

PR #516, opened 2023-11-28
feat(compliance): add spec compliance for EREP-10 and STO-040/041

### LINKED ISSUE #452: [spec] EREP-010
**Describe the feature**
https://github.com/eth-infinitism/account-abstraction/blob/d13ab308091c53acd8af7eda52babb0f8d301a01/eip/EIPS/eip-aa-rules.md?plain=1#L234

> * **[EREP-010]** For each `paymaster`, the mempool must maintain the total gas `UserOperations` using this `paymaster` may consume.
    * Do not add a `UserOperation` to the mempool if the maximum total gas usage, including the new `UserOperation`, is above the deposit of the `paymaster` at the current gas price.
    
  We currently only track this during bundling, not while gating the mempool.

0xfourzerofour:
Closes #452 #448 #480 

## Proposed Changes

  - add spec compliance by creating structure to keep track of paymaster balances
  - add compliance for STO-040/041 
  - add compliance for EREP-10
  - Change error codes to return opcode error instead of unstaked error in compliance with the 4337 team

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,7 @@ test-spec-integrated: ## Run spec tests in integrated mode
 test-spec-modular: ## Run spec tests in modular mode
 	test/spec-tests/remote/run-spec-tests.sh
 	
+.PHONY: fmt
+fmt: ## format code with nightly rust
+	cargo +nightly fmt
+
--- a/crates/pool/proto/op_pool/op_pool.proto
+++ b/crates/pool/proto/op_pool/op_pool.proto
@@ -575,10 +575,11 @@ message InvalidStorageAccess {
 
 message NotStaked {
   Entity entity = 1;
-  string accessed = 2;
-  bytes slot = 3;
-  bytes min_stake = 4;
-  bytes min_unstake_delay = 5;
+  bytes accessed_address = 2;
+  EntityType accessed_entity = 3;
+  bytes slot = 4;
+  bytes min_stake = 5;
+  bytes min_unstake_delay = 6;
 }
 
 message UnintendedRevert {
--- a/crates/pool/proto/op_pool/op_pool.proto
+++ b/crates/pool/proto/op_pool/op_pool.proto
@@ -382,6 +382,10 @@ message MempoolError {
     UnsupportedAggregatorError unsupported_aggregator = 9;
     InvalidSignatureError invalid_signature = 10;
     UnknownEntryPointError unknown_entry_point = 11;
+    MultipleRolesViolation multiple_roles_violation = 12;
+    SenderAddressUsedAsAlternateEntity sender_address_used_as_alternate_entity = 13;
+    AssociatedStorageIsAlternateSender associated_storage_is_alternate_sender = 14;
+    PaymasterBalanceTooLow paymaster_balance_too_low = 15;
   }
 }
 
@@ -389,11 +393,18 @@ message UnknownEntryPointError {
   bytes entry_point = 1;
 }
 
+message AssociatedStorageIsAlternateSender {}
+
 message OperationAlreadyKnownError {}
 
 message ReplacementUnderpricedError {
   bytes current_fee = 1;
   bytes current_priority_fee = 2;
+}
+
+message PaymasterBalanceTooLow {
+  bytes current_balance = 1;
+  bytes required_balance = 2;
 }
 
 message MaxOperationsReachedError {
@@ -403,6 +414,14 @@ message MaxOperationsReachedError {
 
 message EntityThrottledError {
   Entity entity = 1;
+}
+
+message MultipleRolesViolation {
+  Entity entity = 1;
+}
+
+message SenderAddressUsedAsAlternateEntity {
+  bytes sender_address = 1;
 }
 
 message DiscardedOnInsertError {}
@@ -516,10 +535,16 @@ message SimulationViolationError {
     CalledBannedEntryPointMethod called_banned_entry_point_method = 14;
     CodeHashChanged code_hash_changed = 15;
     AggregatorValidationFailed aggregator_validation_failed = 16;
+    UnstakedPaymasterContext unstaked_paymaster_context = 17;
+    UnstakedAggregator unstaked_aggregator = 18;
   }
 }
 
 message InvalidSignature {}
+
+message UnstakedAggregator {}
+
+message UnstakedPaymasterContext {}
 
 message UnintendedRevertWithMessage {
   Entity entity = 1;
@@ -550,8 +575,10 @@ message InvalidStorageAccess {
 
 message NotStaked {
   Entity entity = 1;
-  bytes min_stake = 2;
-  bytes min_unstake_delay = 3;
+  string accessed = 2;
+  bytes slot = 3;
+  bytes min_stake = 4;
+  bytes min_unstake_delay = 5;
 }
 
 message UnintendedRevert {
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -21,13 +21,14 @@ use anyhow::{ensure, Context};
 use ethers::{
     contract,
     prelude::EthEvent,
-    types::{Address, Block, Filter, H256, U256},
+    types::{Address, Block, Filter, Log, H256, U256},
 };
 use futures::future;
 use rundler_provider::Provider;
 use rundler_task::block_watcher;
 use rundler_types::{
-    contracts::i_entry_point::UserOperationEventFilter, Timestamp, UserOperationId,
+    contracts::{entry_point::DepositedFilter, i_entry_point::UserOperationEventFilter},
+    Timestamp, UserOperationId,
 };
 use tokio::{
     select,
@@ -66,6 +67,9 @@ pub struct ChainUpdate {
     pub reorg_depth: u64,
     pub mined_ops: Vec<MinedOp>,
     pub unmined_ops: Vec<MinedOp>,
+    pub entity_deposits: Vec<DepositInfo>,
+    pub unmined_entity_deposits: Vec<DepositInfo>,
+    pub reorg_larger_than_history: bool,
 }
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
@@ -74,6 +78,15 @@ pub struct MinedOp {
     pub entry_point: Address,
     pub sender: Address,
     pub nonce: U256,
+    pub actual_gas_cost: U256,
+    pub paymaster: Option<Address>,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct DepositInfo {
+    pub address: Address,
+    pub entrypoint: Address,
+    pub amount: U256,
 }
 
 impl MinedOp {
@@ -99,6 +112,7 @@ struct BlockSummary {
     timestamp: Timestamp,
     parent_hash: H256,
     ops: Vec<MinedOp>,
+    entity_deposits: Vec<DepositInfo>,
 }
 
 impl<P: Provider> Chain<P> {
@@ -171,6 +185,11 @@ impl<P: Provider> Chain<P> {
             current_block_number < new_block_number + self.settings.history_size,
             "new block number {new_block_number} should be greater than start of history (current block: {current_block_number})"
         );
+
+        if new_block_number <= current_block_number.saturating_sub(self.settings.history_size) {
+            return self.reset_and_initialize(new_head).await;
+        }
+
         if current_block_number + self.settings.history_size < new_block_number {
             warn!(
                 "New block {new_block_number} number is {} blocks ahead of the previously known head. Chain history will skip ahead.",
@@ -178,6 +197,7 @@ impl<P: Provider> Chain<P> {
             );
             return self.reset_and_initialize(new_head).await;
         }
+
         let added_blocks = self
             .load_added_blocks_connecting_to_existing_chain(current_block_number, new_head)
             .await?;
@@ -198,7 +218,14 @@ impl<P: Provider> Chain<P> {
             .flat_map(|block| &block.ops)
             .copied()
             .collect();
-        Ok(self.new_update(0, mined_ops, vec![]))
+
+        let entity_deposits: Vec<_> = self
+            .blocks
+            .iter()
+            .flat_map(|block| &block.entity_deposits)
+            .copied()
+            .collect();
+        Ok(self.new_update(0, mined_ops, vec![], entity_deposits, vec![], false))
     }
 
     /// Given a collection of blocks to add to the chain, whose numbers may
@@ -214,6 +241,12 @@ impl<P: Provider> Chain<P> {
             .flat_map(|block| &block.ops)
             .copied()
             .collect();
+
+        let entity_deposits: Vec<_> = added_blocks
+            .iter()
+            .flat_map(|block| &block.entity_deposits)
+            .copied()
+            .collect();
         let reorg_depth = current_block_number + 1 - added_blocks[0].number;
         let unmined_ops: Vec<_> = self
             .blocks
@@ -222,6 +255,17 @@ impl<P: Provider> Chain<P> {
             .flat_map(|block| &block.ops)
             .copied()
             .collect();
+
+        let unmined_entity_deposits: Vec<_> = self
+            .blocks
+            .iter()
+            .skip(self.blocks.len() - reorg_depth as usize)
+            .flat_map(|block| &block.entity_deposits)
+            .copied()
+            .collect();
+
+        let is_reorg_larger_than_history = reorg_depth >= self.settings.history_size;
+
         for _ in 0..reorg_depth {
             self.blocks.pop_back();
         }
@@ -236,7 +280,14 @@ impl<P: Provider> Chain<P> {
             ChainMetrics::increment_total_reorg_depth(reorg_depth);
         }
 
-        self.new_update(reorg_depth, mined_ops, unmined_ops)
+        self.new_update(
+            reorg_depth,
+            mined_ops,
+            unmined_ops,
+            entity_deposits,
+            unmined_entity_deposits,
+            is_reorg_larger_than_history,
+        )
     }
 
     async fn load_added_blocks_connecting_to_existing_chain(
@@ -328,39 +379,86 @@ impl<P: Provider> Chain<P> {
         let opses = future::try_join_all(future_opses)
             .await
             .context("should load ops for new blocks")?;
-        for (i, ops) in opses.into_iter().enumerate() {
+        for (i, (ops, deposits)) in opses.into_iter().enumerate() {
             blocks[i].ops = ops;
+            blocks[i].entity_deposits = deposits;
         }
         Ok(())
     }
 
-    async fn load_ops_in_block_with_hash(&self, block_hash: H256) -> anyhow::Result<Vec<MinedOp>> {
+    async fn load_ops_in_block_with_hash(
+        &self,
+        block_hash: H256,
+    ) -> anyhow::Result<(Vec<MinedOp>, Vec<DepositInfo>)> {
         let _permit = self
             .load_ops_semaphore
             .acquire()
             .await
             .expect("semaphore should not be closed");
+
+        let deposit = DepositedFilter::abi_signature();
+        let uo_filter = UserOperationEventFilter::abi_signature();
+        let events: Vec<&str> = vec![&deposit, &uo_filter];
+
         let filter = Filter::new()
             .address(self.settings.entry_point_addresses.clone())
-            .event(&UserOperationEventFilter::abi_signature())
+            .events(events)
             .at_block_hash(block_hash);
         let logs = self
             .provider
             .get_logs(&filter)
             .await
             .context("chain state should load user operation events")?;
-        logs.into_iter()
-            .map(|log| {
-                let entry_point = log.address;
-                let event = contract::parse_log::<UserOperationEventFilter>(log)?;
-                Ok(MinedOp {
+
+        let deposits = self.load_entity_deposits(&logs);
+        let mined_ops = self.load_mined_ops(&logs);
+
+        Ok((mined_ops, deposits))
+    }
+
+    fn load_mined_ops(&self, logs: &Vec<Log>) -> Vec<MinedOp> {
+        let mut mined_ops = vec![];
+        for log in logs {
+            let entry_point = log.address;
+            if let Ok(event) = contract::parse_log::<UserOperationEventFilter>(log.clone()) {
+                let paymaster = if event.paymaster.is_zero() {
+                    None
+                } else {
+                    Some(event.paymaster)
+                };
+
+                let mined = MinedOp {
                     hash: event.user_op_hash.into(),
                     entry_point,
                     sender: event.sender,
                     nonce: event.nonce,
-                })
-            })
-            .collect()
+                    actual_gas_cost: event.actual_gas_cost,
+                    paymaster,
+                };
+
+                mined_ops.push(mined);
+            }
+        }
+
+        mined_ops
+    }
+
+    fn load_entity_deposits(&self, logs: &Vec<Log>) -> Vec<DepositInfo> {
+        let mut deposits = vec![];
+        for log in logs {
+            let entrypoint = log.address;
+            if let Ok(event) = contract::parse_log::<DepositedFilter>(log.clone()) {
+                let info = DepositInfo {
+                    entrypoint,
+                    address: event.account,
+                    amount: event.total_deposit,
+                };
+
+                deposits.push(info);
+            }
+        }
+
+        deposits
     }
 
     fn block_with_number(&self, number: u64) -> Option<&BlockSummary> {
@@ -376,6 +474,9 @@ impl<P: Provider> Chain<P> {
         reorg_depth: u64,
         mined_ops: Vec<MinedOp>,
         unmined_ops: Vec<MinedOp>,
+        entity_deposits: Vec<DepositInfo>,
+        unmined_entity_deposits: Vec<DepositInfo>,
+        reorg_larger_than_history: bool,
     ) -> ChainUpdate {
         let latest_block = self
             .blocks
@@ -389,6 +490,9 @@ impl<P: Provider> Chain<P> {
             reorg_depth,
             mined_ops,
             unmined_ops,
+            entity_deposits,
+            unmined_entity_deposits,
+            reorg_larger_than_history,
         }
     }
 }
@@ -423,6 +527,7 @@ impl BlockSummary {
             timestamp: block.timestamp.as_u64().into(),
             parent_hash: block.parent_hash,
             ops: Vec::new(),
+            entity_deposits: Vec::new(),
         })
     }
 }
@@ -479,6 +584,7 @@ mod tests {
 
     use ethers::{
         abi::AbiEncode,
+        prelude::EthEvent,
         types::{FilterBlockOption, Log, H160},
         utils,
     };
@@ -494,11 +600,16 @@ mod tests {
     struct MockBlock {
         hash: H256,
         op_hashes: Vec<H256>,
+        deposit_addresses: Vec<Address>,
     }
 
     impl MockBlock {
-        fn new(hash: H256, op_hashes: Vec<H256>) -> Self {
-            Self { hash, op_hashes }
+        fn new(hash: H256, op_hashes: Vec<H256>, deposit_addresses: Vec<Address>) -> Self {
+            Self {
+                hash,
+                op_hashes,
+                deposit_addresses,
+            }
         }
     }
 
@@ -543,7 +654,18 @@ mod tests {
             let Some(block) = block else {
                 return vec![];
             };
-            block.op_hashes.iter().copied().map(fake_log).collect()
+
+            let mut joined_logs: Vec<Log> = Vec::new();
+            joined_logs.extend(block.op_hashes.iter().copied().map(fake_log));
+            joined_logs.extend(
+                block
+                    .deposit_addresses
+                    .iter()
+                    .copied()
+                    .map(fake_deposit_log),
+            );
+
+            joined_logs
         }
     }
 
@@ -551,10 +673,10 @@ mod tests {
     async fn test_initial_load() {
         let (mut chain, controller) = new_chain();
         controller.set_blocks(vec![
-            MockBlock::new(hash(0), vec![hash(101), hash(102)]),
-            MockBlock::new(hash(1), vec![hash(103)]),
-            MockBlock::new(hash(2), vec![]),
-            MockBlock::new(hash(3), vec![hash(104), hash(105)]),
+            MockBlock::new(hash(0), vec![hash(101), hash(102)], vec![]),
+            MockBlock::new(hash(1), vec![hash(103)], vec![]),
+            MockBlock::new(hash(2), vec![], vec![]),
+            MockBlock::new(hash(3), vec![hash(104), hash(105)], vec![]),
         ]);
         let update = chain.sync_to_block(controller.get_head()).await.unwrap();
         // With a history size of 3, we should get updates from all blocks except the first one.
@@ -568,6 +690,9 @@ mod tests {
                 reorg_depth: 0,
                 mined_ops: vec![fake_mined_op(103), fake_mined_op(104), fake_mined_op(105),],
                 unmined_ops: vec![],
+                entity_deposits: vec![],
+                unmined_entity_deposits: vec![],
+                reorg_larger_than_history: false,
             }
         );
     }
@@ -576,15 +701,15 @@ mod tests {
     async fn test_simple_advance() {
         let (mut chain, controller) = new_chain();
         controller.set_blocks(vec![
-            MockBlock::new(hash(0), vec![hash(101), hash(102)]),
-            MockBlock::new(hash(1), vec![hash(103)]),
-            MockBlock::new(hash(2), vec![]),
-            MockBlock::new(hash(3), vec![hash(104), hash(105)]),
+            MockBlock::new(hash(0), vec![hash(101), hash(102)], vec![]),
+            MockBlock::new(hash(1), vec![hash(103)], vec![]),
+            MockBlock::new(hash(2), vec![], vec![]),
+            MockBlock::new(hash(3), vec![hash(104), hash(105)], vec![]),
         ]);
         chain.sync_to_block(controller.get_head()).await.unwrap();
         controller
             .get_blocks_mut()
-            .push(MockBlock::new(hash(4), vec![hash(106)]));
+            .push(MockBlock::new(hash(4), vec![hash(106)], vec![]));
         let update = chain.sync_to_block(controller.get_head()).await.unwrap();
         assert_eq!(
             update,
@@ -596,6 +721,9 @@ mod tests {
                 reorg_depth: 0,
                 mined_ops: vec![fake_mined_op(106)],
                 unmined_ops: vec![],
+                entity_deposits: vec![],
+                unmined_entity_deposits: vec![],
+                reorg_larger_than_history: false,
             }
         );
     }
@@ -604,9 +732,9 @@ mod tests {
     async fn test_forward_reorg() {
         let (mut chain, controller) = new_chain();
         controller.set_blocks(vec![
-            MockBlock::new(hash(0), vec![hash(100)]),
-            MockBlock::new(hash(1), vec![hash(101)]),
-            MockBlock::new(hash(2), vec![hash(102)]),
+            MockBlock::new(hash(0), vec![hash(100)], vec![]),
+            MockBlock::new(hash(1), vec![hash(101)], vec![]),
+            MockBlock::new(hash(2), vec![hash(102)], vec![Address::zero()]),
         ]);
         chain.sync_to_block(controller.get_head()).await.unwrap();
         {
@@ -614,9 +742,9 @@ mod tests {
             let mut blocks = controller.get_blocks_mut();
             blocks.pop();
             blocks.extend([
-                MockBlock::new(hash(12), vec![hash(112)]),
-                MockBlock::new(hash(13), vec![hash(113)]),
-                MockBlock::new(hash(14), vec![hash(114)]),
+                MockBlock::new(hash(12), vec![hash(112)], vec![]),
+                MockBlock::new(hash(13), vec![hash(113)], vec![]),
+                MockBlock::new(hash(14), vec![hash(114)], vec![]),
             ]);
         }
         let update = chain.sync_to_block(controller.get_head()).await.unwrap();
@@ -630,6 +758,9 @@ mod tests {
                 reorg_depth: 1,
                 mined_ops: vec![fake_mined_op(112), fake_mined_op(113), fake_mined_op(114)],
                 unmined_ops: vec![fake_mined_op(102)],
+                entity_deposits: vec![],
+                unmined_entity_deposits: vec![fake_mined_deposit(Address::zero(), 0.into())],
+                reorg_larger_than_history: false,
             }
         );
     }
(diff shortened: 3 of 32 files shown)


Makefile
```diff
@@ -29,3 +29,7 @@ test-spec-integrated: ## Run spec tests in integrated mode
 test-spec-modular: ## Run spec tests in modular mode
 	test/spec-tests/remote/run-spec-tests.sh
 	
+.PHONY: format
+format: ## format code with nightly rust
```
dancoombs, line 33: nit: maybe call this `fmt` just since thats what most tools use. So we'd call `make fmt`

crates/pool/src/chain.rs
```diff
@@ -95,6 +108,7 @@ struct BlockSummary {
     hash: H256,
     parent_hash: H256,
     ops: Vec<MinedOp>,
+    entity_deposits: Vec<DepositInfo>,
```
dancoombs, line 111: Is there logic to handle reorgs here?

I.e. what happens if a deposit is reorged away?

crates/pool/src/mempool/error.rs
```diff
@@ -36,12 +36,24 @@ pub enum MempoolError {
     /// Max operations reached for this sender
     #[error("Max operations ({0}) reached for sender {1}")]
     MaxOperationsReached(usize, Address),
+    /// Multiple roles violation
+    #[error("A {} at {} in this UserOperation is used as a sender entity in another UserOperation currently in mempool.", .0.kind, .0.address)]
+    MultipleRolesViolation(Entity),
+    /// An Associated storage slot that is accessed in the UserOperation is being used as a sender by another UserOperation in the mempool.
+    #[error("An Associated storage slot that is accessed in the UserOperation is being used as a sender by another UserOperation in the mempool")]
```
dancoombs, line 43: nit: remove capital A in Associated

dancoombs, line 43: I'm not sure I understand this, how is a storage slot used as a sender?

crates/pool/src/mempool/mod.rs
```diff
@@ -66,6 +70,8 @@ pub trait Mempool: Send + Sync + 'static {
     /// Updates the reputation of an entity.
     fn update_entity(&self, entity_update: EntityUpdate);
 
+    async fn paymaster_balance(&self, paymaster: Address) -> ProviderResult<PaymasterMetadata>;
```
dancoombs, line 73: Add doc comment

crates/pool/src/mempool/error.rs
0xfourzerofour: It mewans that the the storage slot address is being used as a sender in another user operation

crates/pool/src/chain.rs
0xfourzerofour: Yeh that is a good question, let me think about reorgs

crates/pool/src/chain.rs
```diff
@@ -63,6 +66,8 @@ pub struct ChainUpdate {
     pub reorg_depth: u64,
     pub mined_ops: Vec<MinedOp>,
     pub unmined_ops: Vec<MinedOp>,
+    pub entity_deposits: Vec<DepositInfo>,
+    pub entity_deposit_reverts: Vec<DepositInfo>,
```
dancoombs, line 70: Maybe `unmined_entity_deposits?` Reverts makes me think of a revert on chain, this is a reorg.

crates/pool/src/chain.rs
```diff
@@ -836,6 +948,8 @@ mod tests {
             entry_point: ENTRY_POINT_ADDRESS,
             sender: Address::zero(),
             nonce: U256::zero(),
+            actual_gas_cost: U256::zero(),
+            paymaster: None,
```
dancoombs, line 952: Should add some tests for entity_deposits and reorgs

crates/pool/src/mempool/entity_tracker.rs
```diff
@@ -0,0 +1,56 @@
...
+//
+// You should have received a copy of the GNU General Public License along with Rundler.
+// If not, see https://www.gnu.org/licenses/.
+
+use rundler_types::EntityType;
+
+/// Object to track count and type of entity in mempool
+
+#[derive(Debug, Default)]
+pub(crate) struct EntityCountTracker {
```
dancoombs, line 19: nit: maybe just `EntityCounter`?

crates/pool/src/mempool/error.rs
```diff
@@ -38,12 +38,24 @@ pub enum MempoolError {
     /// Max operations reached for this sender
     #[error("Max operations ({0}) reached for sender {1}")]
     MaxOperationsReached(usize, Address),
+    /// Multiple roles violation
```
dancoombs, line 41: Could be useful to start adding the spec rule alongside these errors. So like EREP-20 or whatever. Might make sense just to label these and then we can to and label the old ones.

crates/pool/src/mempool/pool.rs
```diff
@@ -208,6 +258,33 @@ impl PoolInner {
...
+    }
+
+    pub(crate) fn paymaster_exists(&self, paymaster: Address) -> bool {
+        self.paymaster_balances.paymaster_exists(paymaster)
+    }
+
+    pub(crate) fn update_paymaster_balance_after_mine(&mut self, mined_op: &MinedOp) {
+        self.paymaster_balances
+            .update_paymaster_balance_from_mined_op(mined_op)
+    }
```
dancoombs, line 272: Do we update the paymaster balance when an op is unmined?

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -124,6 +155,9 @@ where
                 // Remove throttled ops that were included in the block
                 state.throttled_ops.remove(&op.hash);
 
+                // update paymaster balances
+                state.pool.update_paymaster_balance_after_mine(op);
```
dancoombs, line 159: Looks like we need to do the opposite when an op is unmined.

0xfourzerofour: Yeh did not think about that I can add that in

crates/pool/src/mempool/mod.rs
```diff
@@ -160,10 +167,19 @@ pub struct PoolOperation {
     pub entities_needing_stake: Vec<EntityType>,
     /// Whether the account is staked.
     pub account_is_staked: bool,
+    /// paymaster metadata
+    pub paymaster_metadata: Option<PaymasterMetadata>,
```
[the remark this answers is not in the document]
0xfourzerofour: This is a weird one, because its not good to add the network calls into the locked state. I get the balance from local memory or network outside the Paymaster Tracker and pass it in on add and update. It not as nice as I would like but ideally changes to the uo_pool and mempool inner would need to be made to clean it up

0xfourzerofour:
> There is a lot of complexity and a large amount of edge cases here. Would be worth writing tests on the uo_pool object with different combinations of deposits, reorgs of both UOs and deposits, dropped UOs etc.
> 
> In general I'm concerned that a bug here could lead us to unfairly block a paymaster and require a restart of Rundler to unstick. We probably need more visibility into these balances as well as some sort of kill switch. Metrics aren't ideal since this is high dimensional. Maybe we consider adding an admin API.

Yes, there is a lot of random complexity that is hard to make sure everything is captured. I agree we should have an admin endpoint to reset the balance/ call from the onchain state

crates/pool/src/mempool/mod.rs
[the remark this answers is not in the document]
dancoombs: How big are the changes?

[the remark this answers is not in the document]
0xfourzerofour: There are a few options, one would be to seperate out all the locked data into its own sub structure and call that instead of the entire locked state which would be a large overhaul. That would allow us to have some type of segmented middleware per domain before reaching the locked state.

The Pro of this method is due to the following function. Because we cannot do this without sending information over the wire right now due to the whole pool inner being locked

```
    async fn paymaster_balance(&self, paymaster: Address) -> ProviderResult<PaymasterMetadata> {
        if self.state.read().pool.paymaster_exists(paymaster) {
            let balance = self.state.read().pool.paymaster_balance(paymaster);

            let paymaster_meta = PaymasterMetadata {
                address: paymaster,
                balance,
                exists_in_pool: true,
            };

            return Ok(paymaster_meta);
        }

        let balance = self.entrypoint.balance_of(paymaster, None).await?;

        let paymaster_meta = PaymasterMetadata {
            address: paymaster,
            balance,
            exists_in_pool: false,
        };

        Ok(paymaster_meta)
    }

```

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,515 @@
...
+
+                paymaster_balance.pending =
+                    paymaster_balance.pending.saturating_sub(op_fee.max_op_cost);
+            }
+
+            self.user_op_fees.remove(&id);
+        }
+    }
+
+    pub(crate) fn update_paymaster_balance_from_unmined_op(&mut self, mined_op: &MinedOp) {
```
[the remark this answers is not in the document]
0xfourzerofour: Yeh the struct thing is a bit strange but it is used in a lot of places throughout the codebase so probably best to leave it for now

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -137,6 +171,9 @@ where
                     continue;
                 }
 
+                // update paymaster balances
+                state.pool.update_paymaster_balance_after_unmine(op);
+
                 if let Some(op) = state.pool.unmine_operation(op.hash) {
                     // Only account for a staked entity once
                     for entity_addr in op.staked_entities().map(|e| e.address).unique() {
```
alex-miao, line 179: Not related to any of your changes but for an unmine should we be doing `self.reputation.remove_included(entity_addr)` instead here, looks like we would be double counting

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,515 @@
...
+        if let Some(paymaster_balance) = self.paymaster_balances.get_mut(&paymaster) {
+            paymaster_balance.confirmed =
+                paymaster_balance.confirmed.saturating_sub(deposit_ammount);
+        }
+    }
+
+    pub(crate) fn update_paymaster_balance_from_deposit(
+        &mut self,
+        paymaster: Address,
+        deposit_ammount: U256,
```
alex-miao, line 94: same here

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,515 @@
...
+        }
+    }
+
+    fn replace_values(&mut self, max_op_cost: U256, paymaster: Address) {
+        if self.max_op_cost.ne(&max_op_cost) {
+            self.max_op_cost = max_op_cost;
+        }
+
+        if self.paymaster.ne(&paymaster) {
+            self.max_op_cost = max_op_cost;
```
alex-miao, line 213: did you mean `self.paymaster = paymaster;` here

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,515 @@
...
+impl UserOpFees {
+    fn new(paymaster: Address, max_op_cost: U256) -> Self {
+        Self {
+            paymaster,
+            max_op_cost,
+        }
+    }
+
+    fn replace_values(&mut self, max_op_cost: U256, paymaster: Address) {
+        if self.max_op_cost.ne(&max_op_cost) {
```
alex-miao, line 208: Can we get rid of the inequality checks here

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,515 @@
...
+            ));
+        }
+
+        // if user op already exists and is being replaced
+        if let Some(replacement) = self.user_op_fees.get_mut(id) {
+            let prev_limit = replacement.max_op_cost;
+            let _prev_paymaster = replacement.paymaster;
+
+            replacement.replace_values(max_op_cost, paymaster);
+            paymaster_balance.pending = paymaster_balance
```
alex-miao, line 176: I don't think this takes care of the case when you have a paymaster replacement. In that case you would want to subtract from the previous paymaster but add to the pending of the replacement paymaster

0xfourzerofour: Yes you are correct I did not finish the paymaster replacement. I can do that this weekend

crates/pool/src/mempool/uo_pool.rs
alex-miao: Just realized this is its own ticket: #498, I can handle this in a separate PR

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: fixed paymaster replacement

crates/pool/src/mempool/mod.rs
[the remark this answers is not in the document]
dancoombs: Going to move this info out from `PoolOperation` and as an optional param to `Pool::add_operation`

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,498 @@
...
+            ..Default::default()
+        }
+    }
+
+    pub(crate) fn paymaster_exists(&self, paymaster: Address) -> bool {
+        self.paymaster_balances.contains_key(&paymaster)
+    }
+
+    //TODO track if paymaster has become stale and can be removed from the pool
+    pub(crate) fn update_paymaster_balance_from_mined_op(&mut self, mined_op: &MinedOp) {
```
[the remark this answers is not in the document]
0xfourzerofour: Added removal logic to paymaster tracker

crates/pool/src/chain.rs
```diff
@@ -324,39 +367,86 @@ impl<P: Provider> Chain<P> {
...
+        let mined_ops = self.load_mined_ops(&logs);
+
+        Ok((mined_ops, deposits))
+    }
+
+    fn load_mined_ops(&self, logs: &Vec<Log>) -> Vec<MinedOp> {
+        let mut mined_ops = vec![];
+        for log in logs {
+            let entry_point = log.address;
+            if let Ok(event) = contract::parse_log::<UserOperationEventFilter>(log.clone()) {
```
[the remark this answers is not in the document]
0xfourzerofour: yeh could be cleaner instead of the for loop

crates/pool/src/mempool/pool.rs
```diff
@@ -77,9 +79,11 @@ pub(crate) struct PoolInner {
     /// when enough new blocks have passed.
     mined_hashes_with_block_numbers: BTreeSet<(u64, H256)>,
     /// Count of operations by entity address
-    count_by_address: HashMap<Address, usize>,
+    count_by_address: HashMap<Address, EntityCounter>,
     /// Submission ID counter
     submission_id: u64,
+    /// A field that kees track of paymaster balances across the mempool
```
alex-miao, line 85: typo: "keeps"

crates/pool/src/mempool/pool.rs
```diff
@@ -284,14 +372,30 @@ impl PoolInner {
...
     }
 
-    fn put_back_unmined_operation(&mut self, op: OrderedPoolOperation) -> MempoolResult<H256> {
-        self.add_operation_internal(op.po, Some(op.submission_id))
+    fn put_back_unmined_operation(
+        &mut self,
+        op: OrderedPoolOperation,
+        mined_op: &MinedOp,
+    ) -> MempoolResult<H256> {
+        // TODO make sure mined op cost is not added incorrectly
```
alex-miao, line 380: Just curious what you meant by this TODO

0xfourzerofour: Yeh I can delete that, forgot it was there. Just put it in a while back

crates/sim/src/simulation/simulation.rs
```diff
@@ -415,23 +415,22 @@ where
...
-        if let Some(aggregator_info) = entry_point_out.aggregator_info {
-            if !is_staked(aggregator_info.stake_info, self.sim_settings) {
-                entity_type_needs_stake(
-                    &mut entity_types_needing_stake,
-                    EntityType::Aggregator,
-                    entity_infos,
-                );
-            }
-        }
+        // if let Some(aggregator_info) = entry_point_out.aggregator_info {
```
alex-miao, line 418: Why don't we need this aggregator check anymore, can we delete the commented out code

alex-miao: This is the check for EREP-040 right, is this now taken care of somewhere else

0xfourzerofour: Yeh you are right, I am going to seperate the error into one specifically for that rule so we are not mixing different logic

0xfourzerofour: Just sorted that

alex-miao (approved):

crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -112,6 +112,8 @@ message MempoolOp {
   bool account_is_staked = 8;
 }
 
+
+
```
dancoombs, line 116: nit: remove added whitespace

crates/builder/src/transaction_tracker.rs
```diff
@@ -298,6 +298,7 @@ where
                     .get_transaction_status(tx.tx_hash)
                     .await
                     .context("tracker should check transaction status when the nonce changes")?;
+
```
dancoombs, line 301: nit: remove added whitespace

crates/pool/src/mempool/entity_tracker.rs
```diff
@@ -0,0 +1,56 @@
...
+            EntityType::Paymaster => self.paymaster = self.paymaster.saturating_sub(1),
+            EntityType::Aggregator => self.aggregator = self.aggregator.saturating_sub(1),
+            EntityType::Factory => self.factory = self.factory.saturating_sub(1),
+        }
+    }
+
+    pub(crate) fn includes_non_sender(&self) -> bool {
+        (self.paymaster + self.aggregator + self.factory) > 0
+    }
+}
```
dancoombs, line 56: These will be super trivial, but for the sake of improving our coverage, we should add some simple unit tests here

crates/sim/src/simulation/tracer.rs
```diff
@@ -94,6 +94,10 @@ impl AssociatedSlotsByAddress {
         };
         slot - next_smallest_slot < 128.into()
     }
+
+    pub(crate) fn to_set(&self) -> HashSet<Address> {
```
dancoombs, line 98: Lets give this a more descriptive name, its unclear what we are making into a set. Maybe just `addresses()`?

crates/sim/src/simulation/simulation.rs
```diff
@@ -72,6 +73,8 @@ pub struct SimulationSuccess {
     pub account_is_staked: bool,
     /// List of all addresses accessed during validation
     pub accessed_addresses: HashSet<Address>,
+    /// List of accesses associated addresses
```
dancoombs, line 76: This is pretty confusing. Are these the addresses where we accessed an associated slot? Associated to who? The account?

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -145,12 +156,37 @@ where
...
                 .unmined_ops
                 .iter()
                 .filter(|op| op.entry_point == self.config.entry_point);
             let mut mined_op_count = 0;
             let mut unmined_op_count = 0;
+
+            for deposit in deposits {
+                state
+                    .pool
+                    .update_paymaster_balance_after_deposit(deposit.address, deposit.amount)
```
dancoombs, line 180: Would it be simpler to just pass the full list of deposits to the pool instead of needing to loop here?

dancoombs: Could pass the unmined deposits in the same call, and move this logic into the pool

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -159,6 +195,9 @@ where
                 // Remove throttled ops that were included in the block
                 state.throttled_ops.remove(&op.hash);
 
+                // update paymaster balances
+                state.pool.update_paymaster_balance_after_mine(op);
```
dancoombs, line 199: Can we move this into `pool.mine_operation` so that we don't forget to call this? It looks like that is how we handle unmine.

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    pub(crate) fn add_or_update_balance(
+        &mut self,
+        po: &PoolOperation,
+        paymaster_metadata: &PaymasterMetadata,
+        unmined_op: Option<&MinedOp>,
+    ) -> MempoolResult<()> {
+        let id = po.uo.id();
+        let max_op_cost = po.uo.max_op_cost();
```
dancoombs, line 130: Discussed in Slack, we may want to switch to tracking total outstanding gas, and then checking the paymaster balance at a particular gas cost.

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,539 @@
...
+        self.paymaster_balances.contains_key(&paymaster)
+    }
+
+    pub(crate) fn clear(&mut self) {
+        self.user_op_fees.clear();
+        self.paymaster_balances.clear();
+    }
+
+    //TODO to be used with an admin rpc to debug issues in production
+    pub(crate) fn _clear_user_operation(&mut self, id: UserOperationId) {
```
dancoombs, line 51: How do we plan to use this one?

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,539 @@
...
+        Self {
+            paymaster,
+            max_op_cost,
+        }
+    }
+
+    fn replace_values(&mut self, max_op_cost: U256, paymaster: Address) {
+        self.max_op_cost = max_op_cost;
+        self.paymaster = paymaster;
+    }
```
dancoombs, line 244: Is this useful vs just creating a new one and using assignment?

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,539 @@
...
+        let max_op_cost = po.uo.max_op_cost();
+
+        if self.paymaster_exists(paymaster_metadata.address) {
+            let mut op_cost = max_op_cost;
+
+            // If update is from an unmined op, we need to make sure the op calculation
+            // does not include the previous subtraction of the actual gas cost
+            if let Some(mined) = unmined_op {
+                op_cost = op_cost.saturating_sub(mined.actual_gas_cost);
+            }
```
dancoombs, line 139: Not sure I understand why we need to do this. Can you explain further?

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,539 @@
...
+                paymaster_balance.confirmed.saturating_add(deposit_amount);
+        }
+    }
+
+    pub(crate) fn paymaster_metadata(&self, paymaster: Address) -> Option<PaymasterMetadata> {
+        if let Some(paymaster_balance) = self.paymaster_balances.get(&paymaster) {
+            return Some(PaymasterMetadata {
+                balance: paymaster_balance.pending_balance(),
+                address: paymaster,
+            });
```
dancoombs, line 117: Might as well also return the confirmed balance here, I'm sure that will be useful for the admin API.

crates/pool/src/mempool/pool.rs
```diff
@@ -303,14 +391,29 @@ impl PoolInner {
...
+        mined_op: &MinedOp,
+    ) -> MempoolResult<H256> {
+        let mut paymaster_meta = None;
+        if let Some(paymaster) = op.uo().paymaster() {
+            paymaster_meta = self.paymaster_metadata(paymaster);
+        }
+        self.add_operation_internal(
+            op.po,
+            Some(op.submission_id),
+            paymaster_meta,
```
dancoombs, line 406: This looks odd, why are we passing asking the paymaster tracker for the paymaster metadata and then passing that metadata back into the tracker?

0xfourzerofour: We spoke about this earlier in this PR, its because the locked state vs the network call. We dont want to make network calls inside the locked state so we had to do this

crates/pool/src/mempool/mod.rs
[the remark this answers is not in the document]
0xfourzerofour: @dancoombs here, but we spoke more about it in person before I left

crates/pool/src/chain.rs
```diff
@@ -198,7 +211,14 @@ impl<P: Provider> Chain<P> {
...
             .collect();
-        Ok(self.new_update(0, mined_ops, vec![]))
+
+        let entity_deposits: Vec<_> = self
+            .blocks
+            .iter()
+            .flat_map(|block| &block.entity_deposits)
+            .copied()
+            .collect();
+        Ok(self.new_update(0, mined_ops, vec![], entity_deposits, vec![]))
```
dancoombs, line 221: (can't comment above so adding this here) 

I'm concerned that if we see a reorg greater than block history we will end up in an inconsistent state where the paymaster balance will never correct itself. We likely need to handle this case somehow.

crates/pool/src/mempool/pool.rs
dancoombs: That makes sense, its just super hard to track where this is coming from. I wonder if there is a more intuitive interface we could use

crates/builder/src/transaction_tracker.rs
0xfourzerofour:
```
            for tx in self.transactions.iter().rev() {
                let status = self
                    .sender
                    .get_transaction_status(tx.tx_hash)
                    .await
                    .context("tracker should check transaction status when the nonce changes")?;

                if let TxStatus::Mined { block_number } = status {
                    let (gas_limit, gas_used) = self.get_mined_tx_gas_info(tx.tx_hash).await?;
                    out = TrackerUpdate::Mined {
                        tx_hash: tx.tx_hash,
                        nonce: self.nonce,
                        block_number,
                        attempt_number: tx.attempt_number,
                        gas_limit,
                        gas_used,
                    };
                    break;
                }
            }

```

for this one, I think the whitespace makes the code more legible

dancoombs: It may, but there are 0 other changes in this file other than this single whitespace change, so its good PR practice not to introduce and extra file diff with just a whitespace change.

crates/sim/src/simulation/simulation.rs
0xfourzerofour: It is needed for the STO-O41 rule which states 

* **[STO-041]** `UserOperation` may not use associated storage (of either its account or from staked entity) in a contract that is a "sender" of another UserOperation in the mempool.

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: It is just something I added that I think might be useful later down the road to not clear the whole paymaster tracking but for a specific userop that is blocked

crates/sim/src/simulation/simulation.rs
0xfourzerofour: Yeh its a weird one because the list of addresses are not specifically associated with a single entity but we use them to see if something in the mempool uses one of these addresses as a sender. 

what do you think of 

/// List of addresses that have associated storage slots
/// accessed within the simulation

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: I actually just fixed this, was a bit of a dumb way of doing things. Instead I am just adding back the actual cost value to to the confirmed balance after unmine so dont worry about this logic anymore

crates/pool/src/mempool/pool.rs
0xfourzerofour: Yeh I am not a huge fan of needing to do it but I think for now it is an ok option. I have thought about a few ways to do this but I mainly come back to a larger refactor of the mempool state config (like clearly separate state and also seperate the pool into different domains of concern)

crates/pool/src/chain.rs
0xfourzerofour: I updated the logic to use a helper contract to bulk lookup the paymaster deposit balances

crates/pool/src/mempool/pool.rs
dancoombs: Can you file a ticket to revisit this? Definitely want to clean this up

crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -550,8 +575,10 @@ message InvalidStorageAccess {
 
 message NotStaked {
   Entity entity = 1;
-  bytes min_stake = 2;
-  bytes min_unstake_delay = 3;
+  string accessed = 2;
```
dancoombs, line 578: What does this mean?

crates/pool/src/mempool/mod.rs
```diff
@@ -191,6 +201,13 @@ pub struct PoolOperation {
     pub entity_infos: EntityInfos,
 }
 
+#[derive(Debug, Default, Clone, Eq, PartialEq, Copy)]
+pub struct PaymasterMetadata {
+    pub address: Address,
+    pub confirmed_balance: U256,
+    pub balance: U256,
```
dancoombs, line 208: Lets rename `balance` to something more descriptive, is this pending balance? Is it just the pending portion? How do you calc the pending portion.

This public struct also needs doc comments.

crates/pool/proto/op_pool/op_pool.proto
0xfourzerofour: It could always be an address but it just makes it easier for the person using the api to understand the error message instead of having to figure out if that address is specific entity. I saw this on the reference bundler and liked it

crates/types/src/user_operation.rs
```diff
@@ -62,6 +62,14 @@ impl UserOperation {
         Self::get_address_from_field(&self.init_code)
     }
 
+    /// Get max amount the user operation will cost
+    pub fn max_op_cost(&self) -> U256 {
+        self.call_gas_limit
+            .saturating_add(self.verification_gas_limit)
+            .saturating_add(self.pre_verification_gas)
+            .saturating_mul(self.max_fee_per_gas)
+    }
```
dancoombs, line 71: @alex-miao don't we have a function for this somewhere else?

This implementation is incorrect, you need to multiply `self.verification_gas_limit` by 3 if a paymaster is being used. I think we have a correct implementation somewhere.

0xfourzerofour: Oh yeh true I remember alex implementing that logic

crates/pool/src/chain.rs
```diff
@@ -171,13 +185,19 @@ impl<P: Provider> Chain<P> {
             current_block_number < new_block_number + self.settings.history_size,
             "new block number {new_block_number} should be greater than start of history (current block: {current_block_number})"
         );
+
+        if new_block_number <= current_block_number.saturating_sub(self.settings.history_size) {
+            return self.reset_and_initialize(new_head).await;
+        }
```
dancoombs, line 191: What is this new condition capturing? We should add a warning log like the code below, probably also change the code below into an `else if` for clarity

crates/pool/src/chain.rs
```diff
@@ -74,6 +78,15 @@ pub struct MinedOp {
...
+    pub actual_gas_cost: U256,
+    pub paymaster: Option<Address>,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct DepositInfo {
+    pub address: Address,
+    pub entrypoint: Address,
+    pub amount: U256,
 }
```
dancoombs, line 90: Shouldn't our linting rules be rejecting this since we don't have comments on the public fields?

crates/pool/src/chain.rs
0xfourzerofour: Actually this was meant to be deleted, I have removed it now. I was testing some stuff a couple days ago and forgot to delete it

crates/pool/src/chain.rs
```diff
@@ -66,6 +67,9 @@ pub struct ChainUpdate {
     pub reorg_depth: u64,
     pub mined_ops: Vec<MinedOp>,
     pub unmined_ops: Vec<MinedOp>,
+    pub entity_deposits: Vec<DepositInfo>,
+    pub unmined_entity_deposits: Vec<DepositInfo>,
+    pub reorg_larger_than_history: bool,
```
dancoombs, line 72: nit: lets add comments to these fields

crates/pool/src/chain.rs
```diff
@@ -859,4 +1032,11 @@ mod tests {
         hash.0[0] = n;
         hash
     }
+
+    /// Helper that makes fake addresses.
```
dancoombs, line 1032: nit: doesn't need to be a doc comment

crates/pool/src/chain.rs
0xfourzerofour: Yeh something might be auto removing the need as there are many structs in the chain.rs file without comments for public fields. In general I think comments should not be put on things just for the sake of it as it just adds bloat, I think if its self explanatory it can do without but thats just my opinion

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -43,14 +43,21 @@ use crate::{
 /// Wrapper around a pool object that implements thread-safety
 /// via a RwLock. Safe to call from multiple threads. Methods
 /// block on write locks.
-pub(crate) struct UoPool<R: ReputationManager, P: Prechecker, S: Simulator, SM: StakeManager> {
+pub(crate) struct UoPool<
+    R: ReputationManager,
+    P: Prechecker,
+    S: Simulator,
+    E: EntryPoint,
+    PH: PaymasterHelper,
```
dancoombs, line 51: This is getting obnoxious. We should look into refactoring this in 0.2, mind making a ticket?

0xfourzerofour: Yeh this is one reason why I brought up removing the ReputationManager and only make these traits if really needed

crates/pool/src/chain.rs
dancoombs: I think its better to err on the bloat side and lint on all public fields having comments, else you end up missing comments on non self-explanatory fields.

Since the rest of this file is missing them we can skip for this PR.

dancoombs: Actually wait that warn should cause this to not compile, weird

dancoombs: Ah its because these public fields are not exposed outside of this crate

crates/types/src/user_operation.rs
dancoombs: https://github.com/alchemyplatform/rundler/blob/15cbd85acfc5e9dbc0e51f891d4262aa78f85525/crates/sim/src/gas/gas.rs#L207

Lets move this function to be on `UserOperation`

crates/pool/proto/op_pool/op_pool.proto
dancoombs: I think we should always print the address and if we want to add the accessed entity kind information we can add that as optional.

So `bytes accessed_address` and `Entity accessed_entity` where `accessed_entity` is optional.

crates/types/src/user_operation.rs
0xfourzerofour: Done

0xfourzerofour: I just need to fix the tests with the correct calc

crates/sim/src/simulation/simulation.rs
```diff
@@ -661,8 +668,14 @@ pub enum SimulationViolation {
     #[display("code accessed by validation has changed since the last time validation was run")]
     CodeHashChanged,
     /// The user operation contained an entity that accessed storage without being staked
-    #[display("{0.kind} must be staked")]
-    NotStaked(Entity, U256, U256),
+    #[display("{}")]
```
dancoombs, line 671: This will display `NotStaked`, should we have a sentence like the rest of them?

crates/rpc/src/eth/error.rs
```diff
@@ -224,12 +254,15 @@ impl From<SimulationViolation> for EthRpcError {
...
                 Self::InvalidStorageAccess(entity.kind, slot.address, slot.slot)
             }
-            SimulationViolation::NotStaked(entity, min_stake, min_unstake_delay) => {
-                Self::StakeTooLow(StakeTooLowData::new(entity, min_stake, min_unstake_delay))
-            }
+            SimulationViolation::NotStaked(stake_data) => Self::StakeTooLow(StakeTooLowData::new(
+                stake_data.entity,
+                stake_data.min_stake,
+                stake_data.min_unstake_delay,
+            )),
```
dancoombs, line 265: Since we have more info in the `stake_data`, should we return it as part of this RPC error? Or is this a standardized error?

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -328,6 +394,15 @@ where
         if let Some(agg) = &sim_result.aggregator {
             return Err(MempoolError::UnsupportedAggregator(agg.address));
         }
+
+        if replacement.is_none() {
```
dancoombs, line 398: Is there a reason why this doesn't apply to replacement operations?

crates/pool/src/mempool/paymaster.rs
```diff
@@ -0,0 +1,548 @@
...
+        // check there is enough balance
+        if !paymaster_balance.validate_user_op_cost(max_op_cost) {
+            return Err(MempoolError::PaymasterBalanceTooLow(
+                max_op_cost,
+                paymaster_balance.pending_balance(),
+            ));
+        }
+
+        // if user op already exists and is being replaced
+        if let Some(replacement) = self.user_op_fees.get_mut(id) {
```
dancoombs, line 194: Should this logic also be in the `add_paymaster_balance` function? For example if `add_or_update_balance` is called and the paymaster has changed to a new paymaster?

crates/pool/src/mempool/uo_pool.rs
0xfourzerofour: If it is a replacement, the STO-041 storage rule will fail as the sender address will already be in the mempool from what I can remember from implementation but I will have a run through it again

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: This is a good point, let me think about this. I would rather not duplicate the code so it could be worth seeing if there is a way to consolidate the logic but this is a good catch

crates/pool/src/mempool/uo_pool.rs
dancoombs: That sounds like an edge case we have to handle. Its possible the replacement operation causes this rule to fail, so we need to handle it while accounting for the replacement

crates/pool/src/mempool/paymaster.rs
dancoombs: Sounds good, lets be sure to add a test for this scenario

0xfourzerofour: I simplified the logic a bit and added a test so that we did not have to duplicate logic across functions and added a test for that case

crates/rpc/src/eth/error.rs
0xfourzerofour:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "message": "paymaster stake too low",
    "data": {
      "paymaster": "0x123456789012345678901234567890123456790",
      "minimumStake": "0xde0b6b3a7640000",
      "minimumUnstakeDelay": "0x15180"
    },
    "code": -32504
  }
}
```

This is the standard error that the spec lists in the docs but it does not says that there cannot be additional fields so I would not be apposed to it

crates/pool/src/mempool/uo_pool.rs
0xfourzerofour: This one one is a bit strange, the test still passes when I remove the need to check for replacements as long as I am checking whether the user op sender is not the same as the accessed storage address. I am going to reach out to the 4337 team and clarify but for now it passes the test. 

No other reference bundler has implemented this rule in the mempool yet so I would like to double check with the 4337 team to make sure this is correct as I believe there are edge cases here.

crates/rpc/src/eth/error.rs
dancoombs: We might as well add additional fields

crates/pool/src/mempool/uo_pool.rs
0xfourzerofour: I am just going to leave this by checking that the accessed address is also not equal to the uo's sender. It is not worth spending much more time on this especially if it passes the spec tests and this is a rule that is used to protect the bundler instead of user operations. 

There has also been updates to the spec rule within the last two weeks so I believe this will probably continue to change.

dancoombs (approved):
Lets go! Thanks for working through this, so many edge cases to discover and work out.

🚀

PR #519, opened 2023-11-30
feat(pool): handle SREP-020 and refactor to remove remove_entity endpoint

### LINKED ISSUE #449: [spec] SREP-020

alex-miao:
Closes #449 

## Proposed Changes

  -
  -
  -


crates/builder/src/bundle_proposer.rs
```diff
@@ -293,9 +293,11 @@ where
             .iter()
             .map(|(op, _)| op.uo.sender)
             .collect();
-        let mut groups_by_aggregator = LinkedHashMap::<Option<Address>, AggregatorGroup>::new();
-        let mut rejected_ops = Vec::<(UserOperation, EntityInfos)>::new();
-        let mut entity_updates: BTreeMap<Address, EntityUpdate> = BTreeMap::new();
+        let mut context = ProposalContext {
```
dancoombs, line 296: Probably should just have a `ProposalContext::new()`

crates/pool/src/mempool/mod.rs
```diff
@@ -162,6 +159,8 @@ pub struct PoolOperation {
     pub account_is_staked: bool,
     /// Staking information about all the entities.
     pub entity_infos: EntityInfos,
+    /// The block number at which the operation first entered the pool
+    pub block_seen: u64,
```
dancoombs, line 163: Lets call this `sim_block_number` (since we already have `sim_block_hash` above)

crates/pool/src/mempool/pool.rs
```diff
@@ -178,6 +186,39 @@ impl PoolInner {
...
+    /// Remove all but THROTTLED_ENTITY_MEMPOOL_COUNT operations that are within THROTTLED_ENTITY_LIVE_BLOCKS of head
+    /// using the given entity, returning the hashes of the removed operations.
+    pub(crate) fn throttle_entity(
+        &mut self,
+        entity: Entity,
+        current_block_number: u64,
+    ) -> Vec<H256> {
+        let mut uos_kept = self.config.throttled_entity_mempool_count;
+        let to_remove = self
+            .by_hash
```
dancoombs, line 198: Lets keep the highest bidding UOs, can we iterate by the `best` map?

dancoombs (commented):
Nice! Couple comments

crates/builder/src/bundle_proposer.rs
alex-miao: Added a `new` function to `ProposalContext`

crates/pool/src/mempool/mod.rs
alex-miao: Decided to also refactor the code a bit and make sure whenever we have a block hash we also know the corresponding block number.

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -172,9 +201,16 @@ where
                 .forget_mined_operations_before_block(update.earliest_remembered_block_number);
             // Remove throttled ops that are too old
             let mut to_remove = HashSet::new();
-            for (hash, block) in state.throttled_ops.iter() {
-                if update.latest_block_number - block > self.config.throttled_entity_live_blocks {
-                    to_remove.insert(*hash);
+            for hash in state.throttled_ops.iter() {
```
0xfourzerofour, line 204: What is the reason to remove to block from the hashmap and use a hashset instead?

0xfourzerofour (approved):
overall looks good!

crates/pool/src/mempool/uo_pool.rs
alex-miao: This is because we now the block number that every single op gets added to, not just for throttled ops, since when an entity is throttled it already might have ops in the pool and we need to track what block number they got added at to know which ones can stay. This block number gets stored in `SimulationSuccess`.

0xfourzerofour: nice, thanks for the explanation

dancoombs (approved):

PR #522, opened 2023-12-04
feat(pool): handle UREP-020 and track reputation of unstaked entities

### LINKED ISSUE #354: pool: Still track reputation of non-stake-allowed entities.
**Describe the feature**
Our mempool allowlist feature has the ability to allow entities to bypass staking requirements. This will also allow them to avoid any reputation checks (since non-staked entities don't have reputation). We should allow a configuration option that will still check the reputation of these non-staked-allowed entities. There are use cases where an entity might not be able to stake, and we still want them to be able to access associated storage, however we don't want them to be able to DOS the system. Since stake is only a sybil resistance mechanism, adding to an allowlist serves the same purpose, as long as reputation is still checked.

Note that this will not work in the P2P mempool.

### LINKED ISSUE #454: [spec] UREP-20

alex-miao:
Closes #454, closes #354

## Proposed Changes

  - Track reputation of all entities
  - Move `SAME_SENDER_MEMPOOL_COUNT` check out from `PoolInner` into `UoPool`
  - Do UREP-020 check in the same pool read lock as `SAME_SENDER_MEMPOOL_COUNT` check
  - Fix bug where we increment instead of decrement the reputation of an entity for an unmined transaction

### DIFF
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -114,7 +114,7 @@ pub struct PoolArgs {
         long = "pool.throttled_entity_live_blocks",
         name = "pool.throttled_entity_live_blocks",
         env = "POOL_THROTTLED_ENTITY_LIVE_BLOCKS",
-        default_value = "4"
+        default_value = "10"
     )]
     pub throttled_entity_live_blocks: u64,
 }
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -59,12 +59,12 @@ pub struct PoolArgs {
     pub max_size_in_bytes: usize,
 
     #[arg(
-        long = "pool.max_userops_per_sender",
-        name = "pool.max_userops_per_sender",
-        env = "POOL_MAX_USEROPS_PER_SENDER",
+        long = "pool.same_sender_mempool_count",
+        name = "pool.same_sender_mempool_count",
+        env = "SAME_SENDER_MEMPOOL_COUNT",
         default_value = "4"
     )]
-    pub max_userops_per_sender: usize,
+    pub same_sender_mempool_count: usize,
 
     #[arg(
         long = "pool.min_replacement_fee_increase_percentage",
@@ -156,7 +156,7 @@ impl PoolArgs {
                     chain_id: common.chain_id,
                     // Currently use the same shard count as the number of builders
                     num_shards: common.num_builders,
-                    max_userops_per_sender: self.max_userops_per_sender,
+                    same_sender_mempool_count: self.same_sender_mempool_count,
                     min_replacement_fee_increase_percentage: self
                         .min_replacement_fee_increase_percentage,
                     max_size_of_pool_bytes: self.max_size_in_bytes,
--- a/crates/pool/src/mempool/error.rs
+++ b/crates/pool/src/mempool/error.rs
@@ -35,8 +35,8 @@ pub enum MempoolError {
     /// and the replacement operation has lower gas price.
     #[error("Replacement operation underpriced. Existing priority fee: {0}. Existing fee: {1}")]
     ReplacementUnderpriced(U256, U256),
-    /// Max operations reached for this sender
-    #[error("Max operations ({0}) reached for sender {1}")]
+    /// Max operations reached for unstaked sender [UREP-010] or unstaked non-sender entity [UREP-020]
+    #[error("Max operations ({0}) reached for entity {1}")]
     MaxOperationsReached(usize, Address),
     /// Multiple roles violation
     /// Spec rule: STO-040
--- a/crates/pool/src/mempool/mod.rs
+++ b/crates/pool/src/mempool/mod.rs
@@ -37,7 +37,6 @@ use ethers::types::{Address, H256, U256};
 use mockall::automock;
 use rundler_sim::{EntityInfos, MempoolConfig, PrecheckSettings, SimulationSettings};
 use rundler_types::{Entity, EntityType, EntityUpdate, UserOperation, ValidTimeRange};
-use strum::IntoEnumIterator;
 use tonic::async_trait;
 pub(crate) use uo_pool::UoPool;
 
@@ -215,9 +214,11 @@ pub struct PaymasterMetadata {
 impl PoolOperation {
     /// Returns true if the operation contains the given entity.
     pub fn contains_entity(&self, entity: &Entity) -> bool {
-        self.entity_address(entity.kind)
-            .map(|address| address == entity.address)
-            .unwrap_or(false)
+        if let Some(e) = self.entity_infos.get(entity.kind) {
+            e.address == entity.address
+        } else {
+            false
+        }
     }
 
     /// Returns true if the operation requires the given entity to stake.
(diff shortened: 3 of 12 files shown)


crates/pool/src/mempool/error.rs
```diff
@@ -35,8 +35,8 @@ pub enum MempoolError {
     /// and the replacement operation has lower gas price.
     #[error("Replacement operation underpriced. Existing priority fee: {0}. Existing fee: {1}")]
     ReplacementUnderpriced(U256, U256),
-    /// Max operations reached for this sender
-    #[error("Max operations ({0}) reached for sender {1}")]
+    /// Max operations reached for unstaked sender [UREP-010] or unstaked non-sender entity [UREP-020]
+    #[error("Max operations ({0}) reached for entity {1}")]
     MaxOperationsReached(usize, Address),
```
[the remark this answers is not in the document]
dancoombs: I think this is fine, its descriptive enough for both

crates/pool/src/mempool/reputation.rs
```diff
@@ -323,7 +323,7 @@ impl AddressReputation {
             // make sure we aren't dividing by 0
             0
         } else {
-            included * self.params.inclusion_rate_factor / seen + std::cmp::min(included, 10_000)
+            self.params.inclusion_rate_factor * included / seen + std::cmp::min(included, 10_000)
```
alex-miao, line 326: I honestly think this formula doesn't really make sense, not sure why we need this middle term `self.params.inclusion_rate_factor * included / seen` cause it's some number between 0 and 10, but you already have this constant term `self.params.same_unstaked_entity_mempool_count` that is equal to 10 and its really this `std::cmp::min(included, 10_000)` term that dominates if you do have `included` ops. Also in the [spec](https://github.com/eth-infinitism/account-abstraction/blob/6ade7f1cdb4a0ea1328f5ea8edbfb5fab65deebd/eip/EIPS/eip-aa-rules.md?plain=1#L256) the formula has a hanging parenthesis, so not sure if what's written is what is intended.

bin/rundler/src/cli/pool.rs
```diff
@@ -59,12 +59,12 @@ pub struct PoolArgs {
     pub max_size_in_bytes: usize,
 
     #[arg(
-        long = "pool.max_userops_per_sender",
-        name = "pool.max_userops_per_sender",
-        env = "POOL_MAX_USEROPS_PER_SENDER",
+        long = "pool.same_sender_mempool_count",
+        name = "pool.same_sender_mempool_count",
+        env = "SAME_SENDER_MEMPOOL_COUNT",
```
dancoombs, line 64: Planning on moving this to `ChainSpec` (when I get around to it)

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -1092,8 +1126,8 @@ mod tests {
 
         fn get_ops_allowed(&self, address: Address) -> u64 {
             let counts = self.counts.read();
-            let seen = *counts.seen.get(&address).unwrap();
-            let included = *counts.included.get(&address).unwrap();
+            let seen = *counts.seen.get(&address).unwrap_or(&0);
+            let included = *counts.included.get(&address).unwrap_or(&0);
             let inclusion_based_count = if seen == 0 {
                 // make sure we aren't dividing by 0
                 0
```
dancoombs, line 1133: Lets add tests for this new constraint. Both for UREP-10 and UREP-20

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -321,6 +321,34 @@ where
             hash
         };
 
+        // Check sender count in mempool. If sender has too many operations, must be staked
```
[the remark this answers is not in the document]
dancoombs: This would probably explain why `test_max_allowed_ops_unstaked_sender` is failing currently

crates/pool/src/mempool/reputation.rs
dancoombs: I'll follow up with them. Seems odd.

crates/pool/src/mempool/mod.rs
```diff
@@ -206,6 +206,34 @@ impl PoolOperation {
...
+                unstaked_entities.push(Entity::new(EntityType::Paymaster, paymaster.address))
+            }
+        }
+        if let Some(aggregator) = self.entity_infos.aggregator {
+            if !aggregator.is_staked {
+                unstaked_entities.push(Entity::new(EntityType::Aggregator, aggregator.address))
+            }
+        }
+
+        unstaked_entities.into_iter()
```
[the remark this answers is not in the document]
alex-miao: Not sure there is a clean way to do that since we have to pull this info from `EntityInfos`, which has each entity as a field in a struct so I think we just have to check each field one by one. On another note, this function might actually be better suited to be under `EntityInfos` than `PoolOperation`.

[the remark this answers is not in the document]
dancoombs: Don't we have `self.entity_infos`? This has the same lifetime as self so you should be able to capture self in the iterator.

crates/pool/src/mempool/pool.rs
```diff
@@ -673,20 +660,6 @@ mod tests {
         assert!(pool.best.is_empty());
     }
 
-    #[test]
-    fn too_many_ops() {
```
alex-miao, line 677: This is no longer relevant at the `pool.rs` level, will replicate this test in `uo_pool.rs`.

crates/pool/src/mempool/mod.rs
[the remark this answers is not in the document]
dancoombs: Our entity types really annoy me. We Have `EntityInfo` and `Entity` that have different meanings. We should clean this up. Can you add a ticket?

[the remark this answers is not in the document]
dancoombs: In the meantime, how about this:

On `EntityInfos` in `simulation.rs`:
```rust
    /// entities
    pub fn entities(&'_ self) -> impl Iterator<Item = (EntityType, EntityInfo)> + '_ {
        EntityType::iter().filter_map(|t| self.get(t).map(|info| (t, info)))
    }
```

Then on `PoolOperation` here:
```rust
    pub fn staked_entities(&'_ self) -> impl Iterator<Item = Entity> + '_ {
        self.entity_infos.entities().filter_map(|(t, entity)| {
            if self.requires_stake(t) {
                Entity::new(t, entity.address).into()
            } else {
                None
            }
        })
    }

    pub fn unstaked_entities(&'_ self) -> impl Iterator<Item = Entity> + '_ {
        self.entity_infos.entities().filter_map(|(t, entity)| {
            if entity.is_staked {
                None
            } else {
                Entity::new(t, entity.address).into()
            }
        })
    }
```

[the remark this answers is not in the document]
dancoombs: Might as well also remove `entity_address` and then 

```rust
    pub fn contains_entity(&self, entity: &Entity) -> bool {
        if let Some(e) = self.entity_infos.get(entity.kind) {
            e.address == entity.address
        } else {
            false
        }
    }
```

crates/pool/src/mempool/reputation.rs
dancoombs: Lets leave this with the word of the spec for now

dancoombs (commented):
Lets make sure this is as close to compliant with the latest ERC-7562, while still passing all tests on the commit we've pegged to.

If there are changes in ERC-7562 that cause failures, lets capture them as tickets so we know to update them when we move to a new spec test commit.

One more small nit on the PR to fix up the entities stuff.

0xfourzerofour:
Do these changes fix those errors we spoke about in the spec tests?

bin/rundler/src/cli/pool.rs
```diff
@@ -114,7 +114,7 @@ pub struct PoolArgs {
         long = "pool.throttled_entity_live_blocks",
         name = "pool.throttled_entity_live_blocks",
         env = "POOL_THROTTLED_ENTITY_LIVE_BLOCKS",
-        default_value = "4"
+        default_value = "10"
```
dancoombs, line 117: Was this the issue that caused the tests to fail?

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -415,6 +415,34 @@ where
             entity_infos: sim_result.entity_infos,
         };
 
+        // Check sender count in mempool. If sender has too many operations, must be staked
```
dancoombs, line 418: Can this check happen prior to simulation? As a rule of thumb we want to front load any checks that can be preformed prior to simulation.

dancoombs: The sender, paymaster and factory entities can be extracted prior to simulation for sure. Aggregators are annoying in that they require running the account's validation function.

If this change is too big for this PR lets add a ticket to clean up.

dancoombs (changes requested):
One last comment and then this is good to go. Can you squash to a single commit?

bin/rundler/src/cli/pool.rs
alex-miao: No this is unrelated to the test failing since the issue in the test was that 12 blocks passed which is still greater than the correct default.

crates/pool/src/mempool/uo_pool.rs
alex-miao: Added a ticket for this here: https://github.com/alchemyplatform/rundler/issues/571. I think splitting out getting entity staking state and simulation would be beneficial, even for our current simulation flow and removing the need for `EntityInfos` to be returned by `SimulationError`.

0xfourzerofour:
@dancoombs I enabled squash and merge after that chat the other day, will use the PR title as the commit message

dancoombs (approved):

bin/rundler/src/cli/pool.rs
dancoombs: Got it!

PR #533, opened 2023-12-13
feat(compliance): fixes for spec tests

### LINKED ISSUE #531: STO-033
Update tracer to differentiate between read and write storage access

0xfourzerofour:
Closes #531 

## Proposed Changes

  - Seperated read and write access from tracer
  - change tracer parsing logic to follow similar implementation as the reference bundler https://github.com/eth-infinitism/bundler/blob/ba78f2067e33df399bb7d34a9d1f972834d23c80/packages/validation-manager/src/TracerResultParser.ts#L315
  - Changed error codes to work with spec tests
  - Update error message to work with spec tests

### DIFF
--- a/crates/rpc/src/eth/error.rs
+++ b/crates/rpc/src/eth/error.rs
@@ -35,7 +35,7 @@ const PAYMASTER_VALIDATION_REJECTED_CODE: i32 = -32501;
 const OPCODE_VIOLATION_CODE: i32 = -32502;
 const OUT_OF_TIME_RANGE_CODE: i32 = -32503;
 const THROTTLED_OR_BANNED_CODE: i32 = -32504;
-const STAKE_TOO_LOW_CODE: i32 = -32505;
+const _STAKE_TOO_LOW_CODE: i32 = -32505;
 const UNSUPORTED_AGGREGATOR_CODE: i32 = -32506;
 const SIGNATURE_CHECK_FAILED_CODE: i32 = -32507;
 const EXECUTION_REVERTED: i32 = -32521;
@@ -70,7 +70,7 @@ pub enum EthRpcError {
     #[error("operation is out of time range")]
     OutOfTimeRange(OutOfTimeRangeData),
     /// Entity throttled or banned
-    #[error("entity throttled or banned")]
+    #[error("{} {:#032x} throttled or banned", .0.kind, .0.address)]
     ThrottledOrBanned(Entity),
     /// Entity stake/unstake delay too low
     #[error("entity stake/unstake delay too low")]
@@ -251,6 +251,8 @@ impl From<EthRpcError> for ErrorObjectOwned {
             }
             EthRpcError::OpcodeViolation(_, _)
             | EthRpcError::OpcodeViolationMap(_)
+            | EthRpcError::StakeTooLow(_)
+            | EthRpcError::SimulationFailed(_)
             | EthRpcError::InvalidStorageAccess(_, _, _) => rpc_err(OPCODE_VIOLATION_CODE, msg),
             EthRpcError::OutOfTimeRange(data) => {
                 rpc_err_with_data(OUT_OF_TIME_RANGE_CODE, msg, data)
@@ -258,7 +260,6 @@ impl From<EthRpcError> for ErrorObjectOwned {
             EthRpcError::ThrottledOrBanned(data) => {
                 rpc_err_with_data(THROTTLED_OR_BANNED_CODE, msg, data)
             }
-            EthRpcError::StakeTooLow(data) => rpc_err_with_data(STAKE_TOO_LOW_CODE, msg, data),
             EthRpcError::UnsupportedAggregator(data) => {
                 rpc_err_with_data(UNSUPORTED_AGGREGATOR_CODE, msg, data)
             }
@@ -268,7 +269,6 @@ impl From<EthRpcError> for ErrorObjectOwned {
             EthRpcError::OperationAlreadyKnown => rpc_err(INVALID_PARAMS_CODE, msg),
             EthRpcError::SignatureCheckFailed => rpc_err(SIGNATURE_CHECK_FAILED_CODE, msg),
             EthRpcError::PrecheckFailed(_) => rpc_err(CALL_EXECUTION_FAILED_CODE, msg),
-            EthRpcError::SimulationFailed(_) => rpc_err(CALL_EXECUTION_FAILED_CODE, msg),
             EthRpcError::ExecutionReverted(_) => rpc_err(EXECUTION_REVERTED, msg),
             EthRpcError::OperationRejected(_) => rpc_err(INVALID_PARAMS_CODE, msg),
         }
--- a/crates/rpc/src/types.rs
+++ b/crates/rpc/src/types.rs
@@ -161,6 +161,7 @@ pub struct UserOperationReceipt {
 
 /// Reputation of an entity
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct RpcReputation {
     /// Entity address
     pub address: Address,
@@ -169,7 +170,7 @@ pub struct RpcReputation {
     /// Number of operations included in this interval
     pub ops_included: U256,
     /// Reputation status
-    pub status: ReputationStatus,
+    pub status: Option<ReputationStatus>,
 }
 
 impl From<RpcReputation> for Reputation {
@@ -178,7 +179,7 @@ impl From<RpcReputation> for Reputation {
             address: rpc_reputation.address,
             ops_seen: rpc_reputation.ops_seen.as_u64(),
             ops_included: rpc_reputation.ops_included.as_u64(),
-            status: rpc_reputation.status,
+            status: rpc_reputation.status.unwrap_or(ReputationStatus::Ok),
         }
     }
 }
@@ -191,7 +192,7 @@ impl TryFrom<Reputation> for RpcReputation {
             address: reputation.address,
             ops_seen: reputation.ops_seen.into(),
             ops_included: reputation.ops_included.into(),
-            status: reputation.status,
+            status: Some(reputation.status),
         })
     }
 }
--- a/crates/rpc/src/types.rs
+++ b/crates/rpc/src/types.rs
@@ -162,7 +162,19 @@ pub struct UserOperationReceipt {
 /// Reputation of an entity
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct RpcReputation {
+pub struct RpcReputationInput {
+    /// Entity address
+    pub address: Address,
+    /// Number of operations seen in this interval
+    pub ops_seen: U256,
+    /// Number of operations included in this interval
+    pub ops_included: U256,
+}
+
+/// Reputation of an entity
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcReputationOutput {
     /// Entity address
     pub address: Address,
     /// Number of operations seen in this interval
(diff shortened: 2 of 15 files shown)


crates/rpc/src/eth/error.rs
```diff
@@ -251,14 +251,15 @@ impl From<EthRpcError> for ErrorObjectOwned {
...
+            | EthRpcError::StakeTooLow(_)
+            | EthRpcError::SimulationFailed(_)
             | EthRpcError::InvalidStorageAccess(_, _, _) => rpc_err(OPCODE_VIOLATION_CODE, msg),
             EthRpcError::OutOfTimeRange(data) => {
                 rpc_err_with_data(OUT_OF_TIME_RANGE_CODE, msg, data)
             }
             EthRpcError::ThrottledOrBanned(data) => {
                 rpc_err_with_data(THROTTLED_OR_BANNED_CODE, msg, data)
             }
-            EthRpcError::StakeTooLow(data) => rpc_err_with_data(STAKE_TOO_LOW_CODE, msg, data),
```
alex-miao, line 261: Why don't we need this anymore

0xfourzerofour: it was just added up a few lines to use opcode violation code. Techically it is correct but the spec tests just opcode violations for most errors so we are just complying with the 4337 team

crates/sim/src/simulation/simulation.rs
```diff
@@ -703,9 +696,9 @@ struct ValidationContext {
     entity_infos: EntityInfos,
     tracer_out: SimulationTracerOutput,
     entry_point_out: ValidationOutput,
-    is_unstaked_wallet_creation: bool,
```
alex-miao, line 706: Nice, good to see that we can derive this from what we already have and get rid of this redundancy. I've been thinking that it might be possible to consolidate `entity_infos` and `entities_needing_stake` but haven't looked into it too much.

crates/sim/src/simulation/simulation.rs
```diff
@@ -804,48 +823,81 @@ fn entity_type_needs_stake(
...
+            if initcode_length > 2
+                // special case: account.validateUserOp is allowed to use assoc storage if factory is staked.
+                // [STO-022], [STO-021]
+                && !(entity.address.eq(&sender)
+                    && entity_infos
+                        .factory
+                        .expect("Factory needs to be present and staked")
+                        .is_staked)
+            {
+                required_stake_slot = Some(slot);
```
alex-miao, line 885: nit: we could potentially break out of the loop here, unless you prefer returning `Banned` before `NeedsStake`

crates/sim/tracer/src/validationTracer.ts
```diff
@@ -233,13 +236,19 @@ type StringSet = Record<string, boolean | undefined>;
...
     step(log, db): void {
       if (!entryPointAddress) {
         entryPointAddress = toHex(log.contract.getAddress());
       }
-      if (log.getGas() < log.getCost()) {
+
+      const opcode = log.op.toString();
+
+      if (log.getGas() < log.getCost() || (
+        opcode === 'SSTORE' && log.getGas() < 2300
```
alex-miao, line 250: nit: can we make 2300 a constant

crates/sim/tracer/src/validationTracer.ts
```diff
@@ -312,19 +319,33 @@ type StringSet = Record<string, boolean | undefined>;
...
           // The entry point can access whatever it wants, but otherwise track
           // access for this phase so we can check validity later.
-          computeIfAbsent(
-            currentPhase.storageAccesses,
-            addressHex,
-            (): StringSet => ({})
-          )[slotHex] = true;
+          if (opcode == "SLOAD") {
+            // Read access
+            if (access.reads[slotHex] == null && access.writes[slotHex] == null) {
```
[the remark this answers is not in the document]
0xfourzerofour:
```
          // read slot values before this UserOp was created
          // (so saving it if it was written before the first read)

```

This is the explanation from ref bundler

crates/sim/src/simulation/simulation.rs
```diff
@@ -804,48 +823,81 @@ fn entity_type_needs_stake(
...
+        let is_sender_associated = slots_by_address.is_associated_slot(sender, *slot);
+        // [STO-032]
+        let is_entity_associated = slots_by_address.is_associated_slot(entity.address, *slot);
+        // [STO-031]
+        let is_same_address = address.eq(&entity.address);
+        // [STO-033]
+        let is_read_permission = !access_info.writes.contains_key(slot);
+
+        if is_sender_associated {
+            if initcode_length > 2
```
alex-miao, line 876: Is this because an "empty" initcode is `0x`

alex-miao (approved):
🚀

crates/sim/src/simulation/simulation.rs
0xfourzerofour: yeh exactly

crates/sim/src/simulation/simulation.rs
0xfourzerofour: I think I would prefer to just keep it as is for readability

crates/rpc/src/eth/error.rs
```diff
@@ -35,7 +35,7 @@ const PAYMASTER_VALIDATION_REJECTED_CODE: i32 = -32501;
 const OPCODE_VIOLATION_CODE: i32 = -32502;
 const OUT_OF_TIME_RANGE_CODE: i32 = -32503;
 const THROTTLED_OR_BANNED_CODE: i32 = -32504;
-const STAKE_TOO_LOW_CODE: i32 = -32505;
+const _STAKE_TOO_LOW_CODE: i32 = -32505;
```
dancoombs, line 38: Interesting that this is defined in the spec but never used by the spec tests

crates/rpc/src/types.rs
```diff
@@ -169,7 +170,7 @@ pub struct RpcReputation {
     /// Number of operations included in this interval
     pub ops_included: U256,
     /// Reputation status
-    pub status: ReputationStatus,
+    pub status: Option<ReputationStatus>,
```
dancoombs, line 173: Is there a use case where this is actually optional in an API? Or is there a request type that omits and a response type that includes? We may want to represent that as 2 separate types to have more strict API format.

crates/rpc/src/types.rs
```diff
@@ -178,7 +179,7 @@ impl From<RpcReputation> for Reputation {
             address: rpc_reputation.address,
             ops_seen: rpc_reputation.ops_seen.as_u64(),
             ops_included: rpc_reputation.ops_included.as_u64(),
-            status: rpc_reputation.status,
+            status: rpc_reputation.status.unwrap_or(ReputationStatus::Ok),
```
dancoombs, line 182: Can we remove `status` from our internal `Reputation` object? Just calculate it where we need it based on `ops_seen` and `ops_included`?

crates/sim/src/simulation/tracer.rs
```diff
@@ -71,9 +71,9 @@ pub(crate) struct Phase {
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
-pub(crate) struct StorageAccess {
-    pub(crate) address: Address,
-    pub(crate) slots: Vec<U256>,
+pub(crate) struct AccessInfo {
+    pub(crate) reads: HashMap<U256, String>,
+    pub(crate) writes: HashMap<U256, u32>,
```
dancoombs, line 76: Can you add comments to what the values in these hashmaps represent? Hard to tell from the code

crates/sim/tracer/src/validationTracer.ts
```diff
@@ -233,13 +237,19 @@ type StringSet = Record<string, boolean | undefined>;
...
     step(log, db): void {
       if (!entryPointAddress) {
         entryPointAddress = toHex(log.contract.getAddress());
       }
-      if (log.getGas() < log.getCost()) {
+
+      const opcode = log.op.toString();
+
+      if (log.getGas() < log.getCost() || (
+        opcode === 'SSTORE' && log.getGas() < SSTORE_REQUIRED_GAS
```
dancoombs, line 251: Was this just something we missed before? How come the existing check doesn't catch this?

crates/rpc/src/types.rs
0xfourzerofour: This is more for the fact that the json RPC library we use doe not allow us to omit the field if it is defined like this, but I think it would make more sense to seperate the rpc type with a Reputation input and Reputation Output so that we can calculate on the fly brought up below

crates/sim/tracer/src/validationTracer.ts
0xfourzerofour: Its just something we missed. It is a weird case where even if SSTORE takes less than 2300 gas we still need to  have at least that or it will be an OOG error

crates/sim/tracer/src/validationTracer.ts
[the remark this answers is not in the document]
dancoombs: Can we add a similar comment to our code here?

crates/sim/src/simulation/simulation.rs
```diff
@@ -804,48 +823,81 @@ fn entity_type_needs_stake(
...
+                // special case: account.validateUserOp is allowed to use assoc storage if factory is staked.
+                // [STO-022], [STO-021]
+                && !(entity.address.eq(&sender)
+                    && entity_infos
+                        .factory
+                        .expect("Factory needs to be present and staked")
+                        .is_staked)
```
dancoombs, line 883: This logic looks different than the logic we removed 

`accessed_address == entry_point_address || !is_unstaked_wallet_creation`

what is the reason for the difference?

> special case: account.validateUserOp is allowed to use assoc storage if factory is staked.

This isn't how I read STO-21/STO-22, it seems like this should apply to the paymaster and factory as well? Is this a difference in the spec vs how the reference is implemented?

0xfourzerofour: Yeh our existing implementation was a little hard to understand but this is how it should be implented. STO-022 is just regarding the factory and whether it has an init code

crates/rpc/src/types.rs
0xfourzerofour: This has some pretty annoying consequences, because the status computation requires internal config from the Reputation manager. We could add to the proto so that we can read the status over the wire but that does not seem ideal

crates/sim/src/simulation/simulation.rs
dancoombs: It seems like the code and the description in the ERC don't match though. Why doesn't this apply to the paymaster?

crates/rpc/src/types.rs
dancoombs: Yeah, it would be great to clean up the types here, at least on on the RPC side.

crates/rpc/src/types.rs
dancoombs: Got it, just seems odd to me that the `status` could say `OK` but the `ops_seen`/`ops_included` would indicate otherwise. Any thoughts on how to clean that up?

0xfourzerofour: I am ok adding GetReputationStatus to the op_pool in the proto. I have been playing around with the code and even though the request is over the wire  if running in distributed mode I think it is cleaner than other options

crates/sim/src/simulation/simulation.rs
0xfourzerofour: It matches in my head

Both rules are asking if the account entity (sender) can access the associated storage slots of the contract itself or the associate slot of the sender in another contract.

[STO-021]  The account already exists
- This is checked in the tracer and returned by the AccessedUndeployedContract error

[STO-022]  There is an `initCode` and the `factory` contract is staked
- Checked here on line 879

This is just how I have thought about the rules over the past few days. Do you think it means something else?

dancoombs: I don't see anything that limits this rule to apply only to the account entity. It reads to me like both the paymaster/factory need to be limited as well.

If the factory isn't staked, then no entity can read associated storage during deployment

0xfourzerofour: I think we should ask the 4337 team because I dont think it is clear as the wording makes it ambiguous

https://github.com/eth-infinitism/account-abstraction/blob/develop/eip/EIPS/eip-aa-rules.md?plain=1#L209

0xfourzerofour: https://github.com/eth-infinitism/bundler/blob/ba78f2067e33df399bb7d34a9d1f972834d23c80/packages/validation-manager/src/TracerResultParser.ts#L322

dancoombs: https://github.com/eth-infinitism/bundler/commit/32725c2cc028c32ed7f12aea0e1eb3438d9d8e79

You are correct! The wording can probably be improved on the erc-aa-rules

dancoombs (approved):
🚢 

Great work on this, tough to get all these details ironed out!

PR #536, opened 2023-12-18
feat: change priority fee strategy

dancoombs:


crates/builder/src/bundle_proposer.rs
```diff
@@ -260,28 +239,106 @@ where
...
+    async fn filter_and_simulate(
         &self,
         op: PoolOperation,
         block_hash: H256,
-    ) -> anyhow::Result<(PoolOperation, Result<SimulationSuccess, SimulationError>)> {
+        base_fee: U256,
+        required_op_fees: GasFees,
+    ) -> Option<(PoolOperation, Result<SimulationSuccess, SimulationError>)> {
+        // filter by fees
+        if op.uo.max_fee_per_gas >= required_op_fees.max_fee_per_gas
```
0xfourzerofour, line 256: shouldnt this be <= instead of >=? the previous version was using a filter and returning true or skipped

crates/builder/src/bundle_proposer.rs
```diff
@@ -260,28 +239,106 @@ where
...
+                    reason: Arc::new(format!(
+                        "Failed to calculate required pre-verification gas for op: {e:?}, skipping"
+                    )),
+                },
+            ));
+            e
+        })
+        .ok()?;
+
+        if required_pvg < op.uo.pre_verification_gas {
```
0xfourzerofour, line 296: same here, shouldnt we be skipping if the required is larger than the supplied

crates/provider/src/ethers/provider.rs
```diff
@@ -246,16 +247,8 @@ impl<C: JsonRpcClient + 'static> Provider for EthersProvider<C> {
...
-                let l1_gas = gas_oracle.get_l1_fee(tx).call().await?;
-                Ok(l1_gas)
-            },
-            self.get_base_fee(),
-            self.get_max_priority_fee(),
-        )?;
-
-        Ok(l1_fee / (l2_base_fee + l2_priority_fee))
+        let l1_fee = gas_oracle.get_l1_fee(tx).call().await?;
+        Ok(l1_fee / gas_price)
```
0xfourzerofour, line 251: we should do a checked div here in case of 0

crates/sim/src/gas/oracle.rs
```diff
@@ -36,6 +36,102 @@ pub(crate) trait FeeOracle: Send + Sync + Debug {
...
+        Self { provider, config }
+    }
+}
+
+#[async_trait]
+impl<P> FeeOracle for UsageBasedFeeOracle<P>
+where
+    P: Provider + Debug,
+{
+    async fn estimate_priority_fee(&self) -> Result<U256> {
```
0xfourzerofour, line 95: this is nice

0xfourzerofour (commented):
Overall looks solid just a couple comments I would like to make sure is right before approval

0xfourzerofour (approved):
Ok now it looks good, just fix that unchecked div and LGTM

crates/sim/src/estimation/estimation.rs
```diff
@@ -118,8 +119,17 @@ impl<P: Provider, E: EntryPoint> GasEstimator for GasEstimatorImpl<P, E> {
             .await
             .map_err(anyhow::Error::from)?;
 
-        // Estimate pre verification gas
-        let pre_verification_gas = self.calc_pre_verification_gas(&op).await?;
+        // Estimate pre verification gas at the current fees
+        // If the user provides fees, use them, otherwise use the current bundle fees
+        let (bundle_fees, base_fee) = self.fee_estimator.required_bundle_fees(None).await?;
```
alex-miao, line 124: I think this is being used when the user provides fees so we can put this inside the `else` statement

crates/builder/src/bundle_proposer.rs
dancoombs: Fixed

crates/sim/src/estimation/estimation.rs
dancoombs: Base fee is

0xfourzerofour (approved):

PR #539, opened 2023-12-19
feat(compliance): add debug method for stake status

### LINKED ISSUE #538: add debug method for getStakeStaus
the method `debug_getStakeStatus` is implemented in the reference bundler and is required for 3 of the spec tests so we will need to implement these even if it is not yet defined in the spec definition

0xfourzerofour:
Closes #538 

## Proposed Changes

  - create debug method for stake status to match reference bundler

### DIFF
--- a/crates/pool/src/mempool/mod.rs
+++ b/crates/pool/src/mempool/mod.rs
@@ -93,6 +93,9 @@ pub trait Mempool: Send + Sync + 'static {
 
     /// Overwrites the mempool's reputation for an address
     fn set_reputation(&self, address: Address, ops_seen: u64, ops_included: u64);
+
+    /// Get stake status for address
+    async fn get_stake_status(&self, address: Address) -> MempoolResult<StakeStatus>;
 }
 
 /// Config for the mempool
@@ -128,6 +131,23 @@ pub struct PoolConfig {
     pub throttled_entity_mempool_count: u64,
     /// The maximum number of blocks a user operation with a throttled entity can stay in the mempool
     pub throttled_entity_live_blocks: u64,
+}
+
+/// Stake status structure
+#[derive(Debug, Clone, Copy)]
+pub struct StakeStatus {
+    /// Address is staked
+    pub is_staked: bool,
+    /// Stake information about address
+    pub stake_info: StakeInfo,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct StakeInfo {
+    /// Stake ammount
+    pub stake: u128,
+    /// Unstake delay in seconds
+    pub unstake_delay_sec: u32,
 }
 
 /// Origin of an operation.
--- a/crates/pool/src/mempool/uo_pool.rs
+++ b/crates/pool/src/mempool/uo_pool.rs
@@ -19,6 +19,7 @@ use ethers::{
 };
 use itertools::Itertools;
 use parking_lot::RwLock;
+use rundler_provider::StakeManager;
 use rundler_sim::{Prechecker, Simulator};
 use rundler_types::{Entity, EntityUpdate, EntityUpdateType, UserOperation};
 use rundler_utils::emit::WithEntryPoint;
@@ -30,7 +31,7 @@ use super::{
     error::{MempoolError, MempoolResult},
     pool::PoolInner,
     reputation::{Reputation, ReputationManager, ReputationStatus},
-    Mempool, OperationOrigin, PoolConfig, PoolOperation,
+    Mempool, OperationOrigin, PoolConfig, PoolOperation, StakeInfo, StakeStatus,
 };
 use crate::{
     chain::ChainUpdate,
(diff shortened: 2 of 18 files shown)


crates/pool/src/mempool/mod.rs
```diff
@@ -130,6 +133,25 @@ pub struct PoolConfig {
...
+    pub is_staked: bool,
+    /// Stake information about address
+    pub stake_info: StakeInfo,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct StakeInfo {
+    /// Address of information
+    pub addr: Address,
+    /// Address of information
```
dancoombs, line 149: nit: fix comment

crates/rpc/src/types.rs
```diff
@@ -62,6 +62,35 @@ impl From<Address> for RpcAddress {
...
+#[serde(rename_all = "camelCase")]
+pub struct RpcStakeStatus {
+    pub is_staked: bool,
+    pub stake_info: RpcStakeInfo,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcStakeInfo {
+    pub addr: Address,
```
dancoombs, line 76: Its annoying that they have this here. Why would the address need to be in the response when its in the request?

crates/pool/src/mempool/mod.rs
```diff
@@ -130,6 +133,25 @@ pub struct PoolConfig {
...
+    /// Address is staked
+    pub is_staked: bool,
+    /// Stake information about address
+    pub stake_info: StakeInfo,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct StakeInfo {
+    /// Address of information
+    pub addr: Address,
```
dancoombs, line 148: As I mentioned below, is there a reason why this would need to be in this object? I feel like this would be used to look up this information so its redundant. We could leave this out and just fill it in the RPC response since the RPC struct needs it.

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -433,6 +439,28 @@ where
...
 
+    async fn get_stake_status(&self, address: Address) -> MempoolResult<StakeStatus> {
+        let deposit_info = self.stake_manager.get_deposit_info(address).await?;
+
+        let is_staked = deposit_info
+            .stake
+            .ge(&self.config.sim_settings.min_stake_value)
+            && deposit_info
+                .unstake_delay_sec
+                .ge(&self.config.sim_settings.min_unstake_delay);
```
dancoombs, line 450: Not required, but could be useful to hide this logic in the stake manager itself, and just configure it with these values and have it return the full stake status

crates/pool/src/server/remote/protos.rs
```diff
@@ -199,6 +200,38 @@ impl TryFrom<Reputation> for PoolReputation {
     }
 }
 
+pub const EMPTY_STAKE_INFO_ERROR: &str = "Stake info cannot be empty";
```
dancoombs, line 203: Does this need to be pub?

crates/provider/src/ethers/stake_manager.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use crate::StakeManager;
+
+#[async_trait::async_trait]
+impl<M> StakeManager for IStakeManager<M>
+where
+    M: Middleware + 'static,
+{
+    async fn get_deposit_info(&self, address: Address) -> anyhow::Result<DepositInfo> {
+        let deposit_info: DepositInfo = self.get_deposit_info(address).await?;
```
dancoombs, line 25: I much prefer this to be made explicit when its the same name as the function, so `IStakeManager::get_deposit_info(&self, address)`

crates/provider/src/ethers/stake_manager.rs
```diff
@@ -0,0 +1,29 @@
...
+// You should have received a copy of the GNU General Public License along with Rundler.
+// If not, see https://www.gnu.org/licenses/.
+
+use ethers::{providers::Middleware, types::Address};
+use rundler_types::{contracts::i_stake_manager::IStakeManager, DepositInfo};
+
+use crate::StakeManager;
+
+#[async_trait::async_trait]
+impl<M> StakeManager for IStakeManager<M>
```
dancoombs, line 20: Why did we need this class? Don't we have this function available on the entry point provider already?

crates/rpc/src/debug.rs
```diff
@@ -58,6 +58,14 @@ pub trait DebugApi {
         &self,
         entry_point: Address,
     ) -> RpcResult<Vec<RpcReputationOutput>>;
+
+    /// Dumps the reputations of entities from the given entry point.
```
dancoombs, line 62: nit: fix comment

crates/provider/src/ethers/stake_manager.rs
dancoombs: Looked closer, entry point does not have this.

dancoombs (approved):

dancoombs (approved):

PR #540, opened 2023-12-20
fix(builder): do not send entity update for reverts due to invalid nonce

alex-miao:
## Proposed Changes

  - Do not send to the pool an entity update when an revert is due to an invalid nonce since it can occur without a sender acting maliciously

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -1060,15 +1060,21 @@ impl ProposalContext {
                 SimulationViolation::NotStaked(entity, _, _) => {
                     self.add_entity_update(entity, entity_infos)
                 }
-                SimulationViolation::UnintendedRevertWithMessage(entity_type, _, address) => {
-                    if let Some(entity_address) = address {
-                        self.add_entity_update(
-                            Entity {
-                                kind: entity_type,
-                                address: entity_address,
-                            },
-                            entity_infos,
-                        );
+                SimulationViolation::UnintendedRevertWithMessage(entity_type, message, address) => {
+                    match &message[..4] {
+                        // do not penalize an entity for invalid account nonces, which can occur without malicious intent from the sender
+                        "AA25" => {}
+                        _ => {
+                            if let Some(entity_address) = address {
+                                self.add_entity_update(
+                                    Entity {
+                                        kind: entity_type,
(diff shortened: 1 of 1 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -50,6 +50,8 @@ use crate::emit::{BuilderEvent, OpRejectionReason, SkipReason};
 const TIME_RANGE_BUFFER: Duration = Duration::from_secs(60);
 /// Extra buffer percent to add on the bundle transaction gas estimate to be sure it will be enough
 const BUNDLE_TRANSACTION_GAS_OVERHEAD_PERCENT: u64 = 5;
+/// Error message within an UnintendedRevertWithMessage simulation violation for an invalid nonce
+const INVALID_NONCE_REVERT_MESSAGE: &str = "AA25 invalid account nonce";
```
dancoombs, line 54: Do they reuse AA25 anywhere? Else we can just match on that (we do that elsewhere)

0xfourzerofour: looks like its its the only place they use it

alex-miao: Don't think we reuse, so sounds good

0xfourzerofour: @alex-miao I think this mean you can remove the "invalid account nonce" part and just match on the error code

crates/builder/src/bundle_proposer.rs
```diff
@@ -1060,15 +1060,18 @@ impl ProposalContext {
...
-                        self.add_entity_update(
-                            Entity {
-                                kind: entity_type,
-                                address: entity_address,
-                            },
-                            entity_infos,
-                        );
+                SimulationViolation::UnintendedRevertWithMessage(entity_type, message, address) => {
+                    // do not penalize an entity for invalid account nonces, which can occur without malicious intent from the sender
+                    if message != "AA25 invalid account nonce" {
```
dancoombs, line 1065: Sorry, by my earlier comment I meant to do something like this: https://github.com/alchemyplatform/rundler/blob/3df2215f496371959161ef796f865b4cdfad1bf8/crates/builder/src/bundle_proposer.rs#L560

0xfourzerofour (approved):

PR #546, opened 2024-01-10
feat: use usage based oracle on polygon

dancoombs:


crates/sim/src/gas/oracle.rs
```diff
@@ -10,6 +10,7 @@
 //
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
+#![allow(dead_code)]
```
dancoombs, line 13: This will be removed when we move the oracle settings to ChainSpec, where we can choose different fee oracles from configuration

alex-miao (approved):

PR #548, opened 2024-01-10
wip

dancoombs:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -


bin/tools/src/bin/load_chain_spec.rs
```diff
@@ -0,0 +1,88 @@
...
+            let file = File::open(file).expect("should open file");
+            serde_yaml::from_reader(file).expect("should parse yaml")
+        }
+        None => OptionalChainSpec::default(),
+    };
+
+    let env_spec = match envy::prefixed("CHAIN_").from_env::<OptionalChainSpec>() {
+        Ok(config) => config,
+        Err(error) => panic!("{:#?}", error),
+    };
```
dphilipson, line 51: Is this match different/better than calling `.unwrap()`?

crates/types/src/chain_spec.rs
```diff
@@ -0,0 +1,247 @@
...
+        $vis struct $struct_name {
+            $( #[$doc] $fv $field: $ty ),*
+        }
+
+        paste! {
+            $(#[$attr])*
+            #[derive(Deserialize, Default)]
+            $vis struct [<Optional $struct_name>] {
+                /// The base chain spec to use
+                pub base: Option<String>,
```
dphilipson, line 35: Is this supposed to be here? Is it different from the other fields?

dphilipson (commented):
I love it.

bin/tools/src/bin/load_chain_spec.rs
dphilipson: Also don't forget that you can write stuff like this and the printlns as
```rust
panic!("{error:#?}")
```

dancoombs: Lol no it is not, wasn't very clean writing this bin

crates/types/src/chain_spec.rs
dancoombs: Yeah the idea was that during the optional phase you can define a `base` that merges all configuration from that base configuration into yours, useful for testnets to inherit the config of the mainnet. This field goes away in the actual chain spec config.

PR #549, opened 2024-01-12
feat: add chain spec, use it for gas, add configs

### LINKED ISSUE #486: [config] `ChainSpec` based configuration.
**Describe the feature**
Rundler should contain a `ChainSpec` struct that contains all of the specific configuration required to setup a new chain. Rundler should ship with a set of known chains and their hardcoded chain spec values, as well as with a pure default chain spec.

When using Rundler on a known chain, the chain ID is supplied and the chain spec can be found/used. When using Rundler on an unknown chain, the user would be required to supply their own chain spec overrides, with everything else using the pure defaults.

A similar implementation is used in Lighthouse

`ChainSpec`: https://github.com/sigp/lighthouse/blob/441fc1691b69f9edc4bbdc6665f3efab16265c9b/consensus/types/src/chain_spec.rs#L33
Known chains: https://github.com/sigp/lighthouse/tree/441fc1691b69f9edc4bbdc6665f3efab16265c9b/common/eth2_network_config/built_in_network_configs

dancoombs:
Closes #486 

## Proposed Changes

  - Add a chain spec object
  - Replace all usage of ethers-rs Chain from our code, use chain spec instead
  - Add all the known chain configurations as hardcoded


bin/rundler/src/cli/chain_spec.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+    let c = config_builder
+        .add_source(Environment::with_prefix("CHAIN"))
+        .build()
+        .expect("should build config");
+
+    let id = c.get::<u64>("id").ok();
+    if let Some(id) = id {
+        if id == 0 {
+            panic!("chain id must be non-zero");
```
0xfourzerofour, line 75: These errors could potentially be put into an error enum to clean things up a bit

bin/rundler/src/cli/chain_spec.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    } else {
+        panic!("chain id must be defined");
+    }
+
+    c.try_deserialize().expect("should deserialize config")
+}
+
+macro_rules! define_hardcoded_chain_specs {
```
0xfourzerofour, line 84: This is sick

bin/rundler/src/cli/node/mod.rs
```diff
@@ -43,16 +44,25 @@ pub struct NodeCliArgs {
...
-    let pool_task_args = pool_args.to_args(&common_args, None).await?;
-    let builder_task_args = builder_args.to_args(&common_args, None).await?;
+    let pool_task_args = pool_args
+        .to_args(chain_spec.clone(), &common_args, None)
+        .await?;
+    let builder_task_args = builder_args
+        .to_args(chain_spec.clone(), &common_args, None)
+        .await?;
     let rpc_task_args = rpc_args.to_args(
+        chain_spec.clone(),
```
0xfourzerofour, line 65: does this third copy need to be cloned?

0xfourzerofour (approved):
Overall an awesome implementation that will clean up some of the scattered chain logic, good job on this. The only think that could be changed is the cloning of the chainspec structure in a few places that need to reference, we could use a thread safe immutable reference to save memory but in the grand scheme of things its not a huge overhead

bin/rundler/src/cli/chain_spec.rs
dancoombs: I actually think its okay to have these `panic!` macros in the `bin` code. It only gets called on startup so we'd want this to just crash anyway

bin/rundler/src/cli/node/mod.rs
dancoombs: Nope, fixed. Weird that the linter didn't catch this.

bin/rundler/src/cli/mod.rs
```diff
@@ -91,26 +95,24 @@ enum Command {
...
+
+    /// Chain spec file path
     #[arg(
-        long = "chain_id",
-        name = "chain_id",
-        env = "CHAIN_ID",
-        default_value = "1337",
+        long = "chain_spec",
+        name = "chain_spec",
+        env = "CHAIN_SPEC",
```
alex-miao, line 112: Does this conflict with the code in `chain_spec.rs` where you are grabbing all env variables that are prefixed by `CHAIN`

dancoombs: Shouldn't be an issue unless we try to use `spec` in `ChainSpec`

alex-miao (approved):

PR #555, opened 2024-01-16
feat(rpc): include pending UOs in eth_getUserOperationByHash

### LINKED ISSUE #529: [rpc] Return pending UOs in `eth_getUserOperationByHash`
**Describe the feature**
See https://github.com/eth-infinitism/account-abstraction/pull/382

dancoombs:
Closes #529 

## Proposed Changes

  - Add endpoint to pool to get a pending user operation by hash
  - Use that endpoint in `eth_getUserOperationByHash` to return pending UOs
  
## TODO

- [x] Add unit tests to pool
- [x] Add unit tests to RPC
- [x] Integration test

### DIFF
--- a/crates/pool/src/mempool/mod.rs
+++ b/crates/pool/src/mempool/mod.rs
@@ -80,6 +80,9 @@ pub trait Mempool: Send + Sync + 'static {
     /// Returns the all operations from the pool up to a max size
     fn all_operations(&self, max: usize) -> Vec<Arc<PoolOperation>>;
 
+    /// Looks up a user operation by hash, returns None if not found
+    fn get_user_operation_by_hash(&self, hash: H256) -> Option<Arc<PoolOperation>>;
+
     /// Debug methods
 
     /// Clears the mempool
@@ -168,6 +171,8 @@ pub enum OperationOrigin {
 pub struct PoolOperation {
     /// The user operation stored in the pool
     pub uo: UserOperation,
+    /// The entry point address for this operation
+    pub entry_point: Address,
     /// The aggregator address for this operation, if any.
     pub aggregator: Option<Address>,
     /// The valid time range for this operation.
@@ -264,6 +269,7 @@ mod tests {
                 init_code: factory.as_fixed_bytes().into(),
                 ..Default::default()
             },
+            entry_point: Address::random(),
             aggregator: Some(aggregator),
             valid_time_range: ValidTimeRange::all_time(),
             expected_code_hash: H256::random(),
--- a/crates/pool/src/server/local.rs
+++ b/crates/pool/src/server/local.rs
@@ -150,6 +150,15 @@ impl PoolServer for LocalPoolHandle {
         }
     }
 
+    async fn get_op_by_hash(&self, hash: H256) -> PoolResult<Option<PoolOperation>> {
+        let req = ServerRequestKind::GetOpByHash { hash };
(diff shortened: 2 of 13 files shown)


crates/pool/src/mempool/mod.rs
```diff
@@ -80,6 +80,9 @@ pub trait Mempool: Send + Sync + 'static {
     /// Returns the all operations from the pool up to a max size
     fn all_operations(&self, max: usize) -> Vec<Arc<PoolOperation>>;
 
+    /// Looks up a user operation by hash, returns None if not found
+    fn get_user_operation_by_hash(&self, hash: H256) -> Option<Arc<PoolOperation>>;
```
0xfourzerofour, line 84: just a question, but whats the need to return an atomic reference counter?

crates/pool/src/server/local.rs
```diff
@@ -326,6 +335,15 @@ where
             .collect())
     }
 
+    fn get_op_by_hash(&self, hash: H256) -> PoolResult<Option<PoolOperation>> {
```
0xfourzerofour, line 338: The op hash is hashed using the entypoint as one of the fields so it might make sense to use hash and entrypoint as inputs instead of looping though mempools. I think instead we should use `mempool.get(entrypoint)`

0xfourzerofour: actually I guess that is not ideal because the spec does not include the entrypoint even if I think it makes more sense

crates/pool/src/server/remote/server.rs
```diff
@@ -181,6 +182,33 @@ impl OpPool for OpPoolImpl {
         Ok(Response::new(resp))
     }
 
+    async fn get_op_by_hash(
+        &self,
+        request: Request<GetOpByHashRequest>,
+    ) -> Result<Response<GetOpByHashResponse>> {
+        let req = request.into_inner();
+
+        if req.hash.len() != 32 {
```
0xfourzerofour, line 191: Pretty sure this logic can be done on the ethers side automattically with their implentations

https://github.com/gakonst/ethers-rs/blob/88095ba47eb6a3507f0db1767353b387b27a6e98/ethers-core/src/abi/tokens.rs#L187

https://github.com/gakonst/ethers-rs/blob/88095ba47eb6a3507f0db1767353b387b27a6e98/ethers-core/src/utils/mod.rs#L521

which already checks for the length

0xfourzerofour (approved):
Looks good just a couple comments

crates/pool/src/mempool/mod.rs
dancoombs: So we don't need to copy the `PoolOperation` (which can hold a large amount of data) around. Needs to be atomic since the server can access it from many threads.

It gets copied out of the `Arc` before its returned across the server boundary

crates/pool/src/server/local.rs
dancoombs: Yeah we don't know the entry point because its not passed in as part of the RPC request. So we have to search each mempool

crates/pool/src/server/remote/server.rs
dancoombs: We actually have helpers for this too, we found those interfaces hard to use. I'll switch to our `from_bytes` func

0xfourzerofour (approved):

PR #557, opened 2024-01-16
feat(sim): allow mempool configs to modify simulation results

### LINKED ISSUE #354: pool: Still track reputation of non-stake-allowed entities.

alex-miao:
Fixes #354

## Proposed Changes

  - Mark allowlisted unstaked entities as staked

### DIFF
--- a/crates/sim/src/simulation/mempool.rs
+++ b/crates/sim/src/simulation/mempool.rs
@@ -26,12 +26,12 @@ use crate::simulation::SimulationViolation;
 #[derive(Debug, Clone, Deserialize, Default)]
 pub struct MempoolConfig {
     /// Allowlist to match violations against.
-    allowlist: Vec<AllowlistEntry>,
+    pub(crate) allowlist: Vec<AllowlistEntry>,
 }
 
 /// The entity allowed by an allowlist entry.
 #[derive(Debug, Copy, Clone)]
-enum AllowEntity {
+pub(crate) enum AllowEntity {
     /// Any entity is allowed.
     Any,
     /// Entity of a specific type is allowed.
@@ -67,9 +67,9 @@ impl AllowEntity {
 }
 
 /// An allowlist rule.
-#[derive(Debug, Clone, Deserialize)]
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
 #[serde(tag = "rule", rename_all = "camelCase")]
-enum AllowRule {
+pub(crate) enum AllowRule {
     /// Allowlist a forbidden opcode on a contract.
     ForbiddenOpcode { contract: Address, opcode: Opcode },
     /// Allowlist a forbidden precompile by its address on a contract
@@ -88,13 +88,14 @@ enum AllowRule {
 /// An allowlist entry
 #[serde_as]
 #[derive(Debug, Clone, Deserialize)]
-struct AllowlistEntry {
+
+pub(crate) struct AllowlistEntry {
     /// The entity allowed by this entry.
     #[serde_as(as = "DisplayFromStr")]
-    entity: AllowEntity,
+    pub(crate) entity: AllowEntity,
     /// The rule allowed by this entry.
     #[serde(flatten)]
-    rule: AllowRule,
+    pub(crate) rule: AllowRule,
 }
 
 impl AllowlistEntry {
--- a/crates/sim/src/simulation/simulation.rs
+++ b/crates/sim/src/simulation/simulation.rs
@@ -34,7 +34,7 @@ use rundler_types::{
 };
 
 use super::{
-    mempool::{match_mempools, MempoolConfig, MempoolMatchResult},
+    mempool::{match_mempools, AllowEntity, AllowRule, MempoolConfig, MempoolMatchResult},
     tracer::{
         parse_combined_tracer_str, AccessInfo, AssociatedSlotsByAddress, SimulateValidationTracer,
         SimulationTracerOutput,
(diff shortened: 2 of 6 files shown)


crates/sim/src/simulation/mempool.rs
```diff
@@ -197,6 +197,21 @@ pub(crate) fn match_mempools(
     MempoolMatchResult::Matches(candidate_pools)
 }
 
+/// Get a list of entities that are allowed to act as staked entities despite being unstaked
+pub(crate) fn not_staked_allowlist(mempools: &HashMap<H256, MempoolConfig>) -> Vec<Address> {
+    let mut allowed_addresses = Vec::new();
+    for config in mempools.values() {
+        for entry in config.allowlist.clone() {
```
dancoombs, line 204: Did you have to `.clone()` here? Should just be able to take a reference and then clone in the `push(...)`

crates/sim/src/simulation/simulation.rs
```diff
@@ -598,11 +601,44 @@ where
...
     }
 }
 
+impl<P, T> SimulatorImpl<P, T>
+where
+    P: Provider,
+    T: SimulateValidationTracer,
+{
+    fn modify_entity_infos(&self, entity_infos: EntityInfos) -> EntityInfos {
+        let allow_unstaked = not_staked_allowlist(&self.mempool_configs);
```
dancoombs, line 615: This value shouldn't change right? Can we just calculate it once and store it on the simulator?

crates/sim/src/simulation/mempool.rs
```diff
@@ -197,6 +197,21 @@ pub(crate) fn match_mempools(
     MempoolMatchResult::Matches(candidate_pools)
 }
 
+/// Get a list of entities that are allowed to act as staked entities despite being unstaked
+pub(crate) fn not_staked_allowlist(mempools: &HashMap<H256, MempoolConfig>) -> Vec<Address> {
```
dancoombs, line 201: Based on how we are using this, I think a more descriptive name would be `stake_override_list`

crates/sim/src/simulation/simulation.rs
```diff
@@ -598,11 +601,44 @@ where
...
+                .paymaster
+                .map(|paymaster| modify_entity_info(paymaster, &allow_unstaked)),
+            aggregator: entity_infos
+                .aggregator
+                .map(|aggregator| modify_entity_info(aggregator, &allow_unstaked)),
+        }
+    }
+}
+
+fn modify_entity_info(entity_info: EntityInfo, allow_unstaked: &[Address]) -> EntityInfo {
```
dancoombs, line 631: Maybe call this `override_staked`

crates/sim/src/simulation/simulation.rs
```diff
@@ -598,11 +601,44 @@ where
...
+                .map(|factory| modify_entity_info(factory, &allow_unstaked)),
+            sender: modify_entity_info(entity_infos.sender, &allow_unstaked),
+            paymaster: entity_infos
+                .paymaster
+                .map(|paymaster| modify_entity_info(paymaster, &allow_unstaked)),
+            aggregator: entity_infos
+                .aggregator
+                .map(|aggregator| modify_entity_info(aggregator, &allow_unstaked)),
+        }
+    }
```
dancoombs, line 628: I think it would be cleaner to add this as a function on `EntityInfos` something like `EntityInfos::override_staked(&stake_overrides)`

crates/sim/src/simulation/mempool.rs
```diff
@@ -197,6 +197,21 @@ pub(crate) fn match_mempools(
...
+    for config in mempools.values() {
+        for entry in config.allowlist.clone() {
+            if entry.rule == AllowRule::NotStaked {
+                if let AllowEntity::Address(address) = entry.entity {
+                    allowed_addresses.push(address)
+                }
+            }
+        }
+    }
+    allowed_addresses
```
dancoombs, line 212: Lets return a `HashSet<>` here for quick lookups

crates/sim/src/simulation/simulation.rs
```diff
@@ -598,11 +601,44 @@ where
             accessed_addresses,
             expected_storage: tracer_out.expected_storage,
             requires_post_op: !paymaster_context.is_empty(),
-            entity_infos: context.entity_infos,
+            entity_infos: self.modify_entity_infos(context.entity_infos),
```
dancoombs, line 604: Lets make this step more explicit above, add a comment about what is happening. It seems very hidden here as part of the return value.

crates/sim/src/simulation/mempool.rs
dancoombs: I guess it may be faster to use a `Vec<>` here given that the expected amount of entries is very small, but I think a hashset better signals how this is going to be used.

crates/sim/src/simulation/simulation.rs
alex-miao: Made this a mutable method on `EntityInfos`

dancoombs (approved):
Nice! Good work on the updates, a lot cleaner!

🚢

0xfourzerofour (approved):
Good job, looks good to me

PR #559, opened 2024-01-17
fix(builder): try underpriced bundle txns again

dancoombs:
Relates to #550

See https://github.com/alchemyplatform/rundler/pull/553

### DIFF
--- a/crates/builder/src/sender/flashbots.rs
+++ b/crates/builder/src/sender/flashbots.rs
@@ -21,7 +21,7 @@ use std::{
     task::{Context as TaskContext, Poll},
 };
 
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, Context};
 use ethers::{
     middleware::SignerMiddleware,
     providers::{interval, JsonRpcClient, Middleware, Provider},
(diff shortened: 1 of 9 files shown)


crates/builder/src/sender/flashbots.rs
```diff
@@ -122,7 +120,7 @@ where
     C: JsonRpcClient + 'static,
     S: Signer + 'static,
 {
-    pub(crate) fn new(provider: Arc<Provider<C>>, signer: S) -> Result<Self> {
+    pub(crate) fn new(provider: Arc<Provider<C>>, signer: S) -> anyhow::Result<Self> {
```
0xfourzerofour, line 123: nit: any way that we can keep things consistent with the typed errors? Some implement the new error type and others stay with anyhow::Result?

crates/utils/src/math.rs
```diff
@@ -23,10 +23,28 @@ where
     n * (100 + percent) / 100
 }
 
+/// Increases a number by a percentage, rounding up
+pub fn increase_by_percent_ceil<T>(n: T, percent: u64) -> T
```
0xfourzerofour, line 27: Not a requested change but @alex-miao and I spoke previously about using basis points instead of percent throughout the application which could be a good idea id we want to make it as specific as possible

0xfourzerofour (approved):
Looks good, overall solid just a couple small comments

crates/builder/src/sender/flashbots.rs
dancoombs: Good call, missed this one.

crates/utils/src/math.rs
dancoombs: Yeah this interface is going to bite us if we want to use a fractional percent

0xfourzerofour (approved):

PR #560, opened 2024-01-17
feat(pool): remove expired UOs from pool each block

### LINKED ISSUE #502: [pool] Time based evictions
**Describe the feature**
We should be evicting user operations from the pool that have squatted for over a configurable amount of time OR that are past their `valid_until` time that was captured during simulation.

dancoombs:
Closes #502 

## Proposed Changes
Evict UOs once the last mined block's timestamp has passed their valid_until field

## TODO
- [x] Integration test this, submit a UO with a short expiry and fees that are too low

### DIFF
--- a/crates/pool/src/mempool/pool.rs
+++ b/crates/pool/src/mempool/pool.rs
@@ -22,7 +22,7 @@ use ethers::{
     abi::Address,
     types::{H256, U256},
 };
-use rundler_types::{Entity, UserOperation, UserOperationId};
+use rundler_types::{Entity, Timestamp, UserOperation, UserOperationId};
 use rundler_utils::math;
 use tracing::info;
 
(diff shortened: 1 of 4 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -145,6 +145,25 @@ impl PoolInner {
         self.best.clone().into_iter().map(|v| v.po)
     }
 
+    /// Removes all operations using the given entity, returning the hashes of the removed operations.
+    ///
+    /// NOTE: This method is O(n) where n is the number of operations in the pool.
```
0xfourzerofour, line 150: I wonder if there is a way we can do this using a sorted adjacency list type of data structure to remove the need for O(n). I know there are ways but we can implement that if we see a performance degradation but good to flag

0xfourzerofour (approved):

crates/pool/src/mempool/pool.rs
0xfourzerofour: overall looks sweet tho

dancoombs: Do you have a drop in data structure in mind? Need to maintain linear time removal from the list.

PR #568, opened 2024-01-25
feat(rpc): add debug_bundler_clearMempool endpoint

alex-miao:
## Proposed Changes

  - Add the `debug_bundler_clearMempool` RPC method which is used in the spec test: https://github.com/alchemyplatform/bundler-spec-tests/blob/12bf62752d4af78ed383e6ec4c1a610ebaea1787/tests/utils.py#L184. The spec test uses the method to clear the mempool but not the reputations, which is why we need a different endpoint than `debug_bundler_clearState`

### DIFF
--- a/crates/pool/proto/op_pool/op_pool.proto
+++ b/crates/pool/proto/op_pool/op_pool.proto
@@ -139,9 +139,6 @@ service OpPool {
   // Clears the bundler mempool and reputation data of paymasters/accounts/factories/aggregators
   rpc DebugClearState (DebugClearStateRequest) returns (DebugClearStateResponse);
 
-  // Clears the bundler mempool but not reputation data
-  rpc DebugClearMempool (DebugClearMempoolRequest) returns (DebugClearMempoolResponse);
-
   // Dumps the current UserOperations mempool
   rpc DebugDumpMempool (DebugDumpMempoolRequest) returns (DebugDumpMempoolResponse);
 
@@ -293,7 +290,10 @@ message UpdateEntitiesResponse {
 }
 message UpdateEntitiesSuccess {}
 
-message DebugClearStateRequest {}
+message DebugClearStateRequest {
+  bool clear_mempool = 1;
+  bool clear_reputation = 2;
+}
 message DebugClearStateResponse {
   oneof result {
     DebugClearStateSuccess success = 1;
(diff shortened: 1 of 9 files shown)


crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -299,6 +302,15 @@ message DebugClearStateResponse {
...
 message DebugClearStateSuccess {}
 
+message DebugClearMempoolRequest {}
+message DebugClearMempoolResponse {
+  oneof result {
+    DebugClearMempoolSuccess success = 1;
+    MempoolError failure = 2;
+  }
+}
+message DebugClearMempoolSuccess {}
```
dancoombs, line 312: Lets remove this message and consolidate into `DebugClearState`

```suggestion
message DebugClearStateRequest {
  bool clear_mempool = 1;
  bool clear_reputation = 2;
}
```

dancoombs:
This looks like it includes a bunch of code from your other PR, let us know once its cleaned up so we can see the diff clearly

alex-miao:
> This looks like it includes a bunch of code from your other PR, let us know once its cleaned up so we can see the diff clearly

Got messed up from rebasing the other branch with main, fixed now.

crates/pool/proto/op_pool/op_pool.proto
alex-miao: Looks like the buf linter doesn't like combining these two, it enforces that each RPC needs a distinct request and response and also enforces how the request and response are named.

alex-miao: Did combine the two for the `clear_state` method for `uo_pool.rs` though.

dancoombs: Not sure I understand. The suggestion is to use one RPC endpoint `DebugClearState` with one request type `DebugClearStateRequest` that has args describing which state to clear.

dancoombs: The code suggestion above wasn't meant to be taken verbatim, will require more changes

alex-miao: Ok fixed it

crates/pool/src/server/remote/server.rs
```diff
@@ -285,6 +290,24 @@ impl OpPool for OpPoolImpl {
         Ok(Response::new(resp))
     }
 
+    // async fn debug_clear_mempool(
```
dancoombs, line 293: nit: remove

dancoombs (commented):
LGTM

One last nit

dancoombs (approved):
:shipit:

0xfourzerofour (approved):
This looks good to me, I am going to use some of this logic for the admin methods that I am going to fix

PR #569, opened 2024-01-26
feat(compliance): change error code and add unstaked keyword to error message

0xfourzerofour:
There is a test in the spec tests that checks to make sure the reputation status is in the error message. For one of the cases our bunder does not pass as we need the "unstaked" keyword in the error message and an addition change of error code to the stake too low error code.

https://github.com/eth-infinitism/bundler-spec-tests/blob/2b396495ad4bb526c0a0ad39d3795068a142cd2d/tests/single/bundle/test_bundle.py#L233

## Proposed Changes

  - Update error code
  - Add "unstaked" to error


dancoombs (approved):
LGTM

Issue #543, opened 2023-12-21
[pool] Add a flag to turn off reputation and paymaster balance checks for local mempool


dancoombs:
**Describe the feature**
This will allow us to track these reputations, but not ban, while we are testing our implementation

Issue #528, opened 2023-12-07
Make admin namespace and add paymaster clear method


0xfourzerofour:
Make admin namespace and add paymaster clear method.

This can clear the state of a paymaster tracker either globally or at a per paymaster level

dancoombs:
Need similar endpoints for the reputation system (there are some already in debug_)

PR #570, opened 2024-01-26
feat(admin): create admin namespace and add paymaster clear / toggle for reputation

0xfourzerofour:
[Closes] #543 #528

## Proposed Changes

  - Create 3 new methods `admin_togglePaymasterTracker`, `admin_toggleReputationTracker` and `admin_clearPaymasterTrackerState`

While working on this I realized how much I do not like the current reputation manager setup and would like to refactor quickly before this ticket is merged. Currently the strange seperation beween the locked pool paymaster logic and the locked reputation makes things inconsistent.

I started this feature by separating the toggle and clear functions per datastructure like reputation and paymaster however I think it would be better to have 2 admin methods.

- `admin_clearState` in which there is an enum in the params to choose what state to clear, and if ALL is specified we can clear the pool state. This creatly reduces the amount of boiler plate code needed to get this feature working well.
- `admin_toggleTracking` in which there is an enum in the params to choose what state to clear, and if ALL is specified we can toggle the pool tracking of each option `paymaster` or `reputation`. This also reduces the amount of boiler plate code needed.

I also would like to move the reputation state within the locked pool state to keep things consistent.


crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -159,6 +159,15 @@ service OpPool {
...
   rpc SubscribeNewHeads(SubscribeNewHeadsRequest) returns (stream SubscribeNewHeadsResponse);
+
+  // Toggles the operation of the paymaster tracker for a given pool
+  rpc AdminTogglePaymasterTracker(AdminTogglePaymasterTrackerRequest) returns (AdminTogglePaymasterTrackerResponse);
+
+  // Toggles the operation of the reputation tracker for a given pool
+  rpc AdminToggleReputationTracker(AdminToggleReputationTrackerRequest) returns (AdminToggleReputationTrackerResponse);
+
+  // Clears the current state of the paymaster tracker
+  rpc AdminClearPaymasterTrackerState(AdminClearPaymasterTrackerStateRequest) returns (AdminClearPaymasterTrackerStateResponse);
```
alex-miao, line 170: We can potentially merge this with `DebugClearState`, I added some request params in this PR that we can potentially expand: https://github.com/alchemyplatform/rundler/pull/568

bin/rundler/src/cli/pool.rs
```diff
@@ -117,6 +117,22 @@ pub struct PoolArgs {
...
+        name = "pool.paymaster_tracking_enabled",
+        env = "POOL_PAYMASTER_TRACKING_ENABLED",
+        default_value = "true"
+    )]
+    pub paymaster_tracking_enabled: bool,
+
+    #[arg(
+        long = "pool.reputation_tracking_enabled",
+        name = "pool.reputation_tracking_enabled",
+        env = "POOL_reputation_TRACKING_ENABLED",
```
dancoombs, line 132: nit: fix capitalization

crates/pool/proto/op_pool/op_pool.proto
0xfourzerofour: Yeh I think that would be good but debug is a standardized namespace so its best if we have the custom stuff in the admin namespace

crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -159,6 +159,15 @@ service OpPool {
   // Streaming API to subscribe to be updated upon a new block being added to (or reorged onto)
   // the chain. 
   rpc SubscribeNewHeads(SubscribeNewHeadsRequest) returns (stream SubscribeNewHeadsResponse);
+
+  // Toggles the operation of the paymaster tracker for a given pool
+  rpc AdminTogglePaymasterTracker(AdminTogglePaymasterTrackerRequest) returns (AdminTogglePaymasterTrackerResponse);
+
+  // Toggles the operation of the reputation tracker for a given pool
+  rpc AdminToggleReputationTracker(AdminToggleReputationTrackerRequest) returns (AdminToggleReputationTrackerResponse);
```
dancoombs, line 167: I'm not a fan of the "Toggle" model. I'd rather the user just pass a boolean in to set it to whichever state they want.

I think we can leave this as 2 methods for now, but I have a feeling we're going to end up with a bunch of these. We may want to come up with a model where we can have a single method to control these "flags".

crates/pool/proto/op_pool/op_pool.proto
dancoombs: Agreed we should merge these

dancoombs: Will still need its on `admin_*` RPC, but we can use the same method on the pool interface for clearing state

crates/pool/proto/op_pool/op_pool.proto
0xfourzerofour: Yeh if you read the description that is what I have said I would like to do which is why I left this as a draft

0xfourzerofour: Not with the toggle comment but combining to a single method

crates/pool/src/mempool/paymaster.rs
```diff
@@ -138,7 +145,8 @@ impl PaymasterTracker {
         let id = po.uo.id();
         let max_op_cost = po.uo.max_gas_cost();
 
-        if paymaster_metadata.pending_balance.lt(&max_op_cost) {
+        // Only return an error if tracking is enabled
+        if paymaster_metadata.pending_balance.lt(&max_op_cost) && self.tracker_enabled {
```
dancoombs, line 149: Something that confused me about the naming conventions here. We're still actually preforming the tracking, correct? We're just not allowing the tracking to cause UOs to be rejected?

crates/pool/src/mempool/reputation.rs
```diff
@@ -217,11 +225,20 @@ impl Default for ReputationParams {
             inclusion_rate_factor: 10,
             throttling_slack: 10,
             ban_slack: 50,
+            tracking_enabled: true,
         }
     }
 }
 
 impl ReputationParams {
+    pub(crate) fn new_with_config(tracking_enabled: bool) -> Self {
```
dancoombs, line 234: nit: could this just be `new(tracking_enabled: bool)`?

crates/rpc/src/debug.rs
```diff
@@ -104,7 +104,7 @@ where
     async fn bundler_clear_mempool(&self) -> RpcResult<String> {
         let _ = self
             .pool
-            .debug_clear_state(true, false)
+            .debug_clear_state(true, true, false)
```
dancoombs, line 107: Double checking this, we want clear_mempool to clear paymaster tracking? It seems like we should have the `clear_mempool` bool just wipe out the pending paymaster balances, without actually wiping the whole paymaster tracker.

dancoombs (commented):
Couple small things, mostly LGTM

dancoombs:
Related: https://github.com/alchemyplatform/rundler/issues/573

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: Yeh the naming was something I was not too sure about either. Maybe like `only_tracking` or `blocking_enabled` would be more fit for purpose

crates/rpc/src/debug.rs
0xfourzerofour: And also moving the reputation into the same level as the pool so that we do not need that interface

crates/pool/src/mempool/reputation.rs
0xfourzerofour: yeh, I changed it

crates/rpc/src/debug.rs
dancoombs: So we want to leave this as `true` for clearing the paymaster tracking?

dancoombs (approved):
LGTM, just want to make sure that flag is set correctly.

crates/rpc/src/debug.rs
0xfourzerofour: yeh I think leave it for true if we want to clear the mempool as the paymaster tracking relies on operation id's that are in the mempool

PR #572, opened 2024-01-29
feat(paymaster): add paymaster withdrawal logic to tracker

### LINKED ISSUE #561: Use chain to track paymaster withdrawals
In the current state the chain updates only look for additions to the paymaster stake without looking to see if any withdrawals have been confirmed

0xfourzerofour:
Closes #561 

## Proposed Changes

  - Use the same logic as the entity deposits but look for withdrawal logs instead

### DIFF
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -531,6 +531,7 @@ impl<P: Provider> Chain<P> {
         self.blocks.get((number - earliest_number) as usize)
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn new_update(
         &self,
         reorg_depth: u64,
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -819,7 +819,12 @@ mod tests {
         controller.set_blocks(vec![
             MockBlock::new(hash(0), vec![hash(100)], vec![], vec![]),
             MockBlock::new(hash(1), vec![hash(101)], vec![], vec![]),
-            MockBlock::new(hash(2), vec![hash(102)], vec![Address::zero()], vec![]),
+            MockBlock::new(
+                hash(2),
+                vec![hash(102)],
+                vec![Address::zero()],
+                vec![addr(1)],
+            ),
         ]);
         chain.sync_to_block(controller.get_head()).await.unwrap();
         {
@@ -829,7 +834,7 @@ mod tests {
             blocks.extend([
                 MockBlock::new(hash(12), vec![hash(112)], vec![], vec![]),
                 MockBlock::new(hash(13), vec![hash(113)], vec![], vec![]),
-                MockBlock::new(hash(14), vec![hash(114)], vec![], vec![]),
+                MockBlock::new(hash(14), vec![hash(114)], vec![], vec![addr(3)]),
             ]);
         }
         let update = chain.sync_to_block(controller.get_head()).await.unwrap();
@@ -845,8 +850,8 @@ mod tests {
                 unmined_ops: vec![fake_mined_op(102)],
                 entity_deposits: vec![],
                 unmined_entity_deposits: vec![fake_mined_deposit(Address::zero(), 0.into())],
-                entity_withdrawals: vec![],
-                unmined_entity_withdrawals: vec![],
+                entity_withdrawals: vec![fake_mined_deposit(addr(3), 0.into())],
+                unmined_entity_withdrawals: vec![fake_mined_deposit(addr(1), 0.into())],
                 reorg_larger_than_history: false,
             }
         );
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -94,7 +94,7 @@ pub struct BalanceUpdate {
     pub address: Address,
     pub entrypoint: Address,
     pub amount: U256,
-    pub is_deposit: bool,
+    pub is_addition: bool,
 }
 
 impl MinedOp {
@@ -450,7 +450,8 @@ impl<P: Provider> Chain<P> {
     }
 
     fn load_entity_balance_updates(&self, logs: &Vec<Log>) -> Vec<BalanceUpdate> {
-        let mut deposits = vec![];
+        let mut balance_updates = vec![];
+
         for log in logs {
             let entrypoint = log.address;
             if let Ok(event) = contract::parse_log::<DepositedFilter>(log.clone()) {
@@ -458,10 +459,10 @@ impl<P: Provider> Chain<P> {
                     entrypoint,
                     address: event.account,
                     amount: event.total_deposit,
-                    is_deposit: true,
+                    is_addition: true,
                 };
 
-                deposits.push(info);
+                balance_updates.push(info);
             }
 
             if let Ok(event) = contract::parse_log::<WithdrawnFilter>(log.clone()) {
@@ -469,14 +470,14 @@ impl<P: Provider> Chain<P> {
                     entrypoint,
                     address: event.account,
                     amount: event.amount,
-                    is_deposit: false,
+                    is_addition: false,
                 };
 
-                deposits.push(info);
+                balance_updates.push(info);
             }
         }
 
-        deposits
+        balance_updates
     }
 
     fn block_with_number(&self, number: u64) -> Option<&BlockSummary> {
@@ -1093,13 +1094,13 @@ mod tests {
     fn fake_mined_balance_update(
         address: Address,
         amount: U256,
-        is_deposit: bool,
+        is_addition: bool,
     ) -> BalanceUpdate {
         BalanceUpdate {
             address,
             entrypoint: ENTRY_POINT_ADDRESS,
             amount,
-            is_deposit,
+            is_addition,
         }
     }
 
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -27,7 +27,10 @@ use futures::future;
 use rundler_provider::Provider;
 use rundler_task::block_watcher;
 use rundler_types::{
-    contracts::{entry_point::DepositedFilter, i_entry_point::UserOperationEventFilter},
+    contracts::{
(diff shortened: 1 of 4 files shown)


crates/pool/src/mempool/uo_pool.rs
```diff
@@ -170,6 +170,20 @@ where
...
+                .cloned()
+                .collect();
+
+            let unmined_entity_withdrawals: Vec<DepositInfo> = update
+                .unmined_entity_withdrawals
+                .iter()
+                .filter(|d| d.entrypoint == self.config.entry_point)
+                .cloned()
+                .collect();
+
```
dancoombs, line 186: This sort of code duplication is starting to be littered all over the mempool code. Where we have some sort of iterator on a specific event type, and have to repeat that for each event type. Are there ways to clean this up?

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -170,6 +170,20 @@ where
                 .cloned()
                 .collect();
 
+            let withdrawals: Vec<DepositInfo> = update
+                .entity_withdrawals
+                .iter()
+                .filter(|d| d.entrypoint == self.config.entry_point)
+                .cloned()
+                .collect();
```
dancoombs, line 178: Can we change the paymaster tracker interface so that it can take an iterator or intoIter? Converting these all into vectors is inefficient.

crates/pool/src/mempool/pool.rs
```diff
@@ -359,6 +359,8 @@ impl PoolInner {
         &mut self,
         deposits: &Vec<DepositInfo>,
         unmined_entity_deposits: &Vec<DepositInfo>,
+        withdrawals: &Vec<DepositInfo>,
+        unmined_entity_withdrawals: &Vec<DepositInfo>,
```
dancoombs, line 363: Surprised this isn't a lint rule, wouldn't `&[DepositInfo]` be cleaner here?

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -170,6 +170,20 @@ where
                 .cloned()
                 .collect();
 
+            let withdrawals: Vec<DepositInfo> = update
+                .entity_withdrawals
+                .iter()
+                .filter(|d| d.entrypoint == self.config.entry_point)
```
dancoombs, line 176: Maybe its time to lift this entry point filtering to a layer above `on_chain_update` given that its now copied in 6 places.

dancoombs (commented):
In general there is a lot of duplicated code here that could likely use some refactoring. Probably worth spending a little time cleaning this up if we can think of a better approach.

0xfourzerofour:
@dancoombs I can have a look and move some stuff into another layer as I agree the chain update function is too chaotic if we keep adding new listeners

crates/pool/src/chain.rs
```diff
@@ -71,6 +74,10 @@ pub struct ChainUpdate {
     pub entity_deposits: Vec<DepositInfo>,
     /// List of entity deposits that have been unmined due to a reorg
     pub unmined_entity_deposits: Vec<DepositInfo>,
+    /// List of on-chain entity withdrawals made in the most recent block
+    pub entity_withdrawals: Vec<DepositInfo>,
+    /// List of entity withdrawals that have been unmined due to a reorg
+    pub unmined_entity_withdrawals: Vec<DepositInfo>,
```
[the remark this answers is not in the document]
dancoombs: I think `BalanceUpdate` is fine. Could we possibly combine deposits/withdrawals into a single collection using negative values fro withdrawals?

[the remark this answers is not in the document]
0xfourzerofour: The actual event is a positive value but we could have a field to delineate whether we want to subtract or add. I would just rather store negative numbers as we are doing saturated subs and adds on a unsigned into so would be cleaner to just have like a boolean to do either then we can combine like 4 of the paymaster functions into one

crates/pool/src/mempool/uo_pool.rs
0xfourzerofour: I will make a ticket for this but I agree, having to reference the entrypoint on every chain update is not ideal

crates/pool/src/mempool/uo_pool.rs
0xfourzerofour: I updated the function to take an iterator instead

crates/pool/src/chain.rs
```diff
@@ -443,15 +449,27 @@ impl<P: Provider> Chain<P> {
...
+
+            if let Ok(event) = contract::parse_log::<WithdrawnFilter>(log.clone()) {
+                let info = BalanceUpdate {
+                    entrypoint,
+                    address: event.account,
+                    amount: event.amount,
+                    is_deposit: false,
                 };
 
                 deposits.push(info);
```
alex-miao, line 475: nit: can we rename this vec `balance_updates`

crates/pool/src/mempool/paymaster.rs
```diff
@@ -90,25 +90,18 @@ impl PaymasterTracker {
         keys
     }
 
-    pub(crate) fn update_paymaster_balance_after_deposit_reorg(
+    pub(crate) fn update_paymaster_balance(
         &mut self,
         paymaster: Address,
-        deposit_amount: U256,
+        amount: U256,
+        is_deposit: bool,
```
alex-miao, line 97: Is this bool used for both deposit/withdraw and mine/unmine? Not sure if its the best name but could we rename it to something like `should_add` since you can since an unmine deposit we should be subtracting.

dancoombs: Agreed the framing of `is_deposit` is confusing here given that its used for deposit/withdraw mine/unmine.

crates/pool/src/mempool/paymaster.rs
```diff
@@ -90,25 +90,18 @@ impl PaymasterTracker {
...
-    pub(crate) fn update_paymaster_balance_from_deposit(
-        &mut self,
-        paymaster: Address,
-        deposit_amount: U256,
-    ) {
-        if let Some(paymaster_balance) = self.paymaster_balances.get_mut(&paymaster) {
-            paymaster_balance.confirmed =
-                paymaster_balance.confirmed.saturating_add(deposit_amount);
+            if is_deposit {
+                paymaster_balance.confirmed = paymaster_balance.confirmed.saturating_sub(amount);
```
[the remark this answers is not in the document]
dancoombs: How does `is_deposit` lead to a `saturating_sub `? Shouldn't a deposit lead to adding funds?

crates/pool/src/mempool/paymaster.rs
```diff
@@ -90,25 +90,18 @@ impl PaymasterTracker {
         keys
     }
 
-    pub(crate) fn update_paymaster_balance_after_deposit_reorg(
+    pub(crate) fn update_paymaster_balance(
```
dancoombs, line 93: To disambiguate this from `update_paymaster_balance_from_mined_op` maybe we should call it `update_paymaster_balance_from_event` or something?

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: Yeh I have seperated the names because technically the it is a deposit but the code in the paymaster logic has been changed to `should_add` because if the update is an unmined deposit we check for is_deposit and flip the boolean

crates/pool/src/mempool/paymaster.rs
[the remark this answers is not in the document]
0xfourzerofour: I fixed the naming add added the fields to the mine and unmine tests in the uo_pool to catch it

crates/pool/src/mempool/paymaster.rs
```diff
@@ -97,25 +97,18 @@ impl PaymasterTracker {
         keys
     }
 
-    pub(crate) fn update_paymaster_balance_after_deposit_reorg(
+    pub(crate) fn update_paymaster_balance_from_event(
         &mut self,
         paymaster: Address,
-        deposit_amount: U256,
+        amount: U256,
+        should_add: bool,
```
dancoombs, line 104: nit: This looks exactly like a `DepositUpdate` can we just pass that in?

dancoombs (approved):
One small nit but LGTM

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: its slightly different because the the `BalanceUpdate` stores the transaction that was put on chain but if we sense that it is a reorg and that transaction needs to be added back we need to take the opposite of what that stored transaction was. So if its an unmine the logic is !is_addition for a deposit so we subtract from the paymaster's deposits

crates/pool/src/mempool/paymaster.rs
```diff
@@ -389,6 +382,43 @@ mod tests {
         assert!(res.is_err());
     }
 
+    #[test]
+    fn update_balance_after_deposit() {
```
alex-miao, line 386: nit: also testing withdraw here so maybe rename to something like `test_update_balance`

crates/pool/src/mempool/paymaster.rs
dancoombs:
```
pub struct BalanceUpdate {
    pub address: Address,
    pub entrypoint: Address,
    pub amount: U256,
    pub is_addition: bool,
}
```

You're passing in 3 of the 4 fields? I don't see it storing the transaction

dancoombs: But I get it, we need to flip `is_addition` if its unmined, but that could just be a function on `BalanceUpdate`

0xfourzerofour: Rip I merged as I though that would suffice. I can update my next PR to have that logic if it preferred I just like the separation of chain logic and paymaster logic

PR #576, opened 2024-01-31
fix(reputation): do not increase ops seen on replacement

### LINKED ISSUE #498: [pool] Reputation for replacement operations

alex-miao:
Closes #498

## Proposed Changes

  - Only update reputation if op is not a replacement

### DIFF
--- a/crates/pool/src/mempool/uo_pool.rs
+++ b/crates/pool/src/mempool/uo_pool.rs
@@ -364,7 +364,7 @@ where
 
         // Check if op is already known or replacing another, and if so, ensure its fees are high enough
         // do this before simulation to save resources
-        self.state.read().pool.check_replacement(&op)?;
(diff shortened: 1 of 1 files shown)


crates/pool/src/mempool/uo_pool.rs
```diff
@@ -457,11 +457,13 @@ where
 
         // Update reputation
         pool_op.entities().unique().for_each(|e| {
```
0xfourzerofour, line 459: Maybe put the replacement check outside the `for_each` so we can skip going through it all together on replacements

0xfourzerofour (commented):
I think this    should be sd

0xfourzerofour (approved):
LGTM

Issue #564, opened 2024-01-22
consider remove ReputationManager trait and duplicate code  


0xfourzerofour:
There is no real need for the trait to exist as the Reputation tracker is just updating in memory data and can be easily tested with new data while testing instead of creating a mock implementation that is doing the same same thing.

0xfourzerofour:
While this is done, think about moving the reputation state into the pool state. And also refactor the clear_mempool functionality to reflect the new state

PR #577, opened 2024-02-01
feat(paymaster): refactor paymaster operation submission and add support for removing stale paymasters

### LINKED ISSUE #520: Paymaster Tracking: check if paymaster has become stale and can be removed from the pool
Paymaster Tracking: check if paymaster has become stale and can be removed from the pool

0xfourzerofour:
Closes #520 #564 

## Proposed Changes

  - Refactor the paymaster state to be outside of the pool and hold its own lock
  - Add support for a background process to remove stale paymasters each hour
  - Follow the same pattern in the reputation logic
  - Remove need for Reputation trait as it is just updating in memory data
  - Move the network calls for paymaster into its own manager functions while still not making network calls under a lock.


crates/pool/src/mempool/reputation.rs
```diff
@@ -250,10 +120,93 @@ impl ReputationParams {
...
+            state: RwLock::new(
+                AddressReputationInner::new(params)
+                    .with_blocklist(blocklist)
+                    .with_allowlist(allowlist),
+            ),
+        }
+    }
+
+    pub(crate) async fn run(&self) {
+        let mut tick = interval(Duration::from_secs(self.state.read().params.task_interval));
```
dancoombs, line 154: Since we made this tick configurable, we should also rename the the update function from `hourly_update()` to just `update()` and then should also make the `/ 24` that happens in that function configurable.

dancoombs: I think we should call these `decay_interval` and `decay_factor`

crates/pool/src/mempool/reputation.rs
```diff
@@ -250,10 +120,93 @@ impl ReputationParams {
...
     }
+
+    #[cfg(test)]
+    pub(crate) fn test_parameters(ban_slack: u64, throttling_slack: u64) -> Self {
+        Self {
+            ban_slack,
+            throttling_slack,
+            ..Self::default()
+        }
+    }
```
dancoombs, line 131: nit: this seems like a pretty simple constructor, is it worth defining this here and polluting the class or can we just define it in the tests themselves?

crates/pool/src/task.rs
```diff
@@ -100,7 +100,9 @@ impl Task for PoolTask {
             .await
             .context("should have created mempool")?;
 
-            mempools.insert(pool_config.entry_point, Arc::new(pool));
+            let pool_ref = Arc::new(pool);
+
+            mempools.insert(pool_config.entry_point, Arc::clone(&pool_ref));
```
dancoombs, line 105: Doesn't this result in the same as before?

crates/pool/src/mempool/paymaster.rs
```diff
@@ -17,14 +17,234 @@ use std::collections::HashMap;
...
+
+        // Save paymaster balance after first lookup
+        self.state
+            .write()
+            .add_new_paymaster(paymaster, balance, 0.into());
+
+        Ok(paymaster_meta)
+    }
+
+    // TODO check if its a replacement
```
dancoombs, line 149: What is the plan with this TODO?

crates/pool/src/mempool/paymaster.rs
```diff
@@ -17,14 +17,234 @@ use std::collections::HashMap;
...
+    pub(crate) fn new(paymaster_helper: PH, entry_point: E, config: PaymasterConfig) -> Self {
+        Self {
+            paymaster_helper,
+            entry_point,
+            state: RwLock::new(PaymasterTrackerInner::new(config.tracker_enabled)),
+            config,
+        }
+    }
+
+    pub(crate) async fn get_stake_status(&self, address: Address) -> MempoolResult<StakeStatus> {
```
dancoombs, line 74: Realizing this from the last PR, why is this part of the paymaster tracker? Do we not ask for the stake of other entities? Might make sense to refactor this out.

crates/pool/src/mempool/paymaster.rs
```diff
@@ -17,14 +17,234 @@ use std::collections::HashMap;
...
+            min_stake_value,
+            min_unstake_delay,
+            tracker_enabled,
+        }
+    }
+}
+
+impl<PH, E> PaymasterTracker<PH, E>
+where
+    PH: PaymasterHelper,
```
dancoombs, line 62: Also realizing this now, the `PaymasterHelper` isn't really paymaster specific, it can get balances/stakes of any entity.

Also we should change `getDepositInfo()` on it to a batch interface `getDepositInfos()`, it isn't useful as just a passthrough to the stake manager.

We should also change our builder -> pool interface to make batched calls for stake statuses, to limit the amount of calls that we're making. We can should then split out stake tracking into its own class.

Creating a ticket to handle this in 0.1 here:

dancoombs: Also, I think I'm going to remove the `PaymasterHelper` class and just stick these functions into our `EntryPoint` trait.

crates/pool/src/mempool/paymaster.rs
```diff
@@ -17,14 +17,234 @@ use std::collections::HashMap;
...
+        id: &UserOperationId,
+        paymaster_metadata: &PaymasterMetadata,
+        max_op_cost: U256,
+    ) {
+        self.state
+            .write()
+            .add_new_user_op(id, paymaster_metadata, max_op_cost)
+    }
+
+    #[cfg(test)]
```
dancoombs, line 232: nit: In general I'm not a fan of adding test logic to classes unless strictly necessary. Can we just add these functions to the test module?

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -303,24 +310,16 @@ where
...
-        self.state
-            .write()
-            .pool
-            .set_confirmed_paymaster_balances(&paymaster_addresses, &balances);
+        self.paymaster.reset_confimed_balances().await
+    }
 
-        Ok(())
+    async fn get_stake_status(&self, address: Address) -> MempoolResult<StakeStatus> {
+        self.paymaster.get_stake_status(address).await
```
dancoombs, line 322: Going to move this it a stake tracker class as we need take information for entities other than paymasters

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -449,15 +438,17 @@ where
...
+            let hash = state.pool.add_operation(pool_op.clone())?;
+
             if throttled {
                 state.throttled_ops.insert(hash);
             }
             hash
         };
 
+        // Add op cost to pending paymaster balance
+        self.paymaster.add_or_update_balance(&pool_op).await?;
```
dancoombs, line 450: Just pointing out that this is a different semantic than before. Previously it was impossible for a paymaster to overdraw given that everything was happening under the same pool lock. Now 2 ops can come in at the same time, both pass the initial `check_operation_cost`, both get added to the pool, and then update the pending balance to an overdraw.

Given the complexity of handling the async calls in the paymaster tracker, I'm okay with this this tradeoff because 1) the chances of this race condition are low, 2) we'll catch it during bundling.

It might be worth adding a comment to the check above that this is an optimistic (not a strict) check.

dancoombs (commented):
Nice work! In general much cleaner to reason about.

It does make the paymaster balance checking optimistic and not strict, but I think that is a reasonable tradeoff to make.

Added a ticket to follow up with some stake tracking / paymaster helper refactors.

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: The entrypoint interface does not have the Deposit info, are we going to combine the Entrypoint trait to deal with the StakeManager interface?

crates/pool/src/mempool/paymaster.rs
0xfourzerofour: Yeh I agree we could make this more global

crates/pool/src/task.rs
0xfourzerofour: yeh it does I can remove this code, this was some remaining code after I removed the hourly paymaster updates

crates/pool/src/mempool/paymaster.rs
dancoombs: The `IEntrypoint` interface doesn't, but our `Entrypoint` interface can define anything. We can just add the paymaster helper methods there.

0xfourzerofour: Yeh I agree, its also annoying bringing a bunch of traits in the uo_pool modules it just pretty busy

crates/pool/src/mempool/reputation.rs
0xfourzerofour: This is used in the uo_pool not in the reputation tests so we cannot define it in the uo pool due to private field initialization. I am also not aware that you can define a function a test section within a file and import that to another file? I thought that was not an option with the compiler but I could be wrong

crates/pool/src/mempool/reputation.rs
0xfourzerofour: So you want to have instead of just task interval, task interval / factor. I am trying to understand the reason as the reason for this. We want to say how many times in a day this is executed?

0xfourzerofour: Is the factor not deduced by the `interval` function?

dancoombs: I'm trying to come up with a name for the hardcoded `/ 24` that happens. That seems very coupled to the hourly interval that was hardcoded previously, but is now configurable.

crates/pool/src/mempool/reputation.rs
dancoombs: Can skip this if its not easy to do

crates/pool/src/mempool/reputation.rs
```diff
@@ -213,6 +81,8 @@ pub(crate) struct ReputationParams {
     throttling_slack: u64,
     ban_slack: u64,
     tracking_enabled: bool,
+    decay_interval: u64,
```
dancoombs, line 84: nit: I like to be very explicit with units on time intervals. Lets call this `decay_interval_seconds`

crates/pool/src/mempool/reputation.rs
```diff
@@ -471,14 +437,14 @@ mod tests {
...
         for _ in 0..1000 {
             reputation.add_seen(addr);
             reputation.add_included(addr);
         }
 
-        reputation.hourly_update();
+        reputation.update();
         let counts = reputation.counts.get(&addr).unwrap();
         assert_eq!(counts.ops_seen, 1000 - 1000 / 24);
         assert_eq!(counts.ops_included, 1000 - 1000 / 24);
```
dancoombs, line 450: nit: probably should update the tests to remove the hardcoded `24` and use the settings

dancoombs (approved):
LGTM! A few nits

dancoombs (approved):
🚀

PR #580, opened 2024-02-01
feat(pool): reject op before simulation if pool is full

### LINKED ISSUE #501: [pool] Skip simulation if the pool is full and the UO will not be accepted
**Describe the feature**
The pool should skip simulation if it is currently full and the UO will not be accepted (based on fees) after simulation completes. We should reject before simulating to save resources and protect against a DOS vector.

alex-miao:
Closes #501

## Proposed Changes

  - Check to see if there is enough space in the mempool to hold a new UO if we were to evict UOs with lower `max_fee_per_gas`

### DIFF
--- a/crates/pool/src/mempool/mod.rs
+++ b/crates/pool/src/mempool/mod.rs
@@ -273,6 +273,10 @@ impl PoolOperation {
             + self.uo.heap_size()
             + self.entities_needing_stake.len() * std::mem::size_of::<EntityType>()
     }
+
+    pub(crate) fn max_size(uo: &UserOperation) -> usize {
+        std::mem::size_of::<Self>() + uo.heap_size() + 4 * std::mem::size_of::<EntityType>()
+    }
 }
 
 #[cfg(test)]
--- a/crates/pool/src/mempool/pool.rs
+++ b/crates/pool/src/mempool/pool.rs
@@ -165,6 +165,25 @@ impl PoolInner {
         self.best.clone().into_iter().map(|v| v.po)
     }
 
+    // Check if it is worth adding an user operation to the pool given pool size constraints and the price of the UO.
+    pub(crate) fn check_pool_capacity(&self, uo: &UserOperation) -> MempoolResult<()> {
+        let max_op_size = OrderedPoolOperation::max_size(uo);
+        let mut free_space =
+            self.config.max_size_of_pool_bytes - (std::cmp::max(self.pool_size.0, 0) as usize);
+        let mut req_fee_per_gas = U256::zero();
+        let mut pool_iter = self.best.iter().rev();
+
+        while !(max_op_size <= free_space && uo.max_fee_per_gas > req_fee_per_gas) {
+            if let Some(op) = pool_iter.next() {
+                req_fee_per_gas = op.po.uo.max_fee_per_gas;
+                free_space += op.mem_size();
+            } else {
+                return Err(MempoolError::DiscardedOnInsert);
+            }
+        }
+        Ok(())
+    }
+
     /// Removes all operations using the given entity, returning the hashes of the removed operations.
     ///
     /// NOTE: This method is O(n) where n is the number of operations in the pool.
@@ -557,7 +576,11 @@ impl OrderedPoolOperation {
     }
 
     fn mem_size(&self) -> usize {
-        std::mem::size_of::<OrderedPoolOperation>() + self.po.mem_size()
+        std::mem::size_of::<Self>() + self.po.mem_size()
+    }
+
+    fn max_size(uo: &UserOperation) -> usize {
+        std::mem::size_of::<Self>() + PoolOperation::max_size(uo)
     }
 }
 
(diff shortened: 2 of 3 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -165,6 +165,25 @@ impl PoolInner {
...
+    // Check if it is worth adding an user operation to the pool given pool size constraints and the price of the UO.
+    pub(crate) fn check_pool_capacity(&self, uo: &UserOperation) -> MempoolResult<()> {
+        let max_op_size = OrderedPoolOperation::max_size(uo);
+        let mut free_space =
+            self.config.max_size_of_pool_bytes - (std::cmp::max(self.pool_size.0, 0) as usize);
+        let mut req_fee_per_gas = U256::zero();
+        let mut pool_iter = self.best.iter().rev();
+
+        while !(max_op_size <= free_space && uo.max_fee_per_gas > req_fee_per_gas) {
+            if let Some(op) = pool_iter.next() {
```
alex-miao, line 177: Doing a naive linear scan of the ops with the lowest `max_fee_per_gas values` but I think you could theoretically make this check `O(log n)` by using some efficient [dynamic prefix sum](https://stackoverflow.com/questions/27990143/dynamic-prefix-sum) data structure to efficiently give you the sum of the sizes of all the ops currently in the pool that have a smaller `max_fee_per_gas` value.

0xfourzerofour: I wonder if we could just have an average base + priority fee field over n blocks and then just check if the max_fee_per_gas is more than that instead of looping through the mempool ops?

crates/pool/src/mempool/mod.rs
```diff
@@ -273,6 +273,10 @@ impl PoolOperation {
             + self.uo.heap_size()
             + self.entities_needing_stake.len() * std::mem::size_of::<EntityType>()
     }
+
+    pub(crate) fn max_size(uo: &UserOperation) -> usize {
+        std::mem::size_of::<Self>() + uo.heap_size() + 4 * std::mem::size_of::<EntityType>()
```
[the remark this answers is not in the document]
0xfourzerofour: oh I guess you dont have access to the OrderedPoolOperation struct when you call this so I think this is ok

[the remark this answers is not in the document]
alex-miao: This is meant to be for a UO that has not been turned into a PoolOperation yet since all of this is happening before simulation. I thought it might make more sense to have this be a `PoolOperation` associated function rather than creating a spoof `PoolOperation` from a UO.

crates/pool/src/mempool/pool.rs
alex-miao: Oh interesting how does the paymaster stuff map to this data structure? Might be worth implementing if its not just a one off

0xfourzerofour: It does not right now, however we want to find a good way to remove stale paymasters and I am thinking of using a `last_used` timestamp. And then remove all paymasters if they have not been used in the last `n` hours /minutes etc. but having a datastructure that can somewhat index on fields would make that a lot better as I do not want to loop through all the paymasters too often so 0(log n) would be great.

alex-miao: I'm guessing we would want both types of checks, one to reject based on block history and this one to reject cheaper ops even if all the ops in the pool are already more expensive than recent history.

crates/pool/src/mempool/pool.rs
```diff
@@ -165,6 +165,25 @@ impl PoolInner {
...
 
+    // Check if it is worth adding an user operation to the pool given pool size constraints and the price of the UO.
+    pub(crate) fn check_pool_capacity(&self, uo: &UserOperation) -> MempoolResult<()> {
+        let max_op_size = OrderedPoolOperation::max_size(uo);
+        let mut free_space =
+            self.config.max_size_of_pool_bytes - (std::cmp::max(self.pool_size.0, 0) as usize);
+        let mut req_fee_per_gas = U256::zero();
+        let mut pool_iter = self.best.iter().rev();
+
+        while !(max_op_size <= free_space && uo.max_fee_per_gas > req_fee_per_gas) {
```
dancoombs, line 176: This is hard to grok, trying to work through this in my head.

Applying demorgans:

```
while (max_op_size > free_space || uo.max_fee_per_gas < req_fee_per_gas)
```

That `||` doesn't make sense to me, you'd always want to exit if `max_op_size <= free_space` correct?

Shouldn't this be:

```
while (max_op_size > free_space && uo.max_fee_per_gas > req_fee_per_gas)
```

i.e. iterate while there isn't enough space yet still, but the uo is still paying enough to replace the lowest priced UOs. Exit conditions are: 

1) There is enough free space -> ok 
2) UO doesn't pay enough to replace current lowest, exit the loop, check free space, if still not enough then fail.
3) Pool iterator runs out: this would mean that this UO is the highest priced in the pool but still there isn't enough space, we can check for this condition above because it means that `max_op_size > max_pool_bytes`, so this shouldn't happen.

dancoombs, line 176: Thinking through our discarding logic I think it needs to be rewritten. Would love thoughts on a fix.

Currently we optimize for getting the highest priced operation into the pool. This lets a UO that has a lot of calldata and a slightly higher price cause a discard lots of smaller, slightly lower priced, UOs.

Consider this simple attack:

1) Construct a UO with a TON of calldata such that it will cause the mempool to be full. Price it highest.
2) Send this UO, this will cause the pool to discard the entire existing pool just to hold this single UO.
3) Attacker invalidates this UO (simple to do for a single UO), and the bundler is left with nothingl.

Can you look at how Geth/Reth handle this logic in their mempools?

alex-miao: The way that I think about it is that we only want to return `Ok(())` after the while loop, so that means that the invariant of the loop must be false in order to return `Ok(())`. If we use the condition `max_op_size > free_space && uo.max_fee_per_gas > req_fee_per_gas`, that means that we return `Ok(())` if `max_op_size <= free_space || uo.max_fee_per_gas <= req_fee_per_gas` but we need both of those conditions to be true, not either. This would give us false positives like the last `assert` of `test_check_pool_capacity` where we already have a full pool with two ops of `max_fee_per_gas=10` and it says we would be able to add another op of `max_fee_per_gas=9`

PR #582, opened 2024-02-02
feat(builder): add option to pass in builders to flashbots mev-share rpc

### LINKED ISSUE #283: builder: Send bundle to more builders on L1
**Describe the feature**
https://docs.flashbots.net/flashbots-protect/rpc/mev-share#builders

Add a CLI option to modify the builders that receive bundles. Update transaction status tracking to handle multiple builders.

alex-miao:
Closes #283

## Proposed Changes

  - pass in additional builders through the MEV-Share RPC query params

### DIFF
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -152,14 +152,15 @@ pub struct BuilderArgs {
     )]
     max_fee_increases: u64,
 
-    /// A list of builders to pass into the Flashbots MEV-Share RPC
+    /// A list of builders to pass into the Flashbots Relay RPC. Only used when BUILDER_SENDER is "flashbots" and BUILDER_SUBMIT_URL is a Flashbots RPC
     #[arg(
-        long = "builder.mev_share_builders",
-        name = "builder.mev_share_builders",
-        env = "BUILDER_MEV_SHARE_BUILDERS",
-        value_delimiter = ','
+        long = "builder.flashbots_relay_builders",
+        name = "builder.flashbots_relay_builders",
+        env = "BUILDER_FLASHBOTS_RELAY_BUILDERS",
+        value_delimiter = ',',
+        default_value = "flashbots"
     )]
-    mev_share_builders: Vec<String>,
+    flashbots_relay_builders: Vec<String>,
 
     /// If using Polygon Mainnet, the auth header to use
     /// for Bloxroute polygon_private_tx sender
@@ -270,7 +271,7 @@ impl BuilderArgs {
             replacement_fee_percent_increase: self.replacement_fee_percent_increase,
             max_fee_increases: self.max_fee_increases,
             remote_address,
-            mev_share_builders: self.mev_share_builders.clone(),
+            flashbots_relay_builders: self.flashbots_relay_builders.clone(),
             bloxroute_auth_header: self.bloxroute_auth_header.clone(),
         })
     }
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -152,6 +152,15 @@ pub struct BuilderArgs {
     )]
     max_fee_increases: u64,
 
+    /// A list of builders to pass into the Flashbots MEV-Share RPC
+    #[arg(
+        long = "builder.mev_share_builders",
+        name = "builder.mev_share_builders",
+        env = "BUILDER_MEV_SHARE_BUILDERS",
+        value_delimiter = ','
+    )]
+    mev_share_builders: Vec<String>,
+
     /// If using Polygon Mainnet, the auth header to use
     /// for Bloxroute polygon_private_tx sender
     #[arg(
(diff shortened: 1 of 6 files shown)


bin/rundler/src/cli/builder.rs
```diff
@@ -152,6 +152,15 @@ pub struct BuilderArgs {
     )]
     max_fee_increases: u64,
 
+    /// A list of builders to pass into the Flashbots MEV-Share RPC
+    #[arg(
+        long = "builder.mev_share_builders",
+        name = "builder.mev_share_builders",
+        env = "BUILDER_MEV_SHARE_BUILDERS",
+        value_delimiter = ','
```
dancoombs, line 160: nit: lets be overly verbose here and call this `BUILDER_FLASHBOTS_MEV_SHARE_BUILDERS`

dancoombs: Can you add this to `cli.md` also

crates/builder/src/sender/flashbots.rs
```diff
@@ -176,8 +181,14 @@ struct FlashbotsClient {
...
 impl FlashbotsClient {
-    fn new() -> anyhow::Result<Self> {
-        let client = HttpClientBuilder::default().build("https://rpc.flashbots.net")?;
+    fn new(builders: &Vec<String>) -> anyhow::Result<Self> {
+        let mut flashbots_url = Url::parse("https://rpc.flashbots.net")?;
+        for builder in builders {
+            flashbots_url
+                .query_pairs_mut()
+                .append_pair("builder", builder);
+        }
```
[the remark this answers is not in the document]
dancoombs: Found it nvm: https://docs.flashbots.net/flashbots-protect/mev-share#configuration-reference

crates/builder/src/sender/flashbots.rs
```diff
@@ -176,8 +181,14 @@ struct FlashbotsClient {
 }
 
 impl FlashbotsClient {
-    fn new() -> anyhow::Result<Self> {
-        let client = HttpClientBuilder::default().build("https://rpc.flashbots.net")?;
+    fn new(builders: &Vec<String>) -> anyhow::Result<Self> {
+        let mut flashbots_url = Url::parse("https://rpc.flashbots.net")?;
```
dancoombs, line 185: Thoughts on switching to using the relay RPC instead? I think its better built for our purposes: https://docs.flashbots.net/guide-send-tx-bundle#relayflashbotsnet---for-advanced-users

dancoombs: We'd then need to switch to using `eth_sendPrivateRawTransaction` since we don't make use of the `maxBlockNumber` field. Although when we rework the builder it would be interesting to start using that field so that we submit a bundle every other block or something like that.

alex-miao: Switched it over, noticed there is the `preferences.fast` command that defaults to sending to all builders but also sets the revenue share to 50% as opposed to the default 10%.

bin/rundler/src/cli/builder.rs
```diff
@@ -152,6 +152,16 @@ pub struct BuilderArgs {
     )]
     max_fee_increases: u64,
 
+    /// A list of builders to pass into the Flashbots Relay RPC
+    #[arg(
+        long = "builder.flashbots_relay_builders",
+        name = "builder.flashbots_relay_builders",
+        env = "BUILDER_FLASHBOTS_RELAY_BUILDERS",
+        value_delimiter = ',',
+        default_value = ""
```
dancoombs, line 161: Does this default still send to flashbots?

docs/cli.md
```diff
@@ -170,6 +170,8 @@ List of command line options for configuring the Builder.
   - env: *BUILDER_REPLACEMENT_FEE_PERCENT_INCREASE*
 - `--builder.max_fee_increases`: Maximum number of fee increases to attempt (Seven increases of 10% is roughly 2x the initial fees) (default: `7`)
   - env: *BUILDER_MAX_FEE_INCREASES*
+- `--builder.flashbots_replay_builders`: additional builders to send bundles to through the Flashbots relay RPC (comma-separated)
```
dancoombs, line 173: Spelling: `flashbots_replay_builders` -> `flashbots_relay_builders`

dancoombs, line 173: Can you add a blurb about what this defaults to?

dancoombs: We should also link to https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#eth_sendprivatetransaction so that users know what builders they can choose from

bin/rundler/src/cli/builder.rs
dancoombs: This should probably default to `"flashbots"`

bin/rundler/src/cli/builder.rs
```diff
@@ -152,6 +152,16 @@ pub struct BuilderArgs {
     )]
     max_fee_increases: u64,
 
+    /// A list of builders to pass into the Flashbots Relay RPC
```
dancoombs, line 155: We should make it clear this is only used if the `builder.submit_url` above is a flashbots URL.

PR #583, opened 2024-02-05
feat(chains): add polygon amoy and use alloy-rs

0xfourzerofour:
Add polygon amoy and use alloy-rs for chain definitions

### DIFF
--- a/crates/types/Cargo.toml
+++ b/crates/types/Cargo.toml
@@ -12,6 +12,7 @@ rundler-utils = { path = "../utils" }
 anyhow.workspace = true
 chrono = "0.4.24"
 ethers.workspace = true
+alloy-chains.workspace = true
 parse-display = "0.8.0"
 serde.workspace = true
 serde_json.workspace = true
--- a/crates/types/Cargo.toml
+++ b/crates/types/Cargo.toml
@@ -8,11 +8,12 @@ repository.workspace = true
 
 [dependencies]
 rundler-utils = { path = "../utils" }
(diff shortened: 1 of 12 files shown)


alex-miao (approved):

crates/types/Cargo.toml
```diff
@@ -8,10 +8,11 @@ repository.workspace = true
 
 [dependencies]
 rundler-utils = { path = "../utils" }
-
+constcat = "0.4.1"
 anyhow.workspace = true
 chrono = "0.4.24"
 ethers.workspace = true
+alloy-chains.workspace = true
```
dancoombs, line 15: nit: here and in the other .toml, we've at least attempted to keep the dependencies in alphabetical order. We should probably start enforcing this with https://crates.io/crates/cargo-sort.

For now do you mind fixing at least your additions, and adding a ticket to enable `cargo-sort`?

dancoombs (commented):
LGTM besides the sort nit. Can't wait to remove all these chain references with the `ChainSpec`.

Lets get that sort nit worked out, this merged, and then cherry-pick this to `0.1.0-beta.0`

crates/types/Cargo.toml
0xfourzerofour: Yeh I sorted that, I tried cargo-sort but it does not yet support workspace inheritance and returns an error due to the `version.workspace = true` or `{dependency}.workspace = true`

dancoombs (approved):

PR #599, opened 2024-02-15
feat(sim): reject UOs that don't have a buffer on verification gas limit

### LINKED ISSUE #593: Run `test_enough_verification_gas` and ensure pass
**Describe the feature**
From 4337 team:

> A crafted UserOp can waste a little more than the total prefund of the userop, and thus can pass validation, and still cause a bundle to revert.
> 
> The root cause is that a UserOp may use little more than the total gas advertised in the UserOp's gaslimits
> It doesn't cause a problem if there is any slack (using little less than the gas-limit)
> 
> The solution for bundler: to require that verificationGasLimit will have some unused slack. This has a side-effect on the account to require a slightly higher "deposit" - but since it is not used, it is fully refunded at the end.
> 
> We pushed a fix to the reference bundler (which should be now moved to the "releases/v0.6" branch)
> There is a new test in the "bundler-spec-test" repo, on branch "releases/v0.6", named  "test_enough_verification_gas"
> 
> To test it, run: pdm run test -k test_enough_verification_gas --log-rpc -vvv

dancoombs:
Closes #593 

## Proposed Changes

- Adds a simulation violation check that requires a UO to have at least a 2000 gas buffer on their `verificationGasLimit` compared to what is used during simulation. This accounts for a situation where a UO can use slightly more than their VGL onchain, causing a bundle to revert.

### DIFF
--- a/crates/pool/proto/op_pool/op_pool.proto
+++ b/crates/pool/proto/op_pool/op_pool.proto
@@ -558,6 +558,7 @@ message SimulationViolationError {
     AggregatorValidationFailed aggregator_validation_failed = 16;
     UnstakedPaymasterContext unstaked_paymaster_context = 17;
     UnstakedAggregator unstaked_aggregator = 18;
+    VerificationGasLimitBufferTooLow verification_gas_limit_buffer_too_low = 19;
   }
 }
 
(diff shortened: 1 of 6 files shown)


0xfourzerofour (approved):
Looks good!

PR #607, opened 2024-02-17
feat(builder): add a timer based trigger to bundle building

### LINKED ISSUE #598: Enable triggering of bundles on a timer
**Describe the feature**
Currently our bundle submission only happens on block boundaries. This is fine for networks that have a consistent block time, but for networks that don't (i.e. they only emit a block when a new transaction is received), there needs to be another way to trigger bundle submission.

To enable this we can add a setting to `ChainSpec` that enables a timer based submission for a chain.

Once that is complete, we can remove any changes to our bundle-spec-tests fork that required manual bundle mode.

dancoombs:
Closes #598

## Proposed Changes

  - Adds a timer based trigger to the bundle sender. The bundle sender will send a bundle if the timer elapses before it receives the next block. Currently this is set to maximum time for all networks except for dev where it is set to 250ms. This allows bundles to be sent on dev Geth where blocks are only triggered when transactions arrive.
  - Its possible there will be other networks that have unpredictable block times where this chain spec rule could be useful.
  - Fixes switching between bundling modes. Previously if we transition to manual mode we'd never be able to exit unless the tests switch the mode while bundling building is in flight. Once we looped back around we'd be stuck in manual mode until a new bundle was triggered. Doing this required a refactor of how the server interacts with the bundle senders.

### DIFF
--- a/bin/rundler/chain_specs/dev.toml
+++ b/bin/rundler/chain_specs/dev.toml
@@ -2,3 +2,5 @@ base = "ethereum"
 
 name = "Ethereum Devnet"
 id = 1337
+
+bundle_max_send_interval_millis = 250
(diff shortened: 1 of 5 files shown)


0xfourzerofour (approved):
Looks good to me

PR #610, opened 2024-02-17
feat: add drop local user operation endpoint

### LINKED ISSUE #567: Add the ability to drop a user operation in Rundler's local mempool
**Describe the feature**
Similar to https://docs.flashbots.net/flashbots-protect/cancellations

Users should be able to send a special user operation to cancel (drop) a user operation from the mempool. This is useful if users get stuck with under-estimated fees.

dancoombs:
Closes #567 

## Proposed Changes

  - Adds `rundler_dropLocalUserOperation` rpc endpoint
  - Requirements:
    - The user operation must contain a sender/nonce pair this is present in the local mempool.
    - The user operation must pass entrypoint.simulateValidation. I.e. it must have a valid signature and verificationGasLimit
    - The user operation must have zero values for: preVerificationGas, callGasLimit, calldata, and maxFeePerGas


crates/rpc/src/rundler.rs
```diff
@@ -14,32 +14,58 @@
...
 
 #[rpc(client, server, namespace = "rundler")]
 pub trait RundlerApi {
     /// Returns the maximum priority fee per gas required by Rundler
     #[method(name = "maxPriorityFeePerGas")]
     async fn max_priority_fee_per_gas(&self) -> RpcResult<U256>;
+
+    /// Drops a user operation from the local mempool.
+    ///
+    /// Requirements:
```
denniswon, line 38: Noted. So it costs gas for verification to cancel a UO. Why the 3rd requirement for this end point instead of just ignoring? Just curious

crates/rpc/src/rundler.rs
```diff
@@ -69,4 +99,56 @@ where
...
+        let uo: UserOperation = user_op.into();
+        let id = UserOperationId {
+            sender: uo.sender,
+            nonce: uo.nonce,
+        };
+
+        if uo.pre_verification_gas != U256::zero()
+            || uo.call_gas_limit != U256::zero()
+            || uo.call_data.len() != 0
+            || uo.max_fee_per_gas != U256::zero()
```
denniswon, line 124: Do these requirements exist for simulate uo operation?

denniswon: I mean simulation verification

denniswon (commented):
So uo hash is not needed for this api? What's the difference between uo hash bs uo Id (sender + nonce) for serving uniqueness?

Also after canceling the uo with nonce K, should nonce still be a new nonce value? (K+1) or K can still be the nonce of the next uo?

crates/rpc/src/rundler.rs
dancoombs: It doesn't cost any gas on chain, but the UO must contain a high enough gas limit such that we can simulate it. This value is set by the user because its part of the UO that is signed.

crates/rpc/src/rundler.rs
dancoombs: The reason for these requirements is to ensure that the signed UO is not executable onchain. That is, it should have 0 gas fees associated with it. The most important one here is `max_fee_per_gas == 0`, as that would cause the entrypoint to reject the UO. We could reduce this requirements set to just that.

Simulation verification when a normal UO is submitted doesn't have this requirement because the UO is meant to be included onchain.

dancoombs:
> So uo hash is not needed for this api? What's the difference between uo hash bs uo Id (sender + nonce) for serving uniqueness?

There is no API for a sender to lookup pending UOs. So if they find that they are stuck, i.e. getting replacement underpriced consistently, they can issue a cancel just knowing their address and their nonce. No need to also lookup a hash.

> Also after canceling the uo with nonce K, should nonce still be a new nonce value? (K+1) or K can still be the nonce of the next uo?

K is still the nonce of the next UO, since nothing happened onchain.

dancoombs:
This is missing the 10 block delay that we discussed in the ticket to prevent spam, need to add that before merge.

dancoombs:
Also worth noting that this isn't a perfect drop mechanism. Its possible that the UO has already been bundled (or sent on the P2P network) before the drop.

0xfourzerofour (approved):
Looks good, just a nit

0xfourzerofour (approved):

PR #615, opened 2024-02-21
chore: update deps

dancoombs:
Update all deps except for jsonrpsee.

Jsonrpsee had a pretty significant [change](https://github.com/paritytech/jsonrpsee/blob/master/CHANGELOG.md#v0210---2023-12-13) between 0.20 and 0.21 that requires a new implementation for our metrics layer


0xfourzerofour (dismissed):
LGTM

alex-miao (approved):

PR #618, opened 2024-02-22
feat(metrics): add internal rpc metrics 

### LINKED ISSUE #588: Add metrics to ethers-rs network requests
**Describe the feature**
Instrument all ethers-rs network requests to capture:

* RPC method
* Return HTTP code
* Return JSON-RPC code
* Response time

0xfourzerofour:
Closes #588 

## Proposed Changes

  - Create a middleware to wrap the json rpc client. This will be invoked whenever the Provider is used inside rundler. 
  - Move provider creation to the `provider/ethers` module


crates/pool/src/task.rs
```diff
@@ -83,7 +83,7 @@ impl Task for PoolTask {
                 .map(|config| config.entry_point)
                 .collect(),
         };
-        let provider = eth::new_provider(&self.args.http_url, Some(self.args.http_poll_interval))?;
+        let provider = new_provider(&self.args.http_url, Some(self.args.http_poll_interval))?;
```
dancoombs, line 86: nit: for functions we've typically kept them scoped to make it clear they are being called from another module, so `rundler_provider::new_provider(...)`

crates/provider/src/ethers/metrics_middleware.rs
```diff
@@ -0,0 +1,83 @@
...
+        &self,
+        method: &str,
+        duration: Duration,
+        request: &Result<R, C::Error>,
+    ) {
+        let method_str = method.to_string();
+
+        match request {
+            Ok(_) => {
+                histogram!("internal_http_response_code", StatusCode::OK.as_u16() as f64, "method" => method_str.clone());
```
dancoombs, line 49: Should this be a counter instead of a histogram?

crates/provider/src/ethers/metrics_middleware.rs
```diff
@@ -0,0 +1,83 @@
...
+                histogram!("internal_rpc_response_code", 0.0, "method" => method_str.clone());
+            }
+            Err(error) => match error {
+                HttpClientError::ReqwestError(req_err) => {
+                    histogram!("internal_http_response_code", req_err.status().unwrap_or_default().as_u16() as f64, "method" => method_str.clone());
+                }
+                HttpClientError::JsonRpcError(rpc_err) => {
+                    histogram!("internal_rpc_response_code", rpc_err.code as f64, "method" => method_str.clone());
```
dancoombs, line 57: After this is switched to a counter, the concern here is that this creates a large cardinality in the metics, which prom doesn't like. In `aa-service` we did this:

```java
      if (status == null) {
        metricsWithDimensions.inc(RESPONSE_STATUS_OTHER);
      } else if (status.getCode() >= 500 && status.getCode() <= 599) {
        metricsWithDimensions.inc(RESPONSE_STATUS_5XX);
      } else if (status.getCode() >= 400 && status.getCode() <= 499) {
        metricsWithDimensions.inc(RESPONSE_STATUS_4XX);
      } else if (status.getCode() >= 200 && status.getCode() <= 299) {
        metricsWithDimensions.inc(RESPONSE_STATUS_2XX);
      } else {
        metricsWithDimensions.inc(RESPONSE_STATUS_OTHER);
      }
```

which is pretty gross, but it keeps the cardinality down.

0xfourzerofour: I did something similar with RPC let me know what you think, I am not sure how I feel about using labels from an enum but I think it works a little cleaner

crates/provider/src/ethers/metrics_middleware.rs
```diff
@@ -0,0 +1,134 @@
...
+        };
+
+        let rpc = match rpc_code {
+            x if x == -32000 => RpcCode::ParseError,
+            x if x == -32600 => RpcCode::InvalidRequest,
+            x if x == -32601 => RpcCode::MethodNotFound,
+            x if x == -32602 => RpcCode::InvalidParams,
+            x if x == -32603 => RpcCode::InternalError,
+            x if (-32099..=-32000).contains(&x) => RpcCode::ServerError,
+            _ => RpcCode::Success,
```
dancoombs, line 101: I think it should only be success if `x > 0`. If `x < 0`  we should use `Other`

crates/provider/src/ethers/metrics_middleware.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        let mut http_code = StatusCode::OK.as_u16() as u64;
+        let mut rpc_code = 0;
+
+        if let Err(error) = request {
+            match error {
+                HttpClientError::ReqwestError(req_err) => {
+                    http_code = req_err.status().unwrap_or_default().as_u16() as u64;
+                }
+
```
dancoombs, line 79: nit: extra new line

crates/provider/src/ethers/metrics_middleware.rs
```diff
@@ -0,0 +1,134 @@
...
+use ethers::providers::{HttpClientError, JsonRpcClient};
+use metrics::{histogram, increment_counter};
+use parse_display::Display;
+use reqwest::StatusCode;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use tokio::time::Instant;
+
+#[derive(Display, Deserialize)]
+#[display(style = "lowercase")]
+#[serde(rename_all = "snake_case")]
```
dancoombs, line 27: Why do we need both of these?

crates/provider/src/ethers/metrics_middleware.rs
dancoombs: Seems reasonable to me

crates/provider/src/ethers/metrics_middleware.rs
0xfourzerofour: We actually do not, I took it from somewhere else in the codebase but I have removed now

crates/provider/src/ethers/metrics_middleware.rs
```diff
@@ -0,0 +1,133 @@
...
+
+use async_trait::async_trait;
+use ethers::providers::{HttpClientError, JsonRpcClient};
+use metrics::{histogram, increment_counter};
+use parse_display::Display;
+use reqwest::StatusCode;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use tokio::time::Instant;
+
+#[derive(Display, Deserialize)]
```
dancoombs, line 25: Do we still need `Display`? Confused why we need serde and display on these enums, which one does the metrics library use?

dancoombs: Then what is serde used for?

0xfourzerofour: serde is For Serializing and Deserializing back and for from rust data types to usable structures like json and yaml. But from what I understand if you want to_string you need to implement Display. The reason we need both here is because the container attribute need to access the Deserializer from the derive. Well at least thats what I think but when I remove it, it gives an error

dancoombs: I think you should be able to remove both `Deserialize` and `#[serde(...)]`, I don't see that being used anywhere. `Display` is being used for `to_string()` and you can add back that `#[display(style = "lowercase")]` if you want to lowercase (maybe use snake case?) the enums

0xfourzerofour: Yeh that its true, I have just changed

dancoombs (approved):
LGTM! This will be huge for observability

PR #622, opened 2024-02-23
feat(release): update version

0xfourzerofour:
Fixes for release pipeline to test workflows

### DIFF
--- a/.github/workflows/docker-release.yaml
+++ b/.github/workflows/docker-release.yaml
@@ -31,7 +31,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-          toolchain: 1.71.0
+          toolchain: 1.75.0
 
       - name: Install toolchain (nightly)
         run: rustup toolchain add nightly --component rustfmt --profile minimal
@@ -39,6 +39,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: true
+
+      - uses: taiki-e/install-action@cross
 
       - name: Login to Docker Hub
         uses: docker/login-action@v3
@@ -53,10 +55,8 @@ jobs:
 
       - name: Build and push image
         run: |
-          cargo install cross --git https://github.com/cross-rs/cross 
-
           if [ -n "${{ github.event.inputs.version }}" ]; then
-            make GIT_TAG="${{ github.event.inputs.version }}" docker-build 
+            sudo -E env "PATH=$PATH" make GIT_TAG="${{ github.event.inputs.version }}" docker-build 
           else
-            make docker-build-latest
+            sudo -E env "PATH=$PATH" make docker-build-latest
           fi
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ default-members = ["bin/rundler"]
 resolver = "2"
 
 [workspace.package]
-version = "0.1.0-beta"
+version = "0.1.0-rc0"
 edition = "2021"
 rust-version = "1.75"
 license = "MIT OR Apache-2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ default-members = ["bin/rundler"]
 resolver = "2"
 
 [workspace.package]
-version = "0.2.0-alpha.0"
+version = "0.1.0-beta"
 edition = "2021"
 rust-version = "1.75"
 license = "MIT OR Apache-2.0"
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,2 +1,2 @@
 [build]
-dockerfile = "Dockerfile.build"
+dockerfile = "./Dockerfile.build"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,18 @@
 # Adapted from https://github.com/paradigmxyz/reth/blob/main/Dockerfile
 # syntax=docker/dockerfile:1.4
 
+FROM ghcr.io/foundry-rs/foundry:latest as foundry
+
 FROM --platform=$TARGETPLATFORM rust:1.75.0 AS chef-builder
 
 # Install system dependencies
+COPY --from=foundry /usr/local/bin/forge /usr/local/bin/forge
+SHELL ["/bin/bash", "-c"]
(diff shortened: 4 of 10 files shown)


Dockerfile
```diff
@@ -1,19 +1,18 @@
 # Adapted from https://github.com/paradigmxyz/reth/blob/main/Dockerfile
 # syntax=docker/dockerfile:1.4
 
+FROM ghcr.io/foundry-rs/foundry:latest as foundry
```
dancoombs, line 4: Nice

dancoombs: Should we tag this to a specific foundry version?

bin/rundler/Cargo.toml
```diff
@@ -44,3 +44,5 @@ tracing.workspace = true
 tracing-appender = "0.2.2"
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.16", features = ["env-filter", "fmt", "json"] }
+openssl = { version = "0.10", features = ["vendored"] }
+
```
dancoombs, line 48: What is this needed for? We use `rustls` above

Cross.toml
```diff
@@ -1,2 +1,6 @@
 [build]
-dockerfile = "Dockerfile.build"
+dockerfile = "./Dockerfile.build"
+# pre-build = [
+#     "dpkg --add-architecture $CROSS_DEB_ARCH",
+#     "apt-get update && apt-get install --assume-yes libssl-dev:$CROSS_DEB_ARCH",
+# ]
```
dancoombs, line 6: nit: remove

bin/rundler/Cargo.toml
0xfourzerofour: This is also due to the openssl system dependency in ubuntu:20.04, using the `vendored` feature will compile and statically link the correct version that we need to get the dependencies of `rustls` to work. Downloading it via apt-get does not work and has missing cpp headers

dancoombs: Hm, I wasn't aware that `rustls` required any system dependencies? It seems like if openssl is needed then we're missing a feature flag in one of our dependences to have it use `rustls`

dancoombs: Try this diff:

```diff
diff --git a/bin/rundler/Cargo.toml b/bin/rundler/Cargo.toml
index 2d7e49f3..55399297 100644
--- a/bin/rundler/Cargo.toml
+++ b/bin/rundler/Cargo.toml
@@ -27,7 +27,7 @@ dotenv = "0.15.0"
 ethers.workspace = true
 itertools = "0.12.1"
 metrics = "0.22.1"
-metrics-exporter-prometheus = "0.13.1"
+metrics-exporter-prometheus = { version = "0.13.1", default-features = false, features = ["http-listener"] }
 metrics-process = "1.0.10"
 metrics-util = "0.16.2"
 paste = "1.0"
@@ -44,5 +44,3 @@ tracing.workspace = true
 tracing-appender = "0.2.2"
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.16", features = ["env-filter", "fmt", "json"] }
-openssl = { version = "0.10", features = ["vendored"] }
-
diff --git a/deny.toml b/deny.toml
index c6cb720f..b0069153 100644
--- a/deny.toml
+++ b/deny.toml
@@ -22,6 +22,7 @@ deny = [
     # Each entry the name of a crate and a version range. If version is
     # not specified, all versions will be matched.
     #{ name = "ansi_term", version = "=0.11.0" },
+    { name = "openssl" }
 ]
 # Certain crates/versions that will be skipped when doing duplicate detection.
 skip = []
```

dancoombs: We don't use the "push-gateway" feature of `metrics-exporter-prometheus`, which was pulling in openssl

0xfourzerofour: but it looks like `rustls-native-certs` is using openssl which is a dependency of hyper but Ill give it a go

.github/workflows/docker-release.yaml
```diff
@@ -53,10 +55,8 @@ jobs:
 
       - name: Build and push image
         run: |
-          cargo install cross --git https://github.com/cross-rs/cross 
-
           if [ -n "${{ github.event.inputs.version }}" ]; then
-            make GIT_TAG="${{ github.event.inputs.version }}" docker-build 
+            sudo -E env "PATH=$PATH" make GIT_TAG="${{ github.event.inputs.version }}" docker-build
```
dancoombs, line 59: Curious what sudo is needed for? Does this not work locally without sudo?

0xfourzerofour: This is because solc gets downloaded under the hood by forge when we build the contracts and we need sudo to save the solc program or we get a permission denied

bin/rundler/Cargo.toml
0xfourzerofour: looks like that worked nice

.github/workflows/docker-release.yaml
dancoombs: Doesn't that happen inside the build container?

dancoombs: Hm interesting, I can't get this to work locally without `sudo` either, even though I have solc installed

0xfourzerofour: Yeh I don't fully understand why its the case

dancoombs: Got it to work locally without sudo using:

```
CROSS_ROOTLESS_CONTAINER_ENGINE=1 make docker-build
```

dancoombs: Think this is okay to leave here, just need to remember to set that env var when running locally

dancoombs: Had to do this too https://docs.docker.com/desktop/containerd/#enable-the-containerd-image-store

Cargo.toml
```diff
@@ -7,7 +7,7 @@ default-members = ["bin/rundler"]
 resolver = "2"
 
 [workspace.package]
-version = "0.2.0-alpha.0"
+version = "0.1.0-beta"
```
dancoombs, line 10: Lets make this `0.1.0-rc0`

PR #633, opened 2024-03-05
feat(types): add entrypoint 0.7 contract types

dancoombs:
## Proposed Changes

  - Compiles both versions of the entry point contracts
  - Only consumes v0.6 for now, going to begin abstracting these in subsequent PRs


crates/types/build.rs
```diff
@@ -19,48 +19,146 @@ fn main() -> Result<(), Box<dyn error::Error>> {
...
 }
 
-fn abigen_of(contract: &str) -> Result<Abigen, Box<dyn error::Error>> {
-    Ok(Abigen::new(
-        contract,
-        format!("contracts/out/{contract}.sol/{contract}.json"),
-    )?)
+fn generate_v0_7_bindings() -> Result<(), Box<dyn error::Error>> {
+    run_command(
+        Command::new("forge")
```
0xfourzerofour, line 65: Nit: there is some repeated logic here, you could make the forge script more configurable so that you pass the v_6, v_7 and utils into the same function to build the command dynamically

0xfourzerofour (commented):
small nit about code reuse

crates/types/contracts/test/PrecompileAccountTest.sol
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.14;
 
-import "../src/PrecompileAccount.sol";
-import "account-abstraction/interfaces/UserOperation.sol";
+import "../src/v0_6/PrecompileAccount.sol";
```
alex-miao, line 4: Should tests also be split into 0.6/0.7

dancoombs: I tried to clean this up as much as I could in follow up PRs. We do need to use explicit types during tests, and it doesn't make sense to duplicate every test for v0.7. We might want to come up with a better approach for this going forward. But for now, going to mostly rely on the shared code on traits and specific tests for v0.7 types.

crates/types/build.rs
dancoombs: Good idea, updated!

PR #634, opened 2024-03-06
feat: WIP start splitting 0.6 and 0.7 logic

dancoombs:
THIS IS BEING REPLACED BY: #635


crates/builder/src/bundle_proposer.rs
```diff
@@ -99,7 +98,7 @@ where
     builder_index: u64,
     pool: C,
     simulator: S,
-    entry_point: E,
+    entry_point: Arc<E>,
```
dancoombs, line 101: TODO: figure this out, we should be consistent here, we don't need these Arc as everything inside the entry point class should be clone

crates/builder/src/bundle_proposer.rs
```diff
@@ -32,13 +32,12 @@ use mockall::automock;
...
 use rundler_sim::{
-    gas::{self, GasOverheads},
-    EntityInfo, EntityInfos, ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError,
+    gas, EntityInfo, EntityInfos, ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError,
     SimulationResult, SimulationViolation, Simulator, ViolationError,
 };
 use rundler_types::{
-    chain::ChainSpec, Entity, EntityType, EntityUpdate, EntityUpdateType, GasFees, Timestamp,
-    UserOperation, UserOpsPerAggregator,
+    chain::ChainSpec, Entity, EntityType, EntityUpdate, EntityUpdateType, GasFees, GasOverheads,
```
dancoombs, line 39: TODO: GasOverheads should move to chain config

crates/builder/src/bundle_proposer.rs
```diff
@@ -1226,24 +1218,12 @@ impl ProposalContext {
...
-
+fn get_gas_required_for_op(chain_spec: &ChainSpec, gas_spent: U256, op: &UserOperation) -> U256 {
     gas_spent
         + gas::user_operation_pre_verification_gas_limit(chain_spec, op, false)
-        + op.verification_gas_limit * 2
-        + op.call_gas_limit
-        + post_exec_req_gas
+        + op.total_verification_gas_limit()
+        + op.required_pre_execution_buffer()
+        + op.call_gas_limit()
```
dancoombs, line 1226: TODO: verify this change

crates/pool/src/server/remote/error.rs
```diff
@@ -244,13 +244,6 @@ impl From<MempoolError> for ProtoMempoolError {
 impl From<PrecheckViolation> for ProtoPrecheckViolationError {
     fn from(value: PrecheckViolation) -> Self {
         match value {
-            PrecheckViolation::InitCodeTooShort(length) => ProtoPrecheckViolationError {
```
dancoombs, line 247: TODO: removed these, should add check in RPC before sending to pool.

crates/pool/src/server/remote/protos.rs
```diff
@@ -36,7 +36,18 @@ pub const OP_POOL_FILE_DESCRIPTOR_SET: &[u8] =
 
 impl From<&RundlerUserOperation> for UserOperation {
     fn from(op: &RundlerUserOperation) -> Self {
-        UserOperation {
+        match op {
+            RundlerUserOperation::V0_6(op) => op.into(),
+            RundlerUserOperation::V0_7(_) => {
+                unimplemented!("V0_7 user operation is not supported")
```
dancoombs, line 42: TODO: implement

crates/provider/src/ethers/entry_point_v0_6.rs
```diff
@@ -20,61 +20,90 @@ use ethers::{
...
+    AggregatorOut, AggregatorSimOut, Provider,
 };
 
+const ARBITRUM_NITRO_NODE_INTERFACE_ADDRESS: Address = H160([
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xc8,
+]);
+
+const OPTIMISM_BEDROCK_GAS_ORACLE_ADDRESS: Address = H160([
+    0x42, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0F,
+]);
```
dancoombs, line 53: TODO: we should be getting this from ChainSpec

crates/provider/src/ethers/entry_point_v0_6.rs
```diff
@@ -20,61 +20,90 @@ use ethers::{
...
+
 const REVERT_REASON_MAX_LEN: usize = 2048;
 
+// TODO(danc):
+// - Modify this interface to take only `UserOperationV0_6` and remove the `into_v0_6` calls
+// - Places that are already in a V0_6 context can use this directly
+// - Implement a wrapper that takes `UserOperation` and dispatches to the correct entry point version
+//   based on a constructor from ChainSpec.
+// - Use either version depending on the abstraction of the caller.
```
dancoombs, line 62: TODO

crates/provider/src/ethers/entry_point_v0_6.rs
```diff
@@ -240,14 +281,119 @@ where
             .context("should compute balances")?;
         Ok(out.balances)
     }
+
+    async fn aggregate_signatures(
```
dancoombs, line 285: Moved these from provider as they are entry point dependent

crates/provider/src/traits/entry_point.rs
```diff
@@ -98,8 +120,38 @@ pub trait EntryPoint: Send + Sync + 'static {
     ) -> Result<ExecutionResult, String>;
 
     /// Get the deposit info for an address
-    async fn get_deposit_info(&self, address: Address) -> anyhow::Result<DepositInfo>;
+    async fn get_deposit_info(&self, address: Address) -> anyhow::Result<DepositInfoV0_6>;
```
dancoombs, line 123: TODO: this will need to be made generic, can't rely on the 0_6 version.

crates/provider/src/ethers/entry_point_v0_6.rs
```diff
@@ -87,6 +116,10 @@ where
         user_op: UserOperation,
         max_validation_gas: u64,
     ) -> anyhow::Result<TypedTransaction> {
+        let user_op = user_op
+            .into_v0_6()
```
0xfourzerofour, line 120: I wonder if there is a way we could use some generic magic on the function and implement `Into` for the different types of user operations. I have not thought about implementation and just spitballing but surely we would be able to use a `UO` generic an call into() instead of having multiple implementations

crates/provider/src/traits/entry_point.rs
dancoombs: Similar to `ExecutionResult`

crates/rpc/src/debug.rs
```diff
@@ -126,7 +127,7 @@ where
             .await
             .map_err(|e| rpc_err(INTERNAL_ERROR_CODE, e.to_string()))?
             .into_iter()
-            .map(|pop| pop.uo.into())
+            .map(|pop| UserOperationV0_6::from(pop.uo).into())
```
dancoombs, line 130: TODO: Should be able to map these to new RPC types that can handle both formats.

crates/rpc/src/eth/api.rs
```diff
@@ -64,7 +64,7 @@ impl Settings {
 
 #[derive(Debug)]
 struct EntryPointContext<P, E> {
-    gas_estimator: GasEstimatorImpl<P, E>,
+    gas_estimator: GasEstimatorV0_6<P, E>,
```
dancoombs, line 67: TODO: need to handle multiple types of gas estimators here, based on the entry point version. Likely easiest to just use dyn on the trait.

crates/sim/src/estimation/mod.rs
```diff
@@ -11,9 +11,87 @@
...
+    RevertInCallWithBytes(Bytes),
+    /// Other error
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+}
+
+/// Gas estimate for a user operation
+#[derive(Debug, Copy, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GasEstimate {
```
dancoombs, line 45: TODO: This won't work for 0.7 estimation as we need to return more fields. Rethink this interface

crates/provider/src/ethers/entry_point_v0_6.rs
0xfourzerofour: Just skimming here so dont have full context yet

crates/sim/src/gas/gas.rs
```diff
@@ -156,9 +130,9 @@ pub fn user_operation_gas_limit(
     paymaster_post_op: bool,
 ) -> U256 {
     user_operation_pre_verification_gas_limit(chain_spec, uo, assume_single_op_bundle)
-        + uo.call_gas_limit
-        + uo.verification_gas_limit
-            * verification_gas_limit_multiplier(assume_single_op_bundle, paymaster_post_op)
+        + uo.total_verification_gas_limit()
+        + uo.required_pre_execution_buffer()
+        + uo.call_gas_limit()
```
dancoombs, line 135: TODO: verify the logic changes in this file

crates/sim/src/precheck.rs
```diff
@@ -397,9 +389,6 @@ impl<P: Provider, E: EntryPoint> PrecheckerImpl<P, E> {
 /// All possible errors that can be returned from a precheck.
 #[derive(Clone, Debug, parse_display::Display, Eq, PartialEq, Ord, PartialOrd)]
 pub enum PrecheckViolation {
-    /// The init code is too short to contain a factory address.
-    #[display("initCode must start with a 20-byte factory address, but was only {0} bytes")]
-    InitCodeTooShort(usize),
```
dancoombs, line 402: TODO: check these in RPC now.

crates/provider/src/ethers/entry_point_v0_6.rs
dancoombs: We will need multiple implementations regardless because this pulls in the contract interface which is different for v0.6 and v0.7

PR #636, opened 2024-03-11
feat(rpc): introduce entry point routing

dancoombs:
Adds:

- Lots of new RPC types for 0.7
- A new abstraction for interacting with user operation events, which is used for getByHash and getReceipt.
- A new concept, entry point router, that the API uses to route requests to the correct EP (or scan all EPs), and hides type complexity


bin/tools/src/bin/get_example_ops.rs
```diff
@@ -11,28 +11,30 @@
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
 
-use dotenv::dotenv;
-use rundler_dev::DevClients;
-use rundler_rpc::RpcUserOperation;
+// use dotenv::dotenv;
```
dancoombs, line 14: TODO fix

crates/rpc/src/eth/events/v0_6.rs
```diff
@@ -0,0 +1,287 @@
...
+    UserOperation as UserOperationTrait, UserOperationVariant,
+};
+use rundler_utils::{eth, log::LogOnError};
+
+use super::UserOperationEventProvider;
+use crate::types::{RpcUserOperationByHash, RpcUserOperationReceipt};
+
+#[derive(Debug)]
+pub(crate) struct UserOperationEventProviderV0_6<P: Provider> {
+    chain_id: u64,
```
0xfourzerofour, line 40: Not a change but it might make sense to add a static function onto the provider when we create it and pass the chain_id from the spec into the constructor of the Provider as the struct should have access to the chain_id. That means we could just call it from the provider

crates/rpc/src/eth/events/v0_6.rs
```diff
@@ -0,0 +1,287 @@
...
+impl<P: Provider> UserOperationEventProvider for UserOperationEventProviderV0_6<P> {
+    async fn get_mined_by_hash(
+        &self,
+        hash: H256,
+    ) -> anyhow::Result<Option<RpcUserOperationByHash>> {
+        // Get event associated with hash (need to check all entry point addresses associated with this API)
+        let event = self
+            .get_event_by_hash(hash)
+            .await
+            .log_on_error("should have successfully queried for user op events by hash")?;
```
0xfourzerofour, line 56: we should try to proactively remove inline errors and move them into an errors.rs file to clean things up going forward

crates/rpc/src/eth/events/v0_6.rs
```diff
@@ -0,0 +1,287 @@
...
+        // get transaction receipt
+        let tx_receipt = self
+            .provider
+            .get_transaction_receipt(tx_hash)
+            .await
+            .context("should have fetched tx receipt")?
+            .context("Failed to fetch tx receipt")?;
+
+        // filter receipt logs
+        let filtered_logs = super::filter_receipt_logs_matching_user_op(&event, &tx_receipt)
```
0xfourzerofour, line 128: Just a general question, do you prefer importing types and traits at the top of the file and then for imported functions using the module::function_name? I have always liked just importing everything as I think it makes things a little cleaner but we should align on this to make sure we have consistency going forward

crates/rpc/src/eth/router.rs
```diff
@@ -0,0 +1,305 @@
...
+            "No entry point found for address: {:?}",
+            entry_point
+        )))
+    }
+
+    fn get_route(&self, entry_point: &Address) -> EthResult<&Arc<Box<dyn EntryPointRoute>>> {
+        let ep = self.get_ep_version(entry_point)?;
+
+        match ep {
+            EntryPointVersion::V0_6 => Ok(&self.v0_6.as_ref().unwrap().1),
```
0xfourzerofour, line 205: Might want to do a check on the unwrap here so there is not some unwrap on None error

0xfourzerofour (approved):
Overall looks good and some nice cleanups just a couple small comments

crates/rpc/src/eth/events/v0_6.rs
dancoombs: Could make sense, out of the scope of this PR though

crates/rpc/src/eth/events/v0_6.rs
dancoombs: Yeah I want to do a project at some point and start moving away from anyhow. Although I'm not sure there are may issues with using it in builder/pool/rpc as they're not really meant to be SDK-like crates, are more server level.

crates/rpc/src/eth/events/v0_6.rs
dancoombs: For functions from other crates I always at least leave one qualifier. For types and traits I typically import, unless there is a collision

crates/rpc/src/eth/router.rs
dancoombs: This should always unwrap because we did the version check. If it doesn't there is a code error and we'd want to panic.

PR #641, opened 2024-03-20
fix(sim): handle aa51 as OOG in estimation

dancoombs:
## Proposed Changes

  - Handles AA51 as out of gas during validation gas estimation
  - Also fixes the fee logic to not rely on the base fee, uses pre-1559 mechanics

### DIFF
--- a/crates/sim/src/estimation/estimation.rs
+++ b/crates/sim/src/estimation/estimation.rs
@@ -210,12 +210,13 @@ impl<P: Provider, E: EntryPoint> GasEstimatorImpl<P, E> {
 
         // Make one attempt at max gas, to see if success is possible.
         // Capture the gas usage of this attempt and use as the initial guess in the binary search
-
+        let fee = gas_fee
+            .checked_div(simulation_gas + op.pre_verification_gas)
+            .unwrap_or(U256::MAX);
         let initial_op = UserOperation {
             verification_gas_limit: simulation_gas,
-            max_fee_per_gas: gas_fee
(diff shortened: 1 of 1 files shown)


crates/sim/src/estimation/estimation.rs
```diff
@@ -210,12 +210,13 @@ impl<P: Provider, E: EntryPoint> GasEstimatorImpl<P, E> {
...
+        let fee = gas_fee
+            .checked_div(simulation_gas + op.pre_verification_gas)
+            .unwrap_or(U256::MAX);
         let initial_op = UserOperation {
             verification_gas_limit: simulation_gas,
-            max_fee_per_gas: gas_fee
-                .checked_div(simulation_gas + op.pre_verification_gas)
-                .unwrap_or(U256::MAX),
+            max_fee_per_gas: fee,
+            max_priority_fee_per_gas: fee,
```
adam-alchemy, line 219: Just out of curiosity: would this function the same if priority fee per gas was zero instead?

dancoombs: It wouldn't, this is specifically setting them equal so that it triggers this entrypoint logic in postop: 

```
https://github.com/eth-infinitism/account-abstraction/blob/fa61290d37d079e928d92d53a122efcc63822214/contracts/core/EntryPoint.sol#L604
```

adam-alchemy (approved):

mokok123 (approved):

0xfourzerofour (approved):

PR #649, opened 2024-03-26
feat: end to end entry point routing

dancoombs:
## Proposed Changes

  - Implement RPC events trait for v0.7 (most code shared)
  - Wire up entry point 0.7 in various places, only supports `UNSAFE` and a hardcoded gas estimator
  - WIP: needs to be cleaned up and tested


crates/rpc/src/eth/router.rs
```diff
@@ -60,6 +60,7 @@ impl EntryPointRouterBuilder {
                 route.version()
             );
         }
+        tracing::info!("Adding route for v0.7 entry point {:?}", route.address());
```
dancoombs, line 63: TODO remove

crates/builder/src/task.rs
```diff
@@ -255,6 +250,92 @@ where
         Box::new(self)
     }
 
+    async fn create_builders_v0_6<C, E>(
```
0xfourzerofour, line 253: We should add a todo here to merge these builder creator functions into one using generics based on the UO version once entrypoint v7 safe simulation is implemented

crates/pool/src/task.rs
```diff
@@ -182,18 +184,32 @@ impl PoolTask {
...
+                ep,
+                simulator,
+            )
+        } else {
+            panic!("V0_7 safe simulation not supported")
+        }
     }
 
-    async fn create_mempool_v0_6<P: Provider + Middleware>(
+    fn create_mempool_v0_6<P: Provider + Middleware>(
```
0xfourzerofour, line 212: Same thing here, it looks possible but need to play around with the code for this feature to make sure

crates/sim/src/simulation/mempool.rs
```diff
@@ -25,10 +25,20 @@ use crate::simulation::SimulationViolation;
 /// Typically read from a JSON file using the `Deserialize` trait.
 #[derive(Debug, Clone, Deserialize, Default)]
 pub struct MempoolConfig {
+    /// Entry point address this mempool is associated with.
+    #[serde(rename = "camelCase")]
+    pub(crate) entry_point: Address,
```
0xfourzerofour, line 30: Will we have to add multiple mempool configs per network based on entrypoint addresses now?

0xfourzerofour (approved):
Looks good, just added a couple comments to consolidate the code at a later date based on implementation

crates/builder/src/task.rs
dancoombs: Yeah I found the types a little difficult to work with to get a common function. Adding a TODO

crates/sim/src/simulation/mempool.rs
dancoombs: Added a TODO

crates/builder/src/task.rs
```diff
@@ -143,45 +147,36 @@ where
...
-                        ),
-                    )
-                    .await?
-                };
-                sender_handles.push(spawn_guard);
-                bundle_sender_actions.push(bundle_sender_action);
+            match ep.version {
+                EntryPointVersion::V0_6 => {
+                    let (handles, actions) = self
+                        .create_builders_v0_6(ep, Arc::clone(&provider), ep_v0_6.clone())
```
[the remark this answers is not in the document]
dancoombs: From an interface perspective I don't like this. Its possible in the future an entry point wouldn't have an inner provider, that is specific to the ethers implementation.

PR #652, opened 2024-03-28
fix(compliance): swap if statements

0xfourzerofour:
## Proposed Changes
  - Use same logic as before but check within current staked status (swap the if statments)

### DIFF
--- a/crates/sim/src/simulation/simulation.rs
+++ b/crates/sim/src/simulation/simulation.rs
@@ -405,16 +405,17 @@ where
                 match violation {
                     StorageRestriction::Allowed => {}
                     StorageRestriction::NeedsStake(addr, entity_type, slot) => {
-                        if let Some(et) = entity_type {
-                            if let Some(e_info) = entity_infos.get(et) {
-                                if !e_info.is_staked {
+                        if !entity_info.is_staked {
+                            if let Some(et) = entity_type {
+                                if let Some(e_info) = entity_infos.get(et) {
                                     let ent = Entity::new(et, e_info.address);
                                     entity_types_needing_stake
                                         .insert(ent, (addr, entity_type, slot));
                                 }
+                            } else {
+                                entity_types_needing_stake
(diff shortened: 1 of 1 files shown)


crates/sim/src/simulation/simulation.rs
```diff
@@ -405,16 +405,17 @@ where
                 match violation {
                     StorageRestriction::Allowed => {}
                     StorageRestriction::NeedsStake(addr, entity_type, slot) => {
-                        if let Some(et) = entity_type {
-                            if let Some(e_info) = entity_infos.get(et) {
-                                if !e_info.is_staked {
+                        if !entity_info.is_staked {
+                            if let Some(et) = entity_type {
+                                if let Some(e_info) = entity_infos.get(et) {
```
dancoombs, line 410: What happens if entity infos doesn't contain the entity type? Can that happen? Should we assert that it doesn't?

crates/sim/src/simulation/simulation.rs
```diff
@@ -405,16 +405,17 @@ where
                 match violation {
                     StorageRestriction::Allowed => {}
                     StorageRestriction::NeedsStake(addr, entity_type, slot) => {
-                        if let Some(et) = entity_type {
-                            if let Some(e_info) = entity_infos.get(et) {
-                                if !e_info.is_staked {
+                        if !entity_info.is_staked {
```
dphilipson, line 408: I might be missing context, but what's this check for? My understanding is if we hit this `NeedsStake` branch, then that means the entity with type `entity_type` needs stake. `entity_info` is referring to the entity of the current phase, which may not be the same type as `entity_type`. Why do we care if the entity of the current phase is staked?

crates/sim/src/simulation/simulation.rs
dphilipson: That would mean something has gone very wrong in either our or the entry point logic, but it wouldn't hurt to have a `bail!` to notify us if that happens.

PR #659, opened 2024-04-01
feat(simulation): v0.7 verification gas estimation

dphilipson:
Factors out the verification gas estimation logic, which performs a binary search by making many `eth_call`s in a row, into a new trait, `VerificationGasEstimator`, then provide implementations for v0.6 and v0.7.

Some notable fixes along the way:

* Change the Rundler settings options around gas limits. There are now separate gas limits for all four things requiring gas in v0.7.
* Fixed some logic around v0.7's `simulateHandleOp`, which, unlike its counterpart in v0.6, returns its results normally rather than by reverting.


crates/sim/src/estimation/v0_7.rs
```diff
@@ -62,36 +73,140 @@ where
...
+        }
+
+        let get_op_with_limit = |op: UserOperation, args: GetOpWithLimitArgs| {
+            let GetOpWithLimitArgs { gas, fee } = args;
+            UserOperation {
+                verification_gas_limit: self.settings.max_verification_gas.into(),
+                max_fee_per_gas: fee,
+                max_priority_fee_per_gas: fee,
+                paymaster_verification_gas_limit: gas,
+                paymaster_post_op_gas_limit: U128::zero(),
```
dphilipson, line 193: Might need to change this because paymasters could reject operations that don't have gas for post-op.

bin/rundler/src/cli/mod.rs
```diff
@@ -130,12 +130,39 @@ pub struct CommonArgs {
...
+    max_paymaster_verification_gas: u64,
+
+    #[arg(
+        long = "max_paymaster_post_op_gas",
+        name = "max_paymaster_post_op_gas",
+        default_value = "3000000",
+        env = "MAX_PAYMASTER_POST_OP_GAS",
+        global = true
+    )]
+    max_paymaster_post_op_gas: u64,
```
dancoombs, line 164: I'm not sure we need granular configuration for these. Maybe just have these use `MAX_VERIFICATION_GAS` and `MAX_CALL_GAS` respectively?

crates/provider/src/ethers/entry_point/v0_6.rs
```diff
@@ -16,7 +16,7 @@ use std::sync::Arc;
 use anyhow::Context;
 use ethers::{
     abi::AbiDecode,
-    contract::{ContractError, FunctionCall},
+    contract::{ContractError, EthCall as _, FunctionCall},
```
dancoombs, line 19: nit: why the `as _`? Should be able to just import this trait (unless I'm missing some name collision)

crates/provider/src/ethers/entry_point/v0_7.rs
```diff
@@ -333,13 +335,15 @@ where
         }
     }
 
+    // Always returns `Err(String)`. The v0.7 entry point does not use reverts
+    // to indicate successful simulations.
```
dancoombs, line 339: Mentioned this in a chat, I think we should return a more structured error here instead of a string. Something that lets us separate the revert message from revert bytes, such that we can return the bytes in the RPC "data" field

crates/provider/src/ethers/entry_point/v0_7.rs
```diff
@@ -371,12 +387,59 @@ where
...
+{
+    fn get_simulate_op_spoofed_state(&self, base_state: &spoof::State) -> spoof::State {
+        let mut spoof_ep = base_state.clone();
+        spoof_ep
+            .account(self.address())
+            .code(ENTRYPOINTSIMULATIONS_DEPLOYED_BYTECODE.clone());
+        spoof_ep
+    }
+
+    fn to_common_execution_result(execution_result: ExecutionResultV0_7) -> ExecutionResult {
```
dancoombs, line 419: Can you not use the `From<>` impl defined below?

crates/sim/src/estimation/v0_7.rs
```diff
@@ -62,36 +73,140 @@ where
...
+                .max_fee_per_gas(fee)
+                .max_priority_fee_per_gas(fee)
+                .paymaster_verification_gas_limit(paymaster_verification_gas_limit)
+                .paymaster_post_op_gas_limit(U128::zero())
+                .call_gas_limit(U128::zero())
+                .build()
+        };
+
+        self.verification_gas_estimator
+            .estimate_verification_gas(
```
dancoombs, line 167: This is awesome!

crates/sim/src/estimation/v0_7.rs
```diff
@@ -62,36 +73,140 @@ where
...
+                settings.max_call_gas.into(),
+                settings.max_verification_gas.into(),
+                settings.max_paymaster_verification_gas.into(),
+                settings.max_paymaster_post_op_gas.into(),
+            )
+            .pre_verification_gas(pre_verification_gas)
+            .build();
+
+        let verification_future = self.estimate_verification_gas(&op, block_hash, &state_override);
+        let paymaster_verification_future =
```
dancoombs, line 89: Not a change, just a comment. This is essentially doubling the cost of this API to estimate gas for a paymaster, we should consider the implications here.

My thoughts:

- Allow for the user to provide an already known paymaster gas limit that will skip the paymaster estimation step
- Same could go for validation
- This could drastically reduce the amount of eth_calls that we need to make to do estimation

crates/types/src/user_operation/v0_7.rs
```diff
@@ -389,6 +418,63 @@ impl UserOperationOptionalGas {
...
+                // Cap their values to the gas limits from settings
+                call_gas_limit: self
+                    .call_gas_limit
+                    .unwrap_or(max_call_gas)
+                    .min(max_call_gas),
+                verification_gas_limit: self
+                    .verification_gas_limit
+                    .unwrap_or(max_verification_gas)
+                    .min(max_verification_gas),
+                // These aren't used in gas estimation, set to if unset 0 so that there are no payment attempts during gas estimation
```
dancoombs, line 450: nit: this comment isn't totally valid anymore, maybe just remove the comment? The logic should be clear on how they're used in the estimation function.

crates/sim/src/estimation/estimate_verification_gas.rs
```diff
@@ -0,0 +1,231 @@
...
+                Err(anyhow!(
+                    "simulateHandleOp succeeded but should always revert. Make sure the entry point contract is deployed and the address is correct"
+                ))?;
+            }
+        } else if let Some(message) = self
+            .entry_point
+            .decode_simulate_handle_ops_revert(gas_used.result)
+            .err()
+        {
+            return Err(GasEstimationError::RevertInValidation(message));
```
dancoombs, line 129: This would be a great place to parse different validation errors and return more structured error messages

dancoombs (commented):
This is great! Awesome job figuring out a way to share that binary search logic across both entry points and both v0.7 verification fields.

crates/provider/src/ethers/entry_point/v0_6.rs
dphilipson: Yea, no reason, I guess I was thinking that reducing names in scope is marginally better. I can take these out.

crates/types/src/user_operation/v0_7.rs
```diff
@@ -28,7 +28,8 @@ const ENTRY_POINT_INNER_GAS_OVERHEAD: U256 = U256([10_000, 0, 0, 0]);
 /// User Operation for Entry Point v0.7
 ///
 /// Offchain version, must be packed before sending onchain
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+#[non_exhaustive] // Prevent instantiation except with UserOperationBuilder
```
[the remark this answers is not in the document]
dphilipson: Yeah, I actually just learned about it too. The old way of doing this was to add a private unit field, e.g.
```
  _private_instantiation: ()
```
since structs with private fields can't be directly instantiated. When I wrote that, Clippy suggested using `#[non_exhaustive]` instead.

crates/provider/src/ethers/entry_point/v0_7.rs
dphilipson: Oh yeah, I didn't see that. Thanks!

crates/provider/src/ethers/entry_point/v0_7.rs
```diff
@@ -333,13 +335,15 @@ where
...
     fn decode_simulate_handle_ops_revert(
         &self,
         revert_data: Bytes,
     ) -> Result<ExecutionResult, String> {
-        if let Ok(result) = ExecutionResultV0_7::decode(&revert_data) {
-            Ok(result.into())
-        } else if let Ok(failed_op) = FailedOp::decode(&revert_data) {
+        if let Ok(failed_op) = FailedOp::decode(&revert_data) {
+            Err(failed_op.reason)
+        } else if let Ok(failed_op) = FailedOpWithRevert::decode(&revert_data) {
```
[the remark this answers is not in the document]
dphilipson: Good idea. This ended up touching a lot of files, so I did it in a separate PR: https://github.com/alchemyplatform/rundler/pull/675

dancoombs (approved):
:shipit:

crates/sim/src/estimation/estimate_verification_gas.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        let mut max_failure_gas = 1;
+        let mut min_success_gas = self.settings.max_verification_gas;
+
+        if gas_used.gas_used.cmp(&U256::from(u64::MAX)).is_gt() {
+            return Err(GasEstimationError::RevertInValidation(
+                "gas_used cannot be larger than a u64 integer".to_string(),
+            ));
+        }
+        let mut guess = gas_used.gas_used.as_u64() * 2;
```
0xfourzerofour, line 177: might be worth using `saturating_mul` here for safety

crates/sim/src/estimation/estimate_verification_gas.rs
```diff
@@ -0,0 +1,231 @@
...
+        while (min_success_gas as f64) / (max_failure_gas as f64)
+            > (1.0 + GAS_ESTIMATION_ERROR_MARGIN)
+        {
+            num_rounds += 1;
+            if run_attempt_returning_error(guess).await? {
+                min_success_gas = guess;
+            } else {
+                max_failure_gas = guess;
+            }
+            guess = (max_failure_gas + min_success_gas) / 2;
```
0xfourzerofour, line 188: saturating_add and saturating_div here. We recently had an integer overflow recently in some estimation code so its better to stay on the safer side even though an overflow is rare with the numbers here

0xfourzerofour (dismissed):
Looks good, @dancoombs covered most of the concerns I was looking at. I just had a couple of small changes to reduce integer overflow risks

crates/sim/src/estimation/estimate_verification_gas.rs
dphilipson: Put in `saturating_add`. Didn't do `saturating_div` because as far as I can tell it doesn't do anything differently than regular `div` for unsigned integers, but would be very interested to know if it is.

dancoombs (approved):
🚀

PR #662, opened 2024-04-02
refactor: create context interface for simulation

dancoombs:
## Proposed Changes

  - Refactor tracing and context providing from simulation
  - Introduce a `ValidationContextProvider` trait.
  - This trait can be implemented for both v0.6 and v0.7 to surface all of the tracing information needed to compile the simulation violations.
  - The `ValidationContext` result is likely to change when v0.7 is implemented, but wanted to do this refactor first to limit the diff.
  - The trait also exposes a `get_specific_violations` function for any entry point specific logic (likely only to be used for v0.6)

### DIFF
--- a/crates/sim/src/simulation/context.rs
+++ b/crates/sim/src/simulation/context.rs
@@ -1,0 +1,176 @@
+// This file is part of Rundler.
+//
+// Rundler is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Lesser General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version.
+//
+// Rundler is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+// without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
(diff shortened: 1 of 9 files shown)


crates/sim/src/simulation/context.rs
```diff
@@ -0,0 +1,176 @@
...
+        op: Self::UO,
+        block_id: BlockId,
+    ) -> Result<ValidationContext<Self::UO>, ViolationError<SimulationViolation>>;
+
+    /// Get the violations specific to the particular entry point this provider targets.
+    fn get_specific_violations(
+        &self,
+        _context: &ValidationContext<Self::UO>,
+    ) -> Vec<SimulationViolation> {
+        vec![]
```
0xfourzerofour, line 113: should this be statically returning an empty vec?

0xfourzerofour (approved):
Looks good, just a question about one of the trait functions

crates/sim/src/simulation/context.rs
dancoombs: I originally thought it would, but then I found reasons to have overrides for both context impls. This will be removed in the follow up PR.

PR #664, opened 2024-04-03
feat(sim): implement simulation for v0.7 entrypoint using reference tracer

dancoombs:
## Proposed Changes

- Use the reference tracer to implement a simulation context provider for the v0.7 entrypoint.

TODO:

- [x] 2 more failing spec tests to work through
- [x] Make sure the v0.6 spec tests still pass
- [x] Allow RIP-7212 (reference tracer blocks this)
- [x] ~~OP-054 and OP-061 are currently always attributed to the account, can we fix this?~~ https://github.com/alchemyplatform/rundler/issues/665
- [ ] The spec tests don't necessarily test every rule. We should work through ERC-7562 and make sure we have coverage for every rule that simulation should be catching.


crates/rpc/src/types/v0_7.rs
```diff
@@ -95,11 +95,57 @@ impl FromRpc<RpcUserOperation> for UserOperation {
...
+    verification_gas_limit: U128,
+    pre_verification_gas: U256,
+    max_priority_fee_per_gas: U128,
+    max_fee_per_gas: U128,
+    factory: String, // if empty return "0x"
+    factory_data: Bytes,
+    paymaster: String, // if empty return "0x"
+    paymaster_verification_gas_limit: U128,
+    paymaster_post_op_gas_limit: U128,
+    paymaster_data: Bytes,
```
dancoombs, line 115: I'd prefer if all of these were optional, as opposed to returning "0x". The converse, inputting "paymaster: '0x'" will get rejected as it will fail address parsing. 

Either remove this, or fix the v0.7 `RpcUserOperation` to use strings for address and manually parse when converting to a `UserOperation`.

crates/rpc/src/types/mod.rs
```diff
@@ -128,12 +129,29 @@ impl FromRpc<RpcUserOperation> for UserOperationVariant {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, Eq, PartialEq)]
+#[serde(untagged)]
+// Return value for a user operation
+pub(crate) enum RpcUserOperationReturn {
+    V0_6(RpcUserOperationV0_6),
```
0xfourzerofour, line 136: Should these be named the same or is one of them used for multiple rpc uses and not just a return?

crates/sim/tracer/src/validationTracerV0_7.ts
```diff
@@ -0,0 +1,365 @@
...
+//
+// You should have received a copy of the GNU General Public License along with Rundler.
+// If not, see https://www.gnu.org/licenses/.
+
+
+// Adapted with minimal changes from: https://github.com/eth-infinitism/bundler/blob/main/packages/validation-manager/src/BundlerCollectorTracer.ts
+
+
+// javascript code of tracer function
+// NOTE: we process this locally for hardhat, but send to geth for remote tracing.
```
0xfourzerofour, line 19: nit: we probably don't need the hardhat information as we do not really use it as a team

test/spec-tests/local/run-spec-tests-v0_6.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -x
+cd "$(dirname "$0")"
+
+# ASSUMES THAT YOU'VE INSTALLED THE DEPENDENCIES FOR THE SPEC TESTS
+# (cd bundler-spec-tests && pdm install && pdm run update-deps)
+
+cargo build
```
0xfourzerofour, line 8: After this is merged I will add 0.7 to the compliance ci

0xfourzerofour (commented):
Looks good, I am sure some of the spec rules that need to be implemented can be parallelized within the team

crates/rpc/src/types/mod.rs
dancoombs: `RpcUserOperation` and `RpcUserOperationReturn` are the same for v0.6, so I just used the same type

crates/sim/tracer/src/validationTracerV0_7.ts
dancoombs: I'm not sure this comment means anything for us, just was copied from their implementation but it doesn't change any of the code.

crates/rpc/src/types/v0_7.rs
dphilipson: Yeah, `paymaster: "0x"` feels wrong to me. I'm more okay with "0x" for `Bytes` because it represents the empty value for `Bytes`, but it doesn't represent an address which has fixed length. I agree that `paymaster` should be omitted, or if it really has to have some value then maybe possibly the zero address ("0x00..0" with forty zeroes).

dphilipson (approved):
Looks great!

crates/rpc/src/types/v0_7.rs
dancoombs: Yeah agreed, going to update this to just return null. Updating the spec schema to allow null for these fields.

dancoombs: Fixed

0xfourzerofour (approved):
LGTM

PR #666, opened 2024-04-04
feat(sim): cleanups for v0.7 sim

dancoombs:
## Proposed Changes

  - Went through ERC-7562 and made sure we have coverage for every OP/STO rule that simulation should be catching.


0xfourzerofour (approved):
Looks good

PR #670, opened 2024-04-11
fix(sim): use H256 instead of U256 for expected storage

dancoombs:
## Proposed Changes

  - A previous change moved this erroneously to U256 for v0.7, move it back and provide a simple insert function.

### DIFF
--- a/crates/sim/src/types.rs
+++ b/crates/sim/src/types.rs
@@ -14,13 +14,13 @@
 use std::collections::{btree_map, BTreeMap};
 
 use anyhow::bail;
-use ethers::types::{Address, U256};
+use ethers::types::{Address, H256, U256};
 use serde::{Deserialize, Serialize};
 
 /// The expected storage values for a user operation that must
 /// be checked to determine if this operation is valid.
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


crates/sim/src/types.rs
```diff
@@ -14,13 +14,13 @@
...
 use anyhow::bail;
-use ethers::types::{Address, U256};
+use ethers::types::{Address, H256, U256};
 use serde::{Deserialize, Serialize};
 
 /// The expected storage values for a user operation that must
 /// be checked to determine if this operation is valid.
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
-pub struct ExpectedStorage(pub BTreeMap<Address, BTreeMap<U256, U256>>);
+pub struct ExpectedStorage(pub BTreeMap<Address, BTreeMap<H256, H256>>);
```
alex-miao, line 23: What was the issue here that required this conversion? Saw there was something about unmarshaling hex strings, is this because of some difference from the underlying u64 vs u8 representation of U256 vs. H256

dancoombs: Its because of how the json serialization is implemented for H256 vs U256. Geth requires the full 32 bytes in hex, which is how H256 serializes, whereas U256 uses the least number of characters.

alex-miao (approved):

PR #674, opened 2024-04-12
feat(sim): sto-022 factory or entity needs stake, with cleanups

### LINKED ISSUE #660: [simulation] STO-022: Either entity or factory needs to be staked if associated storage access during deployment
**Describe the bug**
Either paymaster or factory needs to be staked if associated storage access during deployment.

This is only an issue with our error reporting. Currently we report that the non-account entity must be staked if it accessing associated storage during account deployment. This should be reported as EITHER the entity OR the factory.

We correctly report that the factory needs to be staked if the factory or the account access associated storage during deployment.

This is currently not very clear in the STO-022 description in ERC-7562, but came from a discussion with the ERC team. They'll be updating the spec to be more clear.

dancoombs:
Closes #660

## Proposed Changes

  - Return either the factory or the accessing entity needs to be staked if accessing associated storage during a deploy
  - Lots of cleanups to make this easier to express

### DIFF
--- a/crates/sim/src/simulation/unsafe_sim.rs
+++ b/crates/sim/src/simulation/unsafe_sim.rs
@@ -18,8 +18,7 @@ use rundler_provider::{
     AggregatorOut, EntryPoint, Provider, SignatureAggregator, SimulationProvider,
 };
 use rundler_types::{
-    pool::SimulationViolation, EntityInfo, EntityInfos, UserOperation, ValidTimeRange,
-    ValidationError,
+    pool::SimulationViolation, EntityInfos, UserOperation, ValidTimeRange, ValidationError,
 };
 
 use crate::{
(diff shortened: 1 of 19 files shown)


dancoombs:
This is failing the spec test that is wrong for this, based on the conversation with the ERC-4337 team. I'll update the executor.

crates/sim/src/simulation/unsafe_sim.rs
```diff
@@ -123,24 +122,17 @@ where
...
-            }),
-            aggregator: validation_result.aggregator_info.map(|a| EntityInfo {
-                address: a.address,
-                is_staked: false,
-            }),
-        };
+        let mut entity_infos = EntityInfos::default();
+        entity_infos.set_sender(op.sender(), false);
+        if let Some(f) = op.factory() {
+            entity_infos.set_factory(f, false);
```
0xfourzerofour, line 128: this set function is nice

0xfourzerofour (approved):
LGTM this is nice

PR #675, opened 2024-04-12
feat(rpc): return validation revert data from rpc

dphilipson:
When validation fails in 0.7 in either simulation or gas estimation, the entry point now surfaces the inner revert which caused validation to fail. Surface this in the error response in the rpc, using the `data` field to return an object with the fields:

- `entry_point_reason`: the entry point string like "AA99 some failure". This is the only field populated in v0.6.
- `inner_revert_bytes`: the raw bytes of the inner revert message.
- `inner_revert_reason`: if the inner revert is caused by a Solidity `require` or `assert`, then the string message provided to it.

### DIFF
--- a/crates/sim/src/estimation/estimate_verification_gas.rs
+++ b/crates/sim/src/estimation/estimate_verification_gas.rs
@@ -1,6 +1,6 @@
-use std::{ops::Deref as _, sync::Arc};
+use std::{ops::Deref, sync::Arc};
 
-use anyhow::{anyhow, Context as _};
+use anyhow::{anyhow, Context};
 use async_trait::async_trait;
 use ethers::types::{spoof, Address, Bytes, H256, U128, U256};
 use rundler_provider::{EntryPoint, Provider, SimulateOpCallData, SimulationProvider};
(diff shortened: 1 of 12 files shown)


crates/sim/src/estimation/estimate_verification_gas.rs
```diff
@@ -144,22 +144,23 @@ where
...
-                {
-                    // This error occurs when out of gas, return false.
-                    Ok(false)
-                } else {
-                    // This is a different error, return it
-                    Err(GasEstimationError::RevertInValidation(error_message))
+            if let Some(revert) = revert {
+                if let Some(error_code) = revert.entry_point_error_code() {
+                    if error_code == "AA13"
+                        || error_code == "AA23"
```
0xfourzerofour, line 150: It might be worth adding each of these to a const slice for readability and just do a `.contains`.

0xfourzerofour (approved):
LGTM just a personal preference on the error code checks

dancoombs (approved):
Awesome work, super clean, thanks for all the cleanups!

crates/sim/src/estimation/estimate_verification_gas.rs
dphilipson: Oh good idea, thank you!

PR #676, opened 2024-04-12
chore(simulation): remove postop gas estimates

dphilipson:
We don't have a reasonable way to compute these and they are soon getting removed in an update to ERC-4337.


dphilipson:
For discussion: should we remove the postop estimate field now, or return a dummy value until the spec actually changes?

I'm inclined to remove it now (this PR) because I think returning incorrect data is worse than breaking the spec in a minor way.

dancoombs:
> For discussion: should we remove the postop estimate field now, or return a dummy value until the spec actually changes?
> 
> I'm inclined to remove it now (this PR) because I think returning incorrect data is worse than breaking the spec in a minor way.

Agreed with removing it now!

dancoombs (approved):

PR #679, opened 2024-04-12
fix(sim): use same method for expected storage in v0.7 from v0.6 tracer

dancoombs:
## Proposed Changes

  - The reference tracer tracks reads/writes within top level calls, not globally. This led to bugs when the factory would write to some slot and then the account would read it (i.e. a deploy). Our existing approach would capture the account read as an expected slot, and fail to send the bundle.
  - The fix uses the same approach from the v0.6 tracer to capture global accesses, only returning slots that were first read.

### DIFF
--- a/crates/sim/tracer/src/validationTracerV0_7.ts
+++ b/crates/sim/tracer/src/validationTracerV0_7.ts
@@ -46,8 +46,22 @@ export interface BundlerTracerResult {
    * values passed into KECCAK opcode
    */
   keccak: string[]
+
+  /**
+   * calls and returns, collected globally
+   */
   calls: Array<ExitInfo | MethodInfo>
+
+  /**
+   * logs, collected globally
+   */
   logs: LogInfo[]
+
+  /**
+   * expected storage slots, collected globally
+   */
+  expectedStorage: Record<string, Record<string, string>>
+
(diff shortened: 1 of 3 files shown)


crates/sim/tracer/src/validationTracerV0_7.ts
```diff
@@ -189,6 +228,22 @@ interface BundlerCollectorTracer extends LogTracer<BundlerTracerResult>, Bundler
     countSlot (list: { [key: string]: number | undefined }, key: any) {
       list[key] = (list[key] ?? 0) + 1
     },
+
+    computeIfAbsent<K extends keyof any, V>(
```
0xfourzerofour, line 232: is this function also defined in the 0.6 logic? If so we could add a common file to remove duplication

0xfourzerofour (approved):
looks good just code duplication concerns

crates/sim/tracer/src/validationTracerV0_7.ts
dancoombs: I had a hard time doing that with how the v0.7 tracer is set up vs the v0.6 tracer. Going to leave this for now as its a pretty simple function

PR #681, opened 2024-04-15
feat(proposer): add field in chainspec to limit size of bundle in bytes 

### LINKED ISSUE #485: [builder] Limit bundle size by number of bytes
**Describe the feature**
Some sequencers/networks have limitations on the maximum size of a transaction by number of bytes. 

On arbitrum sepolia we ran into:
```
{
  "timestamp": "2023-11-06T21:29:55.045843Z",
  "level": "ERROR",
  "fields": {
    "message": "Failed to send bundle. Will retry next block: Error {\n    context: \"should send conditional raw transaction to node\",\n    source: JsonRpcClientError(\n        JsonRpcError(\n            JsonRpcError {\n                code: -32000,\n                message: \"oversized data\",\n                data: None,\n            },\n        ),\n    ),\n}"
  },
  "target": "rundler::builder::bundle_sender"
}
```

as the maximum size of a transaction is [95KB](https://github.com/OffchainLabs/nitro/blob/fdd098e3d79b2cb0def875a2948e2556891602a7/execution/gethexec/sequencer.go#L97). 

Similar to how we limit bundle size by gas, we should optionally limit bundle size by bytes.

We should add a `MAX_BUNDLE_SIZE_BYTES` optional parameter that gets checked/enforced at the same places as the bundle gas limit.

0xfourzerofour:
Closes #485

## Proposed Changes

  - Add field to chain spec for a max bundle size in bytes
  - Values may need to be adjusted but these are some that I found online 
  
  https://github.com/ethereum-optimism/optimism/issues/289
  https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint
  
  I left the default at 100KB and polygon as 300KB as we have not had too many issues before with it

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -55,9 +55,10 @@ const BUNDLE_TRANSACTION_GAS_OVERHEAD_PERCENT: u64 = 5;
 /// 4 bytes for function signature
 /// 32 bytes for user op array offset
 /// 32 bytes for beneficiary
+/// 32 bytes for array count
 /// Ontop of this offset there needs to be another 32 bytes for each
 /// user operation in the bundle to store its offset within the array
-const BUNDLE_BYTE_OVERHEAD: u64 = 4 + 32 + 32;
+const BUNDLE_BYTE_OVERHEAD: u64 = 4 + 32 + 32 + 32;
 
 /// Size of word that stores offset of user op location
 /// within handleOps `ops` array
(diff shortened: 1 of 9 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -367,6 +367,7 @@ where
 
         let ov = GasOverheads::default();
         let mut gas_spent = ov.transaction_gas_overhead;
+        let mut constructed_bundle_size = 0;
```
alex-miao, line 370: Is there some fixed overhead in the size of a bundle outside of ops themselves

dancoombs: Yeah there is, we need to account for it

dancoombs (approved):
:shipit:

PR #682, opened 2024-04-16
refactor(simulation): use abi-generated EntryPoint bytecode

dphilipson:
Rather than hardcoding v0.6 entry point bytecode in two places, one with the init code and one with the deployed code, use abigen to generate Rust code for the v0.6 `EntryPoint.sol` and use the bytecode that was generated.

This makes v0.6 `IEntryPoint.sol` unnecessary, to remove it and convert all occurrences of `v0_6::i_entry_point::IEntryPoint` to `v0_6::entry_point::EntryPoint`.


dancoombs (approved):
LGTM

PR #686, opened 2024-04-18
feat(rpc): add distinct types for EP version gas estimations

dancoombs:
## Proposed Changes

  - Add types for v0.6 and v0.7 gas estimation and only return relevant fields.

### DIFF
--- a/crates/rpc/src/eth/router.rs
+++ b/crates/rpc/src/eth/router.rs
@@ -23,7 +23,10 @@ use rundler_types::{
 use super::events::UserOperationEventProvider;
 use crate::{
     eth::{error::EthResult, EthRpcError},
-    types::{RpcGasEstimate, RpcUserOperationByHash, RpcUserOperationReceipt},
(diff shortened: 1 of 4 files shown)


crates/rpc/src/eth/router.rs
```diff
@@ -143,15 +146,24 @@ impl EntryPointRouter {
...
                         "Invalid user operation for entry point: {:?}",
                         entry_point
                     )));
                 }
-                &self.v0_6.as_ref().unwrap().1
+
+                let e = self
+                    .v0_6
+                    .as_ref()
+                    .unwrap()
```
0xfourzerofour, line 161: Are there any cases where this unwrap try on a None or is it just an option in case someone wanted to run it themselves only with one entrypoint version?

dancoombs: The outer match statement protects against this

alex-miao (approved):

PR #687, opened 2024-04-18
fix(builder): add missing builder index to metrics

dancoombs:
## Proposed Changes

  - Builder index was missing from a few builder metrics, add it.


0xfourzerofour (approved):
LGTM

PR #689, opened 2024-04-18
feat(rpc): return internal errors and increment metric on panic

dancoombs:
## Proposed Changes

  - Previously panics would just drop the connection and silently occur
  - New behavior is to catch them, return an internal error, and increment a metric that can be alerted upon


0xfourzerofour (approved):
looks good

PR #697, opened 2024-04-23
fix(cli): enforce num keys greater than or equal to num builders

dancoombs:
## Proposed Changes

  - On startup check that the configured number of keys is greater than or equal to the total number of builders across the two entrypoints.


0xfourzerofour (approved):
lgtm

dphilipson (approved):

PR #699, opened 2024-04-24
fix(sim): handle explicit zero fees in 0.6 pre-verification gas

dphilipson:
Treat explicitly passed zero values the same as absent values for gas fees in `preVerificationGas` estimation in v0.6, as is consistent with estimating gas limits.


alex-miao (approved):

dancoombs (commented):
Change looks good, I think we need to make the same change to v0.7 as well.

0xfourzerofour (approved):

PR #701, opened 2024-04-26
feat(builder): support multiple flashbots builders

### LINKED ISSUE #283: builder: Send bundle to more builders on L1

dancoombs:
Closes #283

### DIFF
--- a/crates/builder/src/sender/flashbots.rs
+++ b/crates/builder/src/sender/flashbots.rs
@@ -26,45 +26,47 @@ use ethers::{
     middleware::SignerMiddleware,
     providers::{interval, JsonRpcClient, Middleware, Provider},
     types::{
-        transaction::eip2718::TypedTransaction, Address, Bytes, TransactionReceipt, TxHash, H256,
-        U256, U64,
+        transaction::eip2718::TypedTransaction, Address, TransactionReceipt, TxHash, H256, U256,
+        U64,
     },
+    utils,
 };
 use ethers_signers::Signer;
 use futures_timer::Delay;
 use futures_util::{Stream, StreamExt, TryFutureExt};
-use jsonrpsee::{
-    core::{client::ClientT, traits::ToRpcParams},
-    http_client::{transport::HttpBackend, HttpClient, HttpClientBuilder},
-};
+use jsonrpsee::core::traits::ToRpcParams;
 use pin_project::pin_project;
+use reqwest::{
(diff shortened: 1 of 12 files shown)


crates/builder/src/sender/mod.rs
```diff
@@ -82,100 +79,100 @@ pub(crate) trait TransactionSender: Send + Sync + 'static {
...
-                    return Err(SenderConstructorErrors::InvalidChainForSender(
-                        chain_spec.id,
-                        self.into_snake_case(),
-                    ));
-                }
-                TransactionSenderEnum::Flashbots(FlashbotsTransactionSender::new(client, signer)?)
+            Self::Flashbots(args) => {
+                let flashbots_signer = args
+                    .auth_key
+                    .parse::<LocalWallet>()
```
0xfourzerofour, line 157: Nit: I think you do not need to explicitly define the type as it should be inferred by the input type of the function

0xfourzerofour (approved):
Looks good just a tiny nit

crates/builder/src/sender/flashbots.rs
```diff
@@ -96,38 +97,97 @@ where
...
+struct Preferences {
+    fast: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    privacy: Option<OurPrivacy>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    validity: Option<Validity>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct OurPrivacy {
```
alex-miao, line 163: nit: why is this called `OurPrivacy` instead of just `Privacy`

alex-miao (approved):
Awesome, nice work figuring out the auth issues I was having with the Flashbots endpoint before

crates/builder/src/sender/mod.rs
dancoombs: Fixed

crates/builder/src/sender/flashbots.rs
dancoombs: Good catch, left over from some testing I was doing

0xfourzerofour (approved):

dphilipson (approved):

PR #703, opened 2024-05-02
fix(types): define entrypoint specific gas overheads

### LINKED ISSUE #638: Move `GasOverheads` to chain spec (or hardcode)
**Describe the feature**
`GasOverheads` is currently only used in its default form without any way to configure. We should either move these to hardcoded constants or move this into chain spec.

dancoombs:
Closes #638


0xfourzerofour (approved):
LGTM

PR #706, opened 2024-05-10
fix(provider): use state overrides for gas used, not ctor method

dancoombs:
## Proposed Changes

  - Old code used the constructor revert method to check gas used. This ran into the limitation that Geth limits the initCode supplied on contract creation to 48 KiB which is less than what it uses for its transaction limit, 128 KiB.
  - New code uses state overrides instead, which is the preferred method as it makes the contracts easier to reason about.
  - Added a check at the RPC layer for UO size
  - Added a setting to the RPC server for max body size as max transaction size * 2

### DIFF
--- a/crates/provider/src/traits/provider.rs
+++ b/crates/provider/src/traits/provider.rs
@@ -13,7 +13,7 @@
 
 //! Trait for interacting with chain data and contracts.
 
-use std::fmt::Debug;
+use std::{fmt::Debug, sync::Arc};
 
 use ethers::{
     abi::{AbiDecode, AbiEncode},
@@ -25,6 +25,7 @@ use ethers::{
 };
 #[cfg(feature = "test-utils")]
 use mockall::automock;
+use rundler_types::contracts::utils::get_gas_used::GasUsedResult;
 use serde::{de::DeserializeOwned, Serialize};
 
 use super::error::ProviderError;
@@ -127,4 +128,13 @@ pub trait Provider: Send + Sync + Debug + 'static {
 
     /// Get the logs matching a filter
     async fn get_logs(&self, filter: &Filter) -> ProviderResult<Vec<Log>>;
+
+    /// Measures the gas used by a call to target with value and data.
+    async fn get_gas_used(
+        self: &Arc<Self>,
+        target: Address,
+        value: U256,
+        data: Bytes,
+        state_overrides: spoof::State,
+    ) -> ProviderResult<GasUsedResult>;
 }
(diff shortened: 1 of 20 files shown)


crates/provider/src/traits/provider.rs
```diff
@@ -127,4 +128,13 @@ pub trait Provider: Send + Sync + Debug + 'static {
 
     /// Get the logs matching a filter
     async fn get_logs(&self, filter: &Filter) -> ProviderResult<Vec<Log>>;
+
+    /// Measures the gas used by a call to target with value and data.
+    async fn get_gas_used(
+        self: &Arc<Self>,
```
dancoombs, line 134: This wasn't ideal, but I needed an easy way to interface with the `GetGasUsed` contract which requires a `Arc<Middleware>`. Open to simple suggestions here as this does leak the ethers-rs specific impl into the interface.

0xfourzerofour (approved):
Looks good, I think the Arc<> stuff is ok for now, not ideal but can't think of a simple solution

crates/rpc/src/eth/api.rs
```diff
@@ -67,6 +69,13 @@ where
         op: UserOperationVariant,
         entry_point: Address,
     ) -> EthResult<H256> {
+        if op.single_uo_bundle_size_bytes() > self.chain_spec.max_transaction_size_bytes {
+            return Err(EthRpcError::InvalidParams(format!(
+                "User operation in bundle exceeds max transaciton size {}",
```
alex-miao, line 74: nit: typo in "transaction", also would it be helpful to also return the op size

crates/types/contracts/src/utils/GetGasUsed.sol
```diff
@@ -1,24 +1,33 @@
...
+    }
 
-    constructor(address target, uint256 value, bytes memory data) {
-        (uint256 gasUsed, bool success, bytes memory result) = getGas(target, value, data);
-        revert GasUsedResult(gasUsed, success, result);
+    /**
+     * Contract should not be deployed
+     */
+    constructor() {
+        require(block.number < 100, "should not be deployed");
```
alex-miao, line 17: Is there a reason for the block number check or can we just do `revert("should not be deployed");`

dancoombs: Actually not sure, I stole this from the [EntryPointSimulations](https://github.com/eth-infinitism/account-abstraction/blob/62e1cf39220321a6e1c7bef51e974c3bf8a4e2e2/contracts/core/EntryPointSimulations.sol#L38) code we already use. Figure we should keep it standard 🤷‍♂️.

crates/rpc/src/eth/api.rs
dancoombs: Good call, fixing

alex-miao (approved):

PR #707, opened 2024-05-16
fix(builder): filter fee before filtering by gas limit

dancoombs:
## Proposed Changes

  - Filter UOs by fee before applying the bundle gas limit 
  - Before, if there are a lot of underpriced UOs (specifically by base fee or PVG, as they're sorted by priority fee) then they'll block correctly priced UOs from making it to the simulation stage.
  - Now, we filter by fee prior to filtering by gas limit. It does add an extra async step, but will improve when we're doing gas limiting.

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -154,18 +154,9 @@ where
             self.fee_estimator.required_bundle_fees(required_fees)
         )?;
 
-        // Limit the amount of gas in the bundle
-        tracing::debug!("Starting bundle proposal with {} ops", ops.len(),);
+        tracing::info!("Starting bundle proposal with {} ops", ops.len());
 
-        // Do an initial filtering of ops that we want to simulate.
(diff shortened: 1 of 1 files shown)


alex-miao (approved):

crates/builder/src/bundle_proposer.rs
```diff
@@ -266,19 +279,17 @@ where
         }
     }
 
-    // Filter and simulate a single op. Returns None if the op should be skipped.
+    // Check fees for a single user op. Returns False if the op should be skipped.
```
dphilipson, line 282: Comment talks about "False" but it's still returning `None`.

dphilipson (approved):
LGTM!

crates/builder/src/bundle_proposer.rs
dancoombs: Fixed

alex-miao (approved):

PR #712, opened 2024-05-23
feat(simulation): check whether the op fits within a valid time range

0xfourzerofour:
## Proposed Changes

  - Add check during simulation for a valid time range and add a 15 second buffer

### DIFF
--- a/crates/types/src/pool/error.rs
+++ b/crates/types/src/pool/error.rs
@@ -152,6 +152,9 @@ pub enum SimulationViolation {
     /// The signature is invalid for the account
     #[display("invalid account signature")]
     InvalidAccountSignature,
+    /// The user operation has an invalid time range based on the `valid_until` and `valid_after` fields
+    #[display("User Operation expired or has an invalid time range")]
+    InvalidTimeRange,
     /// The signature is invalid for the paymaster
     #[display("invalid paymaster signature")]
     InvalidPaymasterSignature,
--- a/crates/types/src/pool/error.rs
+++ b/crates/types/src/pool/error.rs
@@ -14,7 +14,8 @@
 use ethers::types::{Address, U256};
 
 use crate::{
-    validation_results::ValidationRevert, Entity, EntityType, StorageSlot, ViolationOpCode,
+    validation_results::ValidationRevert, Entity, EntityType, StorageSlot, Timestamp,
+    ViolationOpCode,
 };
 
 /// Pool server error type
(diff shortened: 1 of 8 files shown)


crates/types/src/pool/error.rs
```diff
@@ -152,6 +152,9 @@ pub enum SimulationViolation {
     /// The signature is invalid for the account
     #[display("invalid account signature")]
     InvalidAccountSignature,
+    /// The user operation has an invalid time range based on the `valid_until` and `valid_after` fields
+    #[display("User Operation expired or has an invalid time range")]
+    InvalidTimeRange,
```
dancoombs, line 157: Can we return `valid_until` and `valid_after` as part of this error? Can put it in the `data` field of the RPC error

0xfourzerofour: Yeh I can add that

crates/types/src/validation_results.rs
```diff
@@ -259,6 +261,17 @@ pub struct ValidationReturnInfo {
     pub paymaster_context: Bytes,
 }
 
+const VALID_UNTIL_FUTURE_SECONDS: u64 = 15;
```
dancoombs, line 264: I thought I remember us having something similar to this elsewhere in the codebase. Are we checking these values elsewhere? Maybe in the builder?

dancoombs: Yeah found it https://github.com/alchemyplatform/rundler/blob/02af86989ec90045bbee7e2890062128665adf3b/crates/builder/src/bundle_proposer.rs#L53

dancoombs: Also we have a function that does this buffer logic: https://github.com/alchemyplatform/rundler/blob/02af86989ec90045bbee7e2890062128665adf3b/crates/builder/src/bundle_proposer.rs#L447

dancoombs: Lets consolidate this somewhere

0xfourzerofour: I added it to the userop types file and import it into both

dancoombs (approved):
LGTM!

We should add to our docs that we require a UO to be valid at least 1 minute in the future for acceptance into the mempool. This is important as an anti-DOS mechanism, we don't want to fill our mempool with UOs that are going to expire before bundling.

dancoombs (approved):

PR #715, opened 2024-06-04
fix(sim): handle v0.7 executeUserOp special call in call gas estimation

dancoombs:
## Proposed Changes

  - Handle the special call for `executeUserOp` during gas estimation.
  - This required modifications to the call gas proxy that are v0.7 specific.
  - Update the call gas specializations code to handle multiple proxy implementations.

### DIFF
--- a/crates/sim/src/estimation/v0_7.rs
+++ b/crates/sim/src/estimation/v0_7.rs
@@ -13,15 +13,25 @@
 
 use std::{cmp, ops::Add, sync::Arc};
 
-use ethers::types::{spoof, Bytes, H256, U128, U256};
+use ethers::{
+    contract::EthCall,
+    types::{spoof, Address, Bytes, H256, U128, U256},
+};
+use rand::Rng;
 use rundler_provider::{EntryPoint, L1GasProvider, Provider, SimulationProvider};
 use rundler_types::{
     chain::ChainSpec,
-    contracts::v0_7::entry_point_simulations,
+    contracts::v0_7::{
+        call_gas_estimation_proxy::{
+            EstimateCallGasArgs, EstimateCallGasCall, TestCallGasCall,
+            CALLGASESTIMATIONPROXY_DEPLOYED_BYTECODE,
+        },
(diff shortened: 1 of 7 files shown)


crates/types/contracts/src/v0_7/CallGasEstimationProxy.sol
```diff
@@ -0,0 +1,198 @@
...
+            let len := callData.length
+            if gt(len, 3) { methodSig := calldataload(callData.offset) }
+        }
+
+        bytes memory executeCall;
+        if (methodSig == IAccountExecute.executeUserOp.selector) {
+            executeCall = abi.encodeCall(IAccountExecute.executeUserOp, (userOp, userOpHash));
+        } else {
+            executeCall = callData;
+        }
```
dancoombs, line 170: New code that caused this whole debacle.

crates/types/contracts/src/v0_7/CallGasEstimationProxy.sol
```diff
@@ -0,0 +1,198 @@
+// SPDX-License-Identifier: UNLICENSED
```
dancoombs, line 1: There is likely an opportunity to DRY this up as its mostly copied from the old proxy with minimal modifications.

Not sure its worth it given that we rarely modify this. I'm going to leave it as is for now.

crates/sim/src/estimation/v0_7.rs
```diff
@@ -389,37 +399,96 @@ pub struct CallGasEstimatorSpecializationV07;
...
+        )
     }
 }
 
+/// Offset at which the proxy target address appears in the proxy bytecode. Must
+/// be updated whenever `CallGasEstimationProxy.sol` changes.
+///
+/// The easiest way to get the updated value is to run this module's tests. The
+/// failure will tell you the new value.
+const PROXY_TARGET_OFFSET: usize = 163;
```
0xfourzerofour, line 459: This same value is used for both 0.6 and 0.7 we might want to consolidate

0xfourzerofour (approved):
Looks good apart from a nit

PR #716, opened 2024-06-05
Bundle Sender State Machine Rework

### LINKED ISSUE #362: builder: Remove polling of mined transactions

dancoombs:
Closes #362 

## Proposed Changes

  - Remove polling of transaction status from the transaction tracker
  - Introduce a new state machine to bundle sender
    - Machine is driven by a "trigger" which depends on the bundling mode and the state
    - Each trigger causes a state update
    - State currently jumps between `Building` when there are no outstanding bundle transactions, and `Pending` when there are outstanding transactions.
  - Adds handling for `nonce too low` errors

## TODO (in follow up PRs)

- [x] Add a new state to handle "cancellations" where we replace any pending transactions with an empty transaction to fix the "stuck sender" and "replacement underpriced" cases.
- [x] Add tests for the new state machine
- [x] Add support for dropped transactions to certain senders to short circuit pending wait times if dropped.


dancoombs:
Merged https://github.com/alchemyplatform/rundler/pull/720 and https://github.com/alchemyplatform/rundler/pull/721 into this PR.

PR #720, opened 2024-06-11
feat(builder): cancel transaction when replacement underpriced

### LINKED ISSUE #550: [builder] Handle "replacement transaction underpriced"

dancoombs:
Closes #550 

## Proposed Changes

  - Define bundle abandonment as - there is a pending bundle but after fee increases there are no longer any high enough paying UOs to increase the fees.
  - Modify sender state machine to funnel every bundle abandonment to the same state - "replacement underpriced." That is, when a bundle is abandoned clear the transaction tracker and start from scratch. On networks where the previous txn may still be in the mempool, we will get a "replacement underpriced." On other networks, starting from scratch allows us to continue.
  - Handle "replacement underpriced" errors with cancellations.
  - For the flashbots sender, use a "soft cancellation", which is just an API call and doesn't require polling
  - For all other senders, use a "hard cancellation", which requires a transaction to be sent to clear the mempool.
  
  TODO (in follow up PRs):
- [x] Consider splitting the monolith state machine function to make it easier to grok.
- [ ] Consider delaying moving to the cancellation state by counting the number of "replacement underpriced" errors without landing a bundle. This will reduce the amount of wasted cancellation funds at the expense of being less reactive.
- [ ] Unit testing of `bundle_sender`

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -132,6 +132,12 @@ where
             // Waiting for a bundle to be mined
             // (wait_until_block, fee_increase_count)
             Pending(u64, u64),
+            // Cancelling the last transaction
+            // (fee_increase_count)
+            Cancelling(u64),
+            // Waiting for a cancellation transaction to be mined
+            // (wait_until_block, fee_increase_count)
+            CancelPending(u64, u64),
         }
 
         // initial state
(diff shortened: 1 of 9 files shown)


crates/builder/src/task.rs
```diff
@@ -376,7 +376,7 @@ where
                 // bug in the redis lock implementation that panics if connection
                 // takes longer than the TTL. Generally the TLL should be on the order of 10s of seconds
                 // so this should give ample time for the connection to establish.
-                Duration::from_millis(self.args.redis_lock_ttl_millis / 10),
+                Duration::from_millis(self.args.redis_lock_ttl_millis / 2),
```
dancoombs, line 379: TODO return this

crates/builder/src/bundle_sender.rs
```diff
@@ -132,6 +132,12 @@ where
             // Waiting for a bundle to be mined
             // (wait_until_block, fee_increase_count)
             Pending(u64, u64),
+            // Canceling the last transaction
+            // (fee_increase_count)
+            Canceling(u64),
```
dancoombs, line 137: TODO spelling

crates/builder/src/bundle_sender.rs
```diff
@@ -229,65 +236,194 @@ where
                 }
                 State::Pending(until, fee_increase_count) => {
                     sender_trigger.wait_for_block().await?;
+                    debug!(
```
dancoombs, line 239: Remove

crates/builder/src/bundle_sender.rs
```diff
@@ -229,65 +236,194 @@ where
...
+                        BuilderMetrics::increment_bundle_txn_fee_increases(
+                            self.builder_index,
+                            self.entry_point.address(),
+                        );
+                        state = State::Building(false, fee_increase_count + 1);
+                    }
+                }
+                State::Canceling(fee_increase_count) => {
+                    // cancel the transaction
+                    info!("Cancelling last transaction",);
```
dancoombs, line 325: Remove trailing comma

crates/builder/src/bundle_sender.rs
```diff
@@ -229,65 +236,194 @@ where
...
+                        .unwrap_or_default();
+
+                    let cancel_res = self
+                        .transaction_tracker
+                        .cancel_transaction(self.entry_point.address(), estimated_fees)
+                        .await;
+
+                    match cancel_res {
+                        Ok(Some(_)) => {
+                            info!("Transaction cancellation sent, waiting for confirmation");
```
dancoombs, line 340: Fix log

crates/builder/src/bundle_sender.rs
```diff
@@ -229,65 +236,194 @@ where
...
+
+                            state = State::CancelPending(
+                                sender_trigger.last_block.block_number
+                                    + self.settings.max_blocks_to_wait_for_mine,
+                                fee_increase_count,
+                            );
+                        }
+                        Ok(None) => {
+                            info!("Soft cancellation or no transaction to cancel, starting new bundle attempt");
+                            BuilderMetrics::increment_soft_cancellations_sent(
```
dancoombs, line 354: Fix metric name

PR #721, opened 2024-06-11
refactor(builder): large refactor of bundle sender state machine and metrics

dancoombs:
## Proposed Changes

  - Refactor state machine
  - Refactor metrics by saving an object with common components

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -57,10 +57,11 @@ pub(crate) struct BundleSenderImpl<UO, P, E, T, C> {
     beneficiary: Address,
     proposer: P,
     entry_point: E,
-    transaction_tracker: T,
+    transaction_tracker: Option<T>,
     pool: C,
     settings: Settings,
     event_sender: broadcast::Sender<WithEntryPoint<BuilderEvent>>,
+    metrics: BuilderMetrics,
     _uo_type: PhantomData<UO>,
 }
 
@@ -91,9 +92,6 @@ pub enum SendBundleResult {
         tx_hash: H256,
     },
     NoOperationsInitially,
-    NoOperationsAfterFeeIncreases {
-        attempt_number: u64,
-    },
     StalledAtMaxFeeIncreases,
     Error(anyhow::Error),
 }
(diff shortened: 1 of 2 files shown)


crates/builder/src/bundle_sender.rs
```diff
@@ -57,10 +57,11 @@ pub(crate) struct BundleSenderImpl<UO, P, E, T, C> {
     beneficiary: Address,
     proposer: P,
     entry_point: E,
-    transaction_tracker: T,
+    transaction_tracker: Option<T>,
```
0xfourzerofour, line 60: when would this be None?

crates/builder/src/bundle_sender.rs
```diff
@@ -464,48 +174,266 @@ where
...
+                TrackerUpdate::Mined {
+                    block_number,
+                    attempt_number,
+                    gas_limit,
+                    gas_used,
+                    tx_hash,
+                    nonce,
+                    ..
+                } => {
+                    // mined!
```
0xfourzerofour, line 289: nit: remove comment

crates/builder/src/bundle_sender.rs
```diff
@@ -464,48 +174,266 @@ where
...
+
+        Ok(())
+    }
+
+    async fn handle_cancelling_state(
+        &mut self,
+        state: &mut SenderMachineState<T>,
+        inner: CancellingState,
+    ) -> anyhow::Result<()> {
+        // cancel the transaction
```
0xfourzerofour, line 347: Also no sure if this is needed with the info message or put the comment above the function

crates/builder/src/bundle_sender.rs
```diff
@@ -124,313 +122,25 @@ where
...
-                                fee_increase_count + 1
-                            );
-                            state = State::Cancelling(fee_increase_count + 1);
-                        }
-                    }
-                }
-            }
+        // initial state
+        let mut state =
+            SenderMachineState::new(sender_trigger, self.transaction_tracker.take().unwrap());
```
0xfourzerofour, line 135: If this is always going to be Some, what is the ideology for making it optional?

crates/builder/src/bundle_sender.rs
```diff
@@ -124,313 +122,25 @@ where
...
 
-            // send result to manual caller
-            if let Some(res) = send_bundle_result.take() {
-                if let Some(r) = send_bundle_response.take() {
-                    if r.send(res).is_err() {
-                        error!("Failed to send bundle result to manual caller");
-                    }
+        loop {
+            match self.step_state(&mut state).await {
```
0xfourzerofour, line 138: if we are not doing anything with the Ok value, we could probably use an `is_err` of a `let if Err(e)` to remove unused lines

0xfourzerofour (approved):
Looks good just a few comments

crates/builder/src/bundle_sender.rs
dancoombs: The transaction tracker isn't `Clone`, and I want to be able to move it into the `SenderMachineState` when `send_bundles_in_loop` is called, as its easier to track there. This was the easiest way to do that without making any of the pub `new` functions async.

crates/builder/src/bundle_sender.rs
dancoombs: So that we can move out of the optional, the optional is left with `None` after this is called.

PR #722, opened 2024-06-12
fix(builder): rework raw sender to support dropped/conditional/split-…

### LINKED ISSUE #295: builder: Handle senders without pending transaction support
**Describe the feature**
Not all senders have the ability to check the status of a pending transaction.

Examples:
* Optimism
* Arbitrum
* Flashbots (when using non-Flashbots builders)

The builder should be able to handle this case, and instead of logging the transaction as dropped, it should resend after N blocks with a higher fee.

We can view senders that support pending transactions as a special case that lets us resend after < N blocks if we notice the transaction has specifically been dropped. If the sender does not support pending transactions, we will always wait N blocks. We will likely set N to 2 on most networks, Arbitrum being the special case.

### LINKED ISSUE #296: builder: Consistency issue on Arbitrum with conditional sender
**Describe the bug**
When using the conditional sender we use 1 RPC endpoint in the builder server, and a different endpoint in the transaction sender (Arbitrum's endpoint). This can lead to consistency issues if the server RPC gets notified of a nonce change prior to the conditional RPC processing the block change. The builder server uses the transaction sender to get the transaction status after the nonce change.

This leads the builder to believe that an external transaction was mined, and not the one its looking for the status for.

To solve this, we should ensure that the same RPC that is being used to track nonces is also being used to track transaction status. We should likely start viewing the conditional endpoints as purely submission endpoints and use the regular RPC for status tracking. We won't get pending transaction support this way, and will need to implement #295 first.

**To reproduce**
Send bundle transactions on Arbitrum Goerli. Reproduces about 50% of the time.

**Expected behavior**
The builder should have a consistent view of the chain. If its nonce increases, it should be able to see the status of the transaction that was just mined. 

**Logs**
(added some extra logs in a debug build)
```
  2023-08-13T19:08:10.017849Z  INFO rundler::builder::transaction_tracker: Nonce changed from 132 to 133. Pending transactions: [PendingTransaction { tx_hash: 0x465a215b1ef82c28aefebd4935a3693163766d2585e939212b9bcbb0c5cfd237, gas_fees: GasFees { max_fee_per_gas: 100000000, max_priority_fee_per_gas: 0 }, attempt_number: 0 }]
    at src/builder/transaction_tracker.rs:261

  2023-08-13T19:08:10.141960Z  INFO rundler::builder::sender::conditional: tx_hash 0x465a215b1ef82c28aefebd4935a3693163766d2585e939212b9bcbb0c5cfd237 tx: None
    at src/builder/sender/conditional.rs:61

  2023-08-13T19:08:10.142100Z  INFO rundler::builder::server: Nonce used by external transaction
    at src/builder/server.rs:212

  2023-08-13T19:08:10.142281Z  INFO rundler::common::emit: Transaction failed because nonce was used by another transaction outside of this Rundler.    Nonce: 132    Entrypoint: 0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789
```

dancoombs:
Fixes #295 
Fixes #296

## Proposed Changes

  - Remove stand-alone conditional sender. Replace with a CLI flag on the raw sender. They were almost exactly copy/paste.
  - Add new CLI argument to control which URL is used for transaction status: node or submit.
  - Add dropped status back to transaction tracker
  - Add new CLI argument to control if the raw sender supports the dropped status or not.


0xfourzerofour (approved):
LGTM

bin/rundler/src/cli/builder.rs
```diff
@@ -115,14 +115,48 @@ pub struct BuilderArgs {
...
+    /// Only used when BUILDER_SENDER is "raw"
     #[arg(
         long = "builder.submit_url",
         name = "builder.submit_url",
         env = "BUILDER_SUBMIT_URL"
     )]
     pub submit_url: Option<String>,
 
+    /// If present, the url of the ETH provider that will be used to check
+    /// transaction status. Else will use a the node http for status.
```
dphilipson, line 127: Typo: "will use a the"

dphilipson (approved):
Looks great!

PR #723, opened 2024-06-13
feat(builder): reject ops if condition not met after failure

### LINKED ISSUE #678: [builder] If conditional sending fails, remove UOs from mempool
**Describe the bug**
Currently if a conditional send fails the builder will loop trying to send the same UO, which will always fail since its already captured the bad state. The offending UO should be removed from the pool.

Arbitrum currently doesn't return the actual address/slot that caused the issue, so it may be difficult to determine which UO caused the issue. The option may be to just remove every UO from the bundle.

**To reproduce**
* Intentionally cause Rundler simulation to capture bad conditional values. This requires a code change.
* Send a UO and watch bundling logs.

**Expected behavior**
* The offending UO should be removed from the pool.

**Logs**
```
  2024-04-12T13:47:28.121490Z ERROR rundler_builder::bundle_sender: Failed to send bundle. Will retry next block: Other(
    JsonRpcClientError(
        JsonRpcError(
            JsonRpcError {
                code: -32000,
                message: "Storage slot value condition not met",
                data: None,
            },
        ),
    ),
)
```

will repeat.

dancoombs:
Closes #678 

## Proposed Changes

  - Currently, if a UO fails a conditional check it says in the mempool and can cause further conditional checks to fail, catching the bundler in a failing loop.
  - This PR detects the conditional check failure, and starts a process to remove the failing UO from the pool.
  - Because the failure does not return the slot that failed, the approach here is to notify the proposer that we are in this state. In the next bundle iteration the proposer will check all of the expected slots in a bundle and reject any UOs that fail conditions.
  - Currently issues an `eth_call` per UO per accessed address. If this becomes an issue we can use a batched `eth_call` approach.

## TODO

- [ ] Unit tests


crates/builder/src/bundle_proposer.rs
```diff
@@ -549,6 +568,78 @@ where
...
+                    .await
+                    .map(|reason| (i, reason))
+            })
+            .collect::<Vec<_>>();
+
+        let to_reject = future::join_all(futs)
+            .await
+            .into_iter()
+            .flatten()
+            .collect::<Vec<_>>();
```
dphilipson, line 586: (Nit) There is no need to collect here. Can leave out this line.

crates/builder/src/bundle_proposer.rs
```diff
@@ -549,6 +568,78 @@ where
...
+                            address: *address,
+                            slot: *slot,
+                            expected: *expected,
+                            actual,
+                        });
+                    }
+                }
+                None
+            })
+            .collect::<Vec<_>>();
```
dphilipson, line 632: (Nit) Also don't think this collect is needed.

crates/provider/src/ethers/provider.rs
```diff
@@ -207,6 +208,47 @@ impl<C: JsonRpcClient + 'static> Provider for EthersProvider<C> {
             .await
             .context("should get gas used")?)
     }
+
+    async fn batch_get_storage_at(
+        self: &Arc<Self>,
```
dphilipson, line 213: It looks like `&Self` is fine, no need for the `Arc`.

dphilipson (approved):
Looks great!

PR #725, opened 2024-06-14
feat(builder): allow for multiple private keys

### LINKED ISSUE #724: [builder] Support multiple private keys
**Describe the feature**
Currently the builder task only supports a single private key input. We should update this to support multiple private keys and add a CLI flag to pass those private keys in.

This is increasingly important as we support both entry points by default, which requires two keys. Users should be able to use both entry points with private keys and thus we need to modify the builder to allow for multiple private keys at input.

dancoombs:
Closes #724 

## Proposed Changes

  - Add `builder.private_keys` CLI flag. Deprecate `builder.private_key`, combine if both used.
  - Consume multiple private keys when creating bundle builders

### DIFF
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -13,7 +13,7 @@
 
 use std::{net::SocketAddr, time::Duration};
 
-use anyhow::Context;
+use anyhow::{bail, Context};
 use clap::Args;
 use rundler_builder::{
     self, BloxrouteSenderArgs, BuilderEvent, BuilderEventKind, BuilderTask, BuilderTaskArgs,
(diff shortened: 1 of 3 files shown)


bin/rundler/src/cli/builder.rs
```diff
@@ -284,10 +299,24 @@ impl BuilderArgs {
             num_builders += common.num_builders_v0_7;
         }
 
-        if self.private_key.is_some() {
-            if num_builders > 1 {
+        if (self.private_key.is_some() || !self.private_keys.is_empty())
```
0xfourzerofour, line 302: wouldn't this first check for the private key invalidate the if statement below on line 311?

0xfourzerofour: actually dont mind me

0xfourzerofour (approved):

bin/rundler/src/cli/builder.rs
```diff
@@ -57,13 +57,28 @@ pub struct BuilderArgs {
...
     #[arg(
         long = "builder.private_key",
         name = "builder.private_key",
         env = "BUILDER_PRIVATE_KEY"
     )]
     private_key: Option<String>,
 
+    /// Private keys to use for signing transactions
+    ///
+    /// Cannot use both `builder.private_key` and `builder.aw_kms_key_ids` at the same time.
```
dphilipson, line 73: Typo: "aw_kms_key_ids" -> "aws_kms_key_ids"

bin/rundler/src/cli/builder.rs
```diff
@@ -284,10 +299,24 @@ impl BuilderArgs {
             num_builders += common.num_builders_v0_7;
         }
 
-        if self.private_key.is_some() {
-            if num_builders > 1 {
+        if (self.private_key.is_some() || !self.private_keys.is_empty())
+            && !self.aws_kms_key_ids.is_empty()
+        {
+            return Err(anyhow::anyhow!(
```
dphilipson, line 305: I know this was existing code, but note that these `return Err` expressions can be written more concisely as
```rust
bail!("msg");
```
likewise below.

dphilipson (approved):
Nice work with this!

PR #726, opened 2024-06-14
feat(pool): allow staked senders to have multiple UOs in best_operations

### LINKED ISSUE #251: pool/builder: Allow a staked account to have multiple UOs in a single bundle
**Describe the feature**
Power users may want the ability to send multiple, parallel user operations where they don't care about the ordering. To do this, they can use a separate nonce key per parallel stream of operations, and send each to the mempool. 

If a sender is unstaked, the pool currently limits to 4 operations in the mempool. The pool also will only send a single operation per sender when asked for `best_operations`. 

For a staked sender, this limitation is removed by the spec. The user can have an larger amount of operations in the mempool (limited by reputation rules), and is allowed to have more than 1 user operation in a bundle.

Currently, our builder and pool limit even staked senders to 1 user operation in a bundle. We should remove this limitation.

dancoombs:
Closes #251 

## Proposed Changes

  - Simple change to allow staked accounts to have multiple UOs when calling `best_operations` on the pool. No changes were needed in the builder to allow for multiple UOs from the same sender in a bundle.
  - Tested this by staking a sender on Arb Sepolia and sending 10 UOs (bundler in manual mode to force them all into a single bundle). Transaction [here](https://sepolia.arbiscan.io/tx/0x9f9ddbd5cea92829148f67c3d051c53483bb1cfc3b1145ef15c52502e9cbef73#eventlog).


0xfourzerofour (approved):

dancoombs:
Closing in favor of https://github.com/alchemyplatform/rundler/pull/732

PR #728, opened 2024-06-14
fix(pool): fix race condition in paymaster tracking

dancoombs:
## Proposed Changes

  - Fix race condition

Race

1. Paymaster sends deposit transaction
2. UO uses paymaster, paymaster is new to the pool, queries chain for balance and adds to cache, sees the balance from 1
3. Deposit event from (1) is notified to the pool, balance is added to the balance from (2)

Since there is no ordering guarantee between (2) and (3) you can end up with either balance or 2x balance in the cache.
Instead process events as a reset event, and just query the chain for a balance after an event.

### DIFF
--- a/crates/pool/src/mempool/paymaster.rs
+++ b/crates/pool/src/mempool/paymaster.rs
@@ -26,7 +26,7 @@ use rundler_types::{
 use rundler_utils::cache::LruMap;
 
 use super::MempoolResult;
-use crate::chain::{BalanceUpdate, MinedOp};
+use crate::chain::MinedOp;
 
 /// Keeps track of current and pending paymaster balances
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


crates/pool/src/mempool/paymaster.rs
```diff
@@ -174,7 +152,20 @@ where
         self.state.write().set_tracking(tracking_enabled);
     }
 
-    pub(crate) async fn reset_confimed_balances(&self) -> MempoolResult<()> {
+    pub(crate) async fn reset_confirmed_balances_for(
+        &self,
+        addresses: Vec<Address>,
```
dphilipson, line 157: (Nit) Having this be a `&[Address]` would be more idiomatic.

dphilipson (approved):
Great catch, thanks for fixing!

PR #729, opened 2024-06-17
feat(pool): add time to mine tracking and metrics

### LINKED ISSUE #319: builder: Metric to measure transaction time to mine
**Describe the feature**
It would be super helpful to have a metric that tracks how long it takes for our transactions to be mined after sending them to block builders.

dancoombs:
Closes #319

## Proposed Changes

  - Add time to mine context to user operations in the pool
  - Track the number of user operations that are currently "candidates" i.e. their fees are high enough to be included
  - Track the amount of time/blocks an operation spends in the "candidate" state
  - Add histogram for time/blocks to mine
  - Re-use the expiry loop into a maintenance loop to limit O(n) operations
  - Limitation: time to mine has a granularity of 1(s) so on fast block networks (Arbitrum) we don't have the level of granularity needed, however we do have N blocks.
  - Open question: should we switch candidate time to be based on system time and not block time?

### DIFF
--- a/crates/pool/src/mempool/pool.rs
+++ b/crates/pool/src/mempool/pool.rs
@@ -12,9 +12,10 @@
 // If not, see https://www.gnu.org/licenses/.
 
 use std::{
-    cmp::Ordering,
+    cmp::{self, Ordering},
     collections::{hash_map::Entry, BTreeSet, HashMap, HashSet},
     sync::Arc,
+    time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 use anyhow::Context;
@@ -24,7 +25,7 @@ use ethers::{
 };
 use rundler_types::{
     pool::{MempoolError, PoolOperation},
-    Entity, EntityType, Timestamp, UserOperation, UserOperationId, UserOperationVariant,
+    Entity, EntityType, GasFees, Timestamp, UserOperation, UserOperationId, UserOperationVariant,
 };
 use rundler_utils::math;
 use tracing::info;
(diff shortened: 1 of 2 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -145,22 +151,50 @@ impl PoolInner {
...
-    pub(crate) fn remove_expired(&mut self, expire_before: Timestamp) -> Vec<(H256, Timestamp)> {
+    pub(crate) fn do_maintenance(
+        &mut self,
+        block_number: u64,
+        block_time: Timestamp,
+        candidate_gas_fees: GasFees,
+        base_fee: U256,
+    ) -> Vec<(H256, Timestamp)> {
+        let block_delta_time = block_time - self.prev_block_time;
+        let block_delta_height = block_number - self.prev_block_number;
```
dphilipson, line 169: This seems reasonable, but wanted to make sure that you or other reviewers are aware of this: in the past, block timestamps have not been a reliable way to know when a block was mined. Especially during the proof-of-work days of ethereum, the timestamp on a block was typically the time that the _previous_ block was mined. My theory is that as soon as a block is mined, miners would construct a candidate block using that moment in time as the timestamp, and not update the timestamp as they did their hashing for proof-of-work.

I don't think this is as much of an issue anymore, but I haven't looked into it closely and don't know how much this behavior varies across chains. I do think the code as written is probably good enough.

dphilipson (approved):
Looks great!

crates/pool/src/mempool/pool.rs
dancoombs: Thanks for pointing this out. This coupled with the problem that these are seconds based and some chains have faster than millisecond block times has convinced me to just make this based on the system clock.

PR #730, opened 2024-06-18
feat(tracer): add timeout of 10s to tracer

0xfourzerofour:
## Proposed Changes

  - Change timeout of debug_traceCall for custom tracers to 10s instead of the default 5s
  - Add CLI flag to make it customizable 
  
  DOCS: 
  https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug
  https://pkg.go.dev/time#ParseDuration

### DIFF
--- a/bin/rundler/Cargo.toml
+++ b/bin/rundler/Cargo.toml
@@ -27,6 +27,7 @@ dotenv = "0.15.0"
 ethers.workspace = true
 itertools = "0.12.1"
 metrics = "0.22.1"
+go-parse-duration = "0.1"
 metrics-exporter-prometheus = { version = "0.13.1", default-features = false, features = ["http-listener"] }
 metrics-process = "1.2.1"
 metrics-util = "0.16.2"
--- a/bin/rundler/src/cli/mod.rs
+++ b/bin/rundler/src/cli/mod.rs
@@ -11,7 +11,7 @@
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
 
-use anyhow::Context;
+use anyhow::{bail, Context};
 use clap::{builder::PossibleValuesParser, Args, Parser, Subcommand};
 
 mod builder;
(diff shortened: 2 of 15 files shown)


crates/sim/src/simulation/v0_6/tracer.rs
```diff
@@ -68,6 +69,7 @@ where
         op: UserOperation,
         block_id: BlockId,
         max_validation_gas: u64,
+        tracer_timeout: String,
```
dancoombs, line 72: Since we're modifying this code, would it make sense to pass both `max_validation_gas` and `tracer_timeout` as part of the `SimulateValidationTracerImpl` constructor and remove it from this interface?

bin/rundler/src/cli/mod.rs
```diff
@@ -163,6 +163,15 @@ pub struct CommonArgs {
     )]
     min_unstake_delay: u32,
 
+    #[arg(
```
dancoombs, line 166: Can you add a comment containing the https://pkg.go.dev/time#ParseDuration link?

Also can you add this to the CLI docs with both the links from the PR description?

bin/rundler/src/cli/mod.rs
```diff
@@ -364,6 +373,7 @@ impl From<&CommonArgs> for SimulationSettings {
             value.min_stake_value,
             value.max_simulate_handle_ops_gas,
             value.max_verification_gas,
+            value.tracer_timeout.clone(),
```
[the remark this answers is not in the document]
0xfourzerofour: should be reasonably easy to do to make sure it fits a certain format

crates/sim/src/simulation/v0_6/tracer.rs
0xfourzerofour: yeh that may be a little better to remove extra variables

crates/sim/src/simulation/mod.rs
```diff
@@ -162,6 +162,8 @@ pub struct Settings {
     pub max_simulate_handle_ops_gas: u64,
     /// The maximum amount of verification gas that can be used during the simulation call
     pub max_verification_gas: u64,
+    /// The max duration of the custom javascript tracer
```
dancoombs, line 165: Lets be clear here what the `String` should look like, given that this is the public interface doc.

bin/rundler/Cargo.toml
```diff
@@ -27,6 +27,7 @@ dotenv = "0.15.0"
 ethers.workspace = true
 itertools = "0.12.1"
 metrics = "0.22.1"
+go-parse-duration = "0.1"
```
dancoombs, line 30: Lol I can't believe there is a crate for this

dancoombs (approved):
LGTM 🚀

bin/rundler/Cargo.toml
0xfourzerofour: yeh I was very surprised as well lol

PR #738, opened 2024-06-24
fix(builder): fix builder state machine replacement tracking

dancoombs:
Implement a better tracking mechanism for state machine replacement underpriced tracking.

TODO:

- [x] Unit tests for new tracking
- [x] Update docs with state machine diagram and description

### DIFF
--- a/crates/builder/src/task.rs
+++ b/crates/builder/src/task.rs
@@ -91,8 +91,10 @@ pub struct Args {
     pub max_blocks_to_wait_for_mine: u64,
     /// Percentage to increase the fees by when replacing a bundle transaction
     pub replacement_fee_percent_increase: u64,
-    /// Maximum number of times to increase the fees when replacing a bundle transaction
-    pub max_fee_increases: u64,
+    /// Maximum number of times to increase the fee when cancelling a transaction
+    pub max_cancellation_fee_increases: u64,
+    /// Maximum amount of blocks to spend in a replacement underpriced state before moving to cancel
+    pub max_replacement_underpriced_blocks: u64,
     /// Address to bind the remote builder server to, if any. If none, no server is starter.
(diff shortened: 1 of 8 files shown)


crates/builder/src/task.rs
```diff
@@ -91,8 +91,10 @@ pub struct Args {
     pub max_blocks_to_wait_for_mine: u64,
     /// Percentage to increase the fees by when replacing a bundle transaction
     pub replacement_fee_percent_increase: u64,
-    /// Maximum number of times to increase the fees when replacing a bundle transaction
-    pub max_fee_increases: u64,
+    ///
```
0xfourzerofour, line 94: should there be comments here?

crates/builder/src/transaction_tracker.rs
```diff
@@ -399,6 +433,12 @@ where
         let nonce = self.get_external_nonce().await.unwrap_or(self.nonce);
         self.set_nonce_and_clear_state(nonce);
     }
+
+    fn abandon(&mut self) {
+        self.has_abandoned = true;
```
0xfourzerofour, line 438: Nit: has_abandoned sounds like something in the op is abandoned not the op itself, I think it should be called was_abandoned or is_abandoned

0xfourzerofour (approved):
looks good just a nit

crates/builder/src/task.rs
dancoombs: Yup, done

crates/builder/src/transaction_tracker.rs
dancoombs: We already have `has_dropped` on this class, so I kept the same naming

0xfourzerofour (approved):

Issue #495, opened 2023-11-13
Find and fix increasing memory data scructure


0xfourzerofour:
**Describe the bug**
When running rundler for a longer period of time. There seems to be a steady increase of data being stored that can be pruned. 

Find this increase in data and prune it when it can be removed

PR #748, opened 2024-06-28
fix: various v0.3-rc fixes

dancoombs:
## Proposed Changes

  - Modify base chain config for a new block congestion threshold and min priority fee based on recent chain usage
  - Hardcode a gas limit for cancellation transactions
  - Increase the timeout for KMS connections - was triggering this in local testing
  - Move a few info logs to debug
  - Mix missed clear in mempool
  - Noticed an existing bug in the mempool with removing replaced transactions. Potentially https://github.com/alchemyplatform/rundler/issues/495, will fix in follow up PR.

### DIFF
--- a/crates/builder/src/task.rs
+++ b/crates/builder/src/task.rs
@@ -407,11 +407,11 @@ where
         } else {
             info!("Using AWS KMS signer");
             let signer = time::timeout(
-                // timeout must be << than the lock TTL to avoid a
+                // timeout must be < than the lock TTL to avoid a
                 // bug in the redis lock implementation that panics if connection
                 // takes longer than the TTL. Generally the TLL should be on the order of 10s of seconds
                 // so this should give ample time for the connection to establish.
-                Duration::from_millis(self.args.redis_lock_ttl_millis / 10),
+                Duration::from_millis(self.args.redis_lock_ttl_millis / 4),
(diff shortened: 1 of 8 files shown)


crates/builder/src/task.rs
```diff
@@ -407,11 +407,11 @@ where
         } else {
             info!("Using AWS KMS signer");
             let signer = time::timeout(
-                // timeout must be << than the lock TTL to avoid a
+                // timeout must be < than the lock TTL to avoid a
                 // bug in the redis lock implementation that panics if connection
                 // takes longer than the TTL. Generally the TLL should be on the order of 10s of seconds
                 // so this should give ample time for the connection to establish.
-                Duration::from_millis(self.args.redis_lock_ttl_millis / 10),
+                Duration::from_millis(self.args.redis_lock_ttl_millis / 4),
```
0xfourzerofour, line 414: should we make this denominator configurable?

0xfourzerofour (approved):

crates/builder/src/task.rs
dancoombs: I don't see a need to complicate, this should be fine

dancoombs: Really want to rip all this code out anyway with a remote signer service

PR #751, opened 2024-07-02
fix(cli): error if recursive base config, remove goerli netowrks

dancoombs:
## Proposed Changes

  - Error if the base chain spec has a base key
  - Remove goerli hardcoded networks


0xfourzerofour (approved):
Good catch

PR #752, opened 2024-07-02
feat(sim): ban access to arbitrum stylus contracts during sim

dancoombs:
## Proposed Changes

  - Add a new simulation error for unsupported contract type
  - Capture contract "header" (first 3 bytes) for each call during tracing
  - If header matches the Arbitrum stylus header `0xEFF000`, add a violation


0xfourzerofour (approved):
This looks good, can you add the documentation link to the readme or code that states the header must be that value?

dancoombs:
> This looks good, can you add the documentation link to the readme or code that states the header must be that value?

I'm not aware of anything public, this value came directly from a DM with the arb team

PR #753, opened 2024-07-02
fix(builder): improve replacement underpriced handling

dancoombs:
## Proposed Changes

  - Remove dropped status from transaction tracker. No need to track them as we should still increase fees when transactions are dropped, and there are UOs that are willing to pay.
  - After increasing fees, if no UOs are willing to pay, then abandon and wait for next trigger
  - If replacement underpriced, unabandon to ensure that any transactions are included in the next fee estimation.

The "abandoned" state is used to "retry" after we've reached a point where we can't submit any more UOs. When we get replacement underpriced in this state we'll loop for a few rounds until moving to cancellations. Under normal network conditions, this should only happen if our gas fee estimation is significantly off and transactions are not landing.


0xfourzerofour (approved):
LGTM

PR #769, opened 2024-08-28
chore: update rust version and various deps

dancoombs:
## Proposed Changes

  - Update Rust to v1.80.0
  - Update dependencies without breaking changes
 
 Still need to update:
 
 - `jsonrpsee` is 4 versions behind. The migration looks complex, will attempt in a follow up.
 - `tower` - `jsonrpsee` has a dependency here, will do at the same time as `jsonrpsee`.
 - `reqwest` - `ethers` has a dependency here, will do as part of the `alloy` migration.


0xfourzerofour (approved):
LGTM

crates/types/src/timestamp.rs
```diff
@@ -266,7 +266,7 @@ mod test {
     #[test]
     fn test_out_of_bounds_display() {
         let actual = get_timestamp_out_of_bounds_for_datetime().to_string();
-        assert_eq!(actual, "later than +262143-12-31 23:59:59.999999999 UTC");
+        assert_eq!(actual, "later than +262142-12-31 23:59:59.999999999 UTC");
```
dancoombs, line 269: https://github.com/chronotope/chrono/releases/tag/v0.4.32

andysim3d (approved):

PR #771, opened 2024-08-29
feat(estimation): increase max gas for simulation

0xfourzerofour:
## Proposed Changes

  - Increase the max gas for simulateHandleOps to the alchemy level max as no gas. If the value is too low on some networks we will be unable to estimate properly


dancoombs (changes requested):
Lets leave this as is and just modify the settings in our deployments

PR #772, opened 2024-08-29
fix: add bucket to histogram metrics

andysim3d:
[Closes/Fixes] #

rundler's histgram metrics are incomplete, they miss the bucket metrics. We can only get quantile info without weights. this pr is to add buckets info. the bucket is defined as 2^x, balancing metric density and data accuracy. 

## Proposed Changes

  -
  -
  -

### DIFF
--- a/bin/rundler/src/cli/metrics.rs
+++ b/bin/rundler/src/cli/metrics.rs
@@ -23,6 +23,7 @@ pub fn initialize<'a>(
     sample_interval_millis: u64,
     listen_addr: SocketAddr,
     tags: impl IntoIterator<Item = &'a String>,
+    buckets: &[f64],
 ) -> anyhow::Result<()> {
     let mut builder = PrometheusBuilder::new().with_http_listener(listen_addr);
 
@@ -33,6 +34,8 @@ pub fn initialize<'a>(
     for (k, v) in tags {
         builder = builder.add_global_label(k, v);
     }
+
+    builder = builder.set_buckets(buckets)?;
 
     let (recorder, exporter) = builder.build()?;
     tokio::spawn(exporter);
--- a/bin/rundler/src/cli/mod.rs
+++ b/bin/rundler/src/cli/mod.rs
@@ -46,6 +46,7 @@ pub async fn run() -> anyhow::Result<()> {
         opt.metrics.sample_interval_millis,
         metrics_addr,
         &opt.metrics.tags,
+        &opt.metrics.buckets,
     )
     .context("metrics server should start")?;
 
@@ -443,6 +444,18 @@ pub struct MetricsArgs {
         global = true
     )]
     sample_interval_millis: u64,
+
+    /// Bucket for histogram metric.
+    #[arg(
+        long = "metrics.histogram_buckets", 
+        name = "metrics.histogram_buckets",
+        env = "METRICS_HISTOGRAM_BUCKETS",
+        default_values_t = vec![0.0, 1000.0, 2000.0, 3000.0, 4000.0, 5000.0, 6000.0, 7000.0,
+                                8000.0, 9000.0, 10000.0, 11000.0, 12000.0, 13000.0, 14000.0,
+                                15000.0, 16000.0, 32000.0, 64000.0, 128000.0, 256000.0, 512000.0],
+        global = true
+    )]
+    buckets: Vec<f64>,
 }
 
 /// CLI options for logging


bin/rundler/src/cli/metrics.rs
```diff
@@ -33,6 +37,9 @@ pub fn initialize<'a>(
     for (k, v) in tags {
         builder = builder.add_global_label(k, v);
     }
+    let buckets = generate_exponential_series(0.01, 2.0, 32);
```
dancoombs, line 40: Just ran this to see what it would generate and got:

```
[0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12, 10.24, 20.48, 40.96, 81.92, 163.84, 327.68, 655.36, 1310.72, 2621.44, 5242.88, 10485.76, 20971.52, 41943.04, 83886.08]
```

How do these bucket values impact the resulting histogram?

This might not be giving us the granularity we need in the 1-20s range to get an accurate measurement here. Specifically for the time to mine metric. We likely need different range to measure RPC method histograms.

andysim3d: > How do these bucket values impact the resulting histogram

values in the same bucket are considered even distribution. for instance, if P90 is 5, P95 is 10, and a bucket[5. 10], then P92 will considered as 7. 


normally there are 2 type of histogram bucket, linear (1, 2, 3, 4, ...) and exponential (1, 2, 4, 8, ...). they are mostly chosen by the metric type: for even distribution, linear is better and for normal distribution, exponential is preferred. 

I think the concern here is about the step(because the time is mostly normal distribution), wdyt the value change to following: 
```1, 1.5, 2.25, 3.375, 5.0625, 7.59375, 11.390625, 17.0859375, 25.62890625, 38.443359375, 57.6650390625, 86.49755859375, 129.746337890625, 194.6195068359375, 291.92926025390625, 437.8938903808594, 656.8408355712891, 985.2612533569336, 1477.8918800354004, 2216.8378200531006, 3325.256730079651, 4987.885095119476, 7481.8276426792145, 11222.741464018822, 16834.112196028233, 25251.16829404235, 37876.75244106352, 56815.128661595285, 85222.69299239293, 127834.03948858939, 191751.0592328841, 287626.58884932613```

Comment (dancoombs) [reply bin/rundler/src/cli/metrics.rs:40]: Is it weird to combine the 2? Could make sense to do a linear time bucketing up to ~16 seconds (one per second) to capture the normal cases and then exponential to capture the long tail.

Comment (andysim3d) [reply bin/rundler/src/cli/metrics.rs:40]: it is not weird but unusual to combine them. cause the estimation of the quantile is based on density of bucket. combining them *maybe* makes arbitrary quantile compute return weird value.

Comment (dancoombs) [reply bin/rundler/src/cli/metrics.rs:40]: Maybe we try it and see what it looks like? For experimentation purposes do you think we should move this to be a CLI parameter so we can modify as needed? We can just hardcode the buckets by hand as opposed to calculating them in code

Comment (andysim3d) [reply bin/rundler/src/cli/metrics.rs]: Done

File: bin/rundler/src/cli/metrics.rs [hunk 2]
```diff
@@ -23,6 +23,7 @@ pub fn initialize<'a>(
     sample_interval_millis: u64,
     listen_addr: SocketAddr,
     tags: impl IntoIterator<Item = &'a String>,
+    buckets: &Vec<f64>,
```
dancoombs, line 26: I think this can just be `&[f64]` and then below you won't need to do `as_slice()`

andysim3d: https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes 

it seems clap not support slice as default argument. I would assume `Vec<f64>` is better here

dancoombs: A `Vec<T>` in rust can deref to a `&[T]` by just doing `&v`. You can leave the CLI type as a vector and deref to a slice.

andysim3d: Done

dancoombs (approved):
LGTM, you'll need to fix the commit messages and I can give another check.

andysim3d:
> LGTM, you'll need to fix the commit messages and I can give another check.

commits squashed.

dancoombs (approved):

PR #775, opened 2024-08-30
feat: inject providers from bin

### LINKED ISSUE #416: refactor: inject providers from bin
**Describe the feature**
We currently create ethers providers within task implementations. Instead we should inject these providers from bin.

dancoombs:
Closes #416 

## Proposed Changes

  - Modify tasks to take providers as arguments.
  - Construct providers in bin in a single spot.
  - This preps us for moving to Alloy providers, as there are only a few places where we will need to update. Providers, signers, and bin being the major places.

### DIFF
--- a/crates/rpc/src/task.rs
+++ b/crates/rpc/src/task.rs
@@ -13,14 +13,13 @@
 
 use std::{net::SocketAddr, sync::Arc, time::Duration};
 
-use anyhow::bail;
+use anyhow::{bail, Context};
 use async_trait::async_trait;
-use ethers::providers::{JsonRpcClient, Provider};
 use jsonrpsee::{
     server::{middleware::http::ProxyGetRequestLayer, RpcServiceBuilder, ServerBuilder},
     RpcModule,
 };
-use rundler_provider::{EthersEntryPointV0_6, EthersEntryPointV0_7};
+use rundler_provider::{EntryPointProvider, Provider};
 use rundler_sim::{
     EstimationSettings, FeeEstimator, GasEstimatorV0_6, GasEstimatorV0_7, PrecheckSettings,
 };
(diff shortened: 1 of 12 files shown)


andysim3d (approved):
LGTM!

crates/rpc/src/task.rs
```diff
@@ -80,48 +82,45 @@ pub struct Args {
...
     pool: P,
     builder: B,
+    provider: Arc<PR>,
+    ep_06: Option<E06>,
+    ep_07: Option<E07>,
 }
 
 #[async_trait]
-impl<P, B> Task for RpcTask<P, B>
+impl<P, B, PR, E06, E07> Task for RpcTask<P, B, PR, E06, E07>
```
0xfourzerofour, line 95: I wonder if there is a good way here to not have an Entrypoint provider generic for each new entrypoint and instead do coalescing at runtime by passing in an array instead of multiple keys within the struct for each entrypoint.

This is fine for now but I think it should be rethought instead of stacking generics as it makes the code a little wild

0xfourzerofour: sorta like this

```rust

pub struct RpcTask<P, B, PR, EP> {
    args: Args,
    pool: P,
    builder: B,
    provider: Arc<PR>,
    entry_points: Vec<EP>,
}

#[async_trait]
impl<P, B, PR, EP> Task for RpcTask<P, B, PR, EP>
where
    P: Pool + HealthCheck + Clone,
    B: Builder + HealthCheck + Clone,
    PR: Provider,
    EP: EntryPointProvider<UserOperation>,
{
    async fn run(mut self: Box<Self>, shutdown_token: CancellationToken) -> anyhow::Result<()> {
        for ep in &self.entry_points {
        
            /// do the rest of the stuff per entrypoiny
        
        
        }

```

0xfourzerofour (approved):
logic Looks fine but just general code changes to remove an additional generic would be nice

PR #776, opened 2024-09-03
feat(utils): update utils to use alloy

dancoombs:
## Proposed Changes

  - Update `utils` to use alloy


crates/utils/src/eth.rs
```diff
@@ -11,59 +11,9 @@
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
 
-//! Utilities for working with an Ethereum-like chain via Ethers.
+//! Utilities for working with an EVM chains.
 
-use ethers::{
```
dancoombs, line 16: Will potentially add back new versions of these as I work on the downstream crates. However the new `sol!` macro may make it possible to replace most of these.

andysim3d (approved):

PR #780, opened 2024-09-04
feat(types): update types to use alloy

dancoombs:
## Proposed Changes

  - Downgrade alloy due to https://github.com/alloy-rs/core/issues/717
  - Remove contracts from types crate
  - Update types crate to use alloy and new rundler-contracts crate


crates/contracts/src/v0_6.rs
```diff
@@ -40,10 +40,44 @@ sol! {
         bytes signature;
     }
 
+    #[allow(missing_docs)]
+    #[derive(Default, Debug, PartialEq, Eq)]
+    struct ReturnInfo {
+        uint256 preOpGas;
```
0xfourzerofour, line 46: is this valid rust?

0xfourzerofour: ok its wrapped in sol soz

crates/types/src/user_operation/v0_6.rs
```diff
@@ -11,25 +11,22 @@
...
-pub use crate::contracts::v0_6::i_entry_point::{UserOperation, UserOpsPerAggregator};
 use crate::{
     chain::ChainSpec,
     entity::{Entity, EntityType},
     EntryPointVersion,
 };
 
 /// Gas overhead required by the entry point contract for the inner call
-pub const ENTRY_POINT_INNER_GAS_OVERHEAD: U256 = U256([5_000, 0, 0, 0]);
+pub const ENTRY_POINT_INNER_GAS_OVERHEAD: U256 = uint!(5000_U256);
```
0xfourzerofour, line 29: this macro is so much better

crates/contracts/src/v0_6.rs
dancoombs: Yeah tbh this is really weird to see, Rust proc macros on solidity code. Powerful though

PR #790, opened 2024-09-11
feat(types): modify types to use native u128 where possible

dancoombs:
## Proposed Changes

  - Modify types to use native u128 instead of ruint U128 for local (not onchain) versions of user operations.
  - Impacts gas limit and fee fields
  - Convert to U128/U256 when necessary
  - For EP v0.7 the only fallible conversion is for `preVerificationGas` which onchain can be a U256. It is assumed there are no use cases where PVG needs to be > u128. We we're already relying on a U256 -> u128 conversion for gas limit calculation anyway.
  - For EP v0.6 all the gas and fee fields can have a fallible conversion. It is also assumed there are no legitimate use cases where these values can be larger than u128. For gas/fee representations this would be larger than a block's gas limit or larger than the total amount of token supply for ETH.

### DIFF
--- a/crates/provider/src/alloy/entry_point/v0_6.rs
+++ b/crates/provider/src/alloy/entry_point/v0_6.rs
@@ -314,10 +314,7 @@ where
         user_op: UserOperation,
         max_validation_gas: u128,
     ) -> ProviderResult<(TransactionRequest, StateOverride)> {
-        let pvg: u128 = user_op
-            .pre_verification_gas
-            .uint_try_to()
-            .context("pre verification gas out of bounds")?;
+        let pvg: u128 = user_op.pre_verification_gas;
         let call = self
             .i_entry_point
             .simulateValidation(user_op.into())
@@ -332,10 +329,7 @@ where
         max_validation_gas: u128,
         block_id: Option<BlockId>,
     ) -> ProviderResult<Result<ValidationOutput, ValidationRevert>> {
-        let pvg: u128 = user_op
-            .pre_verification_gas
-            .uint_try_to()
-            .context("pre verification gas out of bounds")?;
+        let pvg: u128 = user_op.pre_verification_gas;
         let blockless = self
             .i_entry_point
             .simulateValidation(user_op.into())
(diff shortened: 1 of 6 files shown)


0xfourzerofour:
looks good but we need to fix the unit tests and linting

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -314,10 +314,7 @@ where
         user_op: UserOperation,
         max_validation_gas: u128,
     ) -> ProviderResult<(TransactionRequest, StateOverride)> {
-        let pvg: u128 = user_op
-            .pre_verification_gas
-            .uint_try_to()
-            .context("pre verification gas out of bounds")?;
+        let pvg: u128 = user_op.pre_verification_gas;
```
0xfourzerofour, line 317: nit: we do not need the explicit type definition if its defined at the UO struct

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -332,10 +329,7 @@ where
         max_validation_gas: u128,
         block_id: Option<BlockId>,
     ) -> ProviderResult<Result<ValidationOutput, ValidationRevert>> {
-        let pvg: u128 = user_op
-            .pre_verification_gas
-            .uint_try_to()
-            .context("pre verification gas out of bounds")?;
+        let pvg: u128 = user_op.pre_verification_gas;
```
0xfourzerofour, line 332: same here

crates/types/src/user_operation/v0_7.rs
```diff
@@ -133,13 +135,13 @@ impl UserOperationTrait for UserOperation {
         &self.call_data
     }
 
-    fn max_gas_cost(&self) -> U256 {
-        U256::from(self.max_fee_per_gas)
+    fn max_gas_cost(&self) -> u128 {
+        self.max_fee_per_gas
             * (self.pre_verification_gas
-                + U256::from(self.call_gas_limit)
```
0xfourzerofour, line 139: low chance of this happening but if someone inputs an arbitrarily large value for one of these fields we may have an overflow calling function so might be a good idea to saturate.

not sure if this is checked before this but worth the safety

0xfourzerofour (approved):
Looks good just a few nits

andysim3d (approved):

PR #792, opened 2024-09-12
feat(sim): update sim crate for new alloy types and provider interfaces

dancoombs:


crates/contracts/build.rs
```diff
@@ -62,6 +62,21 @@ fn generate_v0_7_bindings() -> Result<(), Box<dyn error::Error>> {
         bytecode,
     )?;
 
+    let json =
```
dancoombs, line 65: TODO DRY this

PR #795, opened 2024-09-13
feat(metrics): update histogram buckets and change TTM to seconds

0xfourzerofour:
## Proposed Changes
  - Add better bucket distribution for histograms 
  - Change TTM metrics to use seconds to fit within histogram distribution

### DIFF
--- a/crates/pool/src/mempool/pool.rs
+++ b/crates/pool/src/mempool/pool.rs
@@ -631,7 +631,7 @@ impl PoolMetrics {
             "op_pool_time_to_mine",
             "entry_point" => entry_point.to_string()
         )
-        .record(time_to_mine.candidate_for_time.as_secs_f64() as f64);
+        .record(time_to_mine.candidate_for_time.as_secs_f64());
         metrics::histogram!(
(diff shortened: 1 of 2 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -631,7 +631,7 @@ impl PoolMetrics {
             "op_pool_time_to_mine",
             "entry_point" => entry_point.to_string()
         )
-        .record(time_to_mine.candidate_for_time.as_millis() as f64);
+        .record(time_to_mine.candidate_for_time.as_secs_f64());
```
dancoombs, line 634: Do we still want this? The buckets now start at 50s, so basically everything will land in that bucket.

Pretty hard to find a bucket range that works for both these use cases...

0xfourzerofour: maybe we go exponsential for the first 100ms? so like 1, 2, 4, 8, 16, 32, 64, 100?

dancoombs (approved):

PR #797, opened 2024-09-16
fix failed spec test

andysim3d:
I can't tag release/0.7 in alchemyplatform/bundler-test-executor so use master instead. since alchemyplatform/bundler-test-executor is owned by alchemy so I think it is low risk for now.

this pr also add `debug_bundler_clearReputation` endpoint. it simply calls `debug_bundler_clearState`.

### DIFF
--- a/.github/workflows/compliance.yaml
+++ b/.github/workflows/compliance.yaml
@@ -28,7 +28,7 @@ jobs:
         with:
           path: bundler-test-executors
           repository: alchemyplatform/bundler-test-executor
-          ref: releases/v0.7
+          ref: master
 
       - name: Build rundler image locally
         run: docker buildx build ./rundler -t alchemyplatform/rundler:latest
--- a/.github/workflows/compliance.yaml
+++ b/.github/workflows/compliance.yaml
@@ -28,7 +28,7 @@ jobs:
         with:
           path: bundler-test-executors
           repository: alchemyplatform/bundler-test-executor
-          ref: releases/v0.6
+          ref: releases/v0.7
 
       - name: Build rundler image locally
         run: docker buildx build ./rundler -t alchemyplatform/rundler:latest
--- a/.github/workflows/compliance.yaml
+++ b/.github/workflows/compliance.yaml
@@ -29,6 +29,7 @@ jobs:
           path: bundler-test-executors
           repository: alchemyplatform/bundler-test-executor
           ref: master
+          submodules: recursive
 
       - name: Build rundler image locally
         run: docker buildx build ./rundler -t alchemyplatform/rundler:latest
--- a/.github/workflows/compliance.yaml
+++ b/.github/workflows/compliance.yaml
@@ -19,6 +19,9 @@ jobs:
           path: rundler
           submodules: recursive
 
+      - name: Install Foundry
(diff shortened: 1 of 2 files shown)


.github/workflows/compliance.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         with:
           path: bundler-test-executors
           repository: alchemyplatform/bundler-test-executor
-          ref: releases/v0.6
```
dancoombs, line 31: Lets not merge this, better to stay on `releases/v0.6` and figure out the permissions on the executor repo

crates/rpc/src/debug.rs
```diff
@@ -13,7 +13,10 @@
 
 use anyhow::Context;
 use async_trait::async_trait;
-use ethers::types::{Address, H256};
+use ethers::{
+    core::k256::elliptic_curve::consts::True,
```
dancoombs, line 17: Fix lints

crates/rpc/src/debug.rs
```diff
@@ -184,6 +190,11 @@ where
         )
         .await
     }
+
+    async fn bundler_clear_reputation(&self) -> RpcResult<String> {
+        utils::safe_call_rpc_handler("bundler_clearState", DebugApi::bundler_clear_state(self))
```
dancoombs, line 195: "debug_bundler_clearState" does more than just clear reputation. We should only clear the reputation portion.

dancoombs: The pool call takes 3 bools, we should only set `clearReputation` to true

``
      clearMempool: bool,   // optional, clears the UOs from the pool
      clearPaymaster: bool, // optional, clears the paymaster balances
      clearReputation: bool // optional, clears the reputation manager
 ```

dancoombs (changes requested):
This currently clears more state than it should. Lets make sure this only clears reputation. Also, lets follow up on the `releases/v0.6` permissions.

.github/workflows/compliance.yaml
dancoombs: Also, I'm remembering now that `master` is EP v0.7. `releases/v0.6` is the branch we upkeep for EP v0.6.

We should really be running both

crates/rpc/src/debug.rs
```diff
@@ -184,6 +188,14 @@ where
         )
         .await
     }
+
+    async fn bundler_clear_reputation(&self) -> RpcResult<String> {
+        utils::safe_call_rpc_handler(
+            "bundler_clearState",
+            DebugApi::bundler_clear_reputation(self),
```
dancoombs, line 195: Should this be "bundler_clearReputation"?

crates/rpc/src/debug.rs
```diff
@@ -184,6 +188,14 @@ where
         )
         .await
     }
+
+    async fn bundler_clear_reputation(&self) -> RpcResult<String> {
+        utils::safe_call_rpc_handler(
+            "bundler_clearState",
```
dancoombs, line 194: Should this be "bundler_clearReputation"

dancoombs (approved):
🚀

PR #799, opened 2024-09-17
feat(utils): update utils to use new types

dancoombs:


.gitmodules
```diff
@@ -1,21 +1,21 @@
-[submodule "crates/types/contracts/lib/account-abstraction-versions/v0_7"]
```
dancoombs, line 1: `.gitmodules`, `Cargo.lock`, and `Cargo.toml` require changes from later features, but just made the most sense to compile them all here. Can mostly ignore.

andysim3d (approved):

PR #801, opened 2024-09-18
feat(contracts): add contracts crate, remove from types

dancoombs:


crates/contracts/build.rs
```diff
@@ -0,0 +1,131 @@
...
+        .get("object")
+        .unwrap()
+        .as_str()
+        .unwrap();
+    fs::write(
+        "contracts/out/v0_7/EntryPointSimulations.sol/EntryPointSimulations_deployedBytecode.txt",
+        bytecode,
+    )?;
+
+    let json =
```
dancoombs, line 65: TODO: DRY

crates/contracts/src/v0_6.rs
```diff
@@ -0,0 +1,174 @@
...
+//
+// You should have received a copy of the GNU General Public License along with Rundler.
+// If not, see https://www.gnu.org/licenses/.
+
+// Contracts from https://github.com/eth-infinitism/account-abstraction/tree/releases/v0.6/contracts
+
+use alloy_primitives::Bytes;
+use alloy_sol_macro::sol;
+
+sol! {
```
dancoombs, line 19: The reason for using this macro and shoving all of the contract interfaces into the same macro is that alloy currently doesn't deduplicate shared types across invocations to `sol!`, and `sol!` only supports 1 file import.

Will be filing a feature request with them when I get around it.

Much better to deal with this ugliness here than to have 5 different `UserOperation` structs compiled from different contracts.

crates/contracts/src/v0_7.rs
```diff
@@ -0,0 +1,274 @@
...
+);
+
+sol!(
+    #[allow(missing_docs)]
+    #[sol(rpc)]
+    GetBalances,
+    "contracts/out/v0_7/GetBalances.sol/GetBalances.json"
+);
+
+// EntryPointSimulations deployed bytecode
```
dancoombs, line 246: We need `EntryPointSimulations` in the large `sol!` macro above for types sharing. However there is no good way to get the bytecode in that (other than pasting hex into the macro). This, combined with the build script, pulls the compiled bytecode from foundry into this file.

0xfourzerofour (approved):
LGTM

andysim3d (approved):

PR #802, opened 2024-09-18
feat(types): use alloy/native types in types crate

dancoombs:


crates/types/src/user_operation/v0_6.rs
```diff
@@ -44,8 +43,83 @@ pub const ENTRY_POINT_INNER_GAS_OVERHEAD: U256 = U256([5_000, 0, 0, 0]);
 /// paymaster_and_data = 32 bytes + 32 bytes num elems + var 32
 /// signature = 32 bytes + 32 bytes num elems + var 32
 ///
-/// 15 * 32 = 480
-const ABI_ENCODED_USER_OPERATION_FIXED_LEN: usize = 480;
+/// +1 for good luck? (aka this is what I got when I was testing)
```
0xfourzerofour, line 46: luck is on our side

crates/types/src/user_operation/v0_7.rs
```diff
@@ -233,9 +235,9 @@ impl UserOperationTrait for UserOperation {
 
     fn abi_encoded_size(&self) -> usize {
         ABI_ENCODED_USER_OPERATION_FIXED_LEN
-            + super::byte_array_abi_len(&self.packed.init_code)
-            + super::byte_array_abi_len(&self.packed.call_data)
-            + super::byte_array_abi_len(&self.packed.paymaster_and_data)
+            + super::byte_array_abi_len(&self.packed.initCode)
+            + super::byte_array_abi_len(&self.packed.callData)
```
0xfourzerofour, line 239: hate that the sol stuff defaults to camelcase

0xfourzerofour (approved):
LGTM 🚢

0xfourzerofour:
Looks like their might just be an issue with the build.rs stuff

dancoombs:
> Looks like their might just be an issue with the build.rs stuff

Hm, yeah it looks like Foundry isn't running for some reason in the contracts `build.rs`, I'll look into it but its unrelated to this PR.

crates/types/src/user_operation/v0_7.rs
dancoombs: Same. It actually worked out somewhat okay because we wrap a lot of the onchain types in our own types where we use snake_case. So its isolated to a few files that deal with the onchain types.

crates/types/src/user_operation/v0_6.rs
dancoombs: Yeah I couldn't figure this out, would love to look into it more but I trust the numbers coming out of alloy.

crates/types/src/user_operation/v0_6.rs
```diff
@@ -334,19 +432,15 @@ impl UserOperationOptionalGas {
...
-            self.verification_gas_limit,
-            max_verification_gas,
-            U256::zero(),
-        );
-        let pvg =
-            super::default_if_none_or_equal(self.pre_verification_gas, max_call_gas, U256::zero());
+        let cgl = super::default_if_none_or_equal(self.call_gas_limit, max_call_gas, 0);
+        let vgl =
+            super::default_if_none_or_equal(self.verification_gas_limit, max_verification_gas, 0);
+        let pvg = super::default_if_none_or_equal(self.pre_verification_gas, max_call_gas, 0);
```
andysim3d, line 443: just curious, I find those 3 variables only used to construct UO below, is that possible to inline them like L418 to L424?

crates/types/src/user_operation/v0_7.rs
```diff
@@ -425,28 +424,21 @@ impl UserOperationOptionalGas {
...
     ) -> UserOperationBuilder<'_> {
         // If unset or zero, default these to gas limits from settings
         // Cap their values to the gas limits from settings
-        let cgl = super::default_if_none_or_equal(self.call_gas_limit, max_call_gas, U128::zero());
-        let vgl = super::default_if_none_or_equal(
-            self.verification_gas_limit,
-            max_verification_gas,
-            U128::zero(),
-        );
+        let cgl = super::default_if_none_or_equal(self.call_gas_limit, max_call_gas, 0);
```
andysim3d, line 433: same here.

crates/types/src/user_operation/v0_6.rs
dancoombs: We could, but they're pulled out specifically so that the comment above clearly applies to these conversations

Issue #809, opened 2024-09-24
[pool] Evaluate using Alloy event watchers for chain tracking


dancoombs:
**Describe the feature**
Alloy has an event watching [feature](https://github.com/alloy-rs/alloy/blob/cf9a98fd7ed41f1e86af1d58d3eba9bc1c9799b0/crates/contract/src/event.rs#L70).

We can consider replacing our chain tracking logic with this logic.

A few things to look out for:

1. Need to be able to watch for multiple event types on multiple contract addresses at the same time. Don't want to duplicate polling.
2. Needs to work on ws/http. We've defaulted to http in the past due to ws reliability/latency issues. We may want to re-evalutate this.
3. Needs to work with or without using `eth_getFilterChanges` as we've also seen reliability/latency/support issues with RPC providers implementations of this method.

andysim3d:
Synced with @0xfourzerofour offline. the benefits of using the event watcher is we could remove the decoding code and simplified the topic handling logic [1]. But also we agreed the priority is not that high and could be postponed. 



[1]: https://github.com/alchemyplatform/rundler/blob/6c3691afdee003eac3e5e0bba144dd327978945f/crates/pool/src/chain.rs#L485-L524

dancoombs:
> Synced with @0xfourzerofour offline. the benefits of using the event watcher is we could remove the decoding code and simplified the topic handling logic [1](https://github.com/alchemyplatform/rundler/blob/6c3691afdee003eac3e5e0bba144dd327978945f/crates/pool/src/chain.rs#L485-L524). But also we agreed the priority is not that high and could be postponed.

I still fail to see how the event watcher could be used in a way that can handle reorgs properly, without adding a bunch of complexity.

0xfourzerofour:
> > Synced with @0xfourzerofour offline. the benefits of using the event watcher is we could remove the decoding code and simplified the topic handling logic [1](https://github.com/alchemyplatform/rundler/blob/6c3691afdee003eac3e5e0bba144dd327978945f/crates/pool/src/chain.rs#L485-L524). But also we agreed the priority is not that high and could be postponed.
> 
> I still fail to see how the event watcher could be used in a way that can handle reorgs properly, without adding a bunch of complexity.

[...]

PR #810, opened 2024-09-24
feat(pool): update pool to use alloy types

dancoombs:


crates/pool/src/chain.rs
```diff
@@ -17,23 +17,26 @@ use std::{
     time::Duration,
 };
 
+use alloy_primitives::{Address, B256, U256};
```
dancoombs, line 20: Kept this file a simple type upgrade for now, but added a ticket to use new alloy features to do this better https://github.com/alchemyplatform/rundler/issues/809

crates/pool/src/mempool/mod.rs
```diff
@@ -134,7 +134,7 @@ pub struct PoolConfig {
     pub same_sender_mempool_count: usize,
     /// The minimum fee bump required to replace an operation in the mempool
     /// Applies to both priority fee and fee. Expressed as an integer percentage value
-    pub min_replacement_fee_increase_percentage: u64,
+    pub min_replacement_fee_increase_percentage: u32,
```
0xfourzerofour, line 137: It will be interesting seeing how much of a difference all of the changes to uint sizing has to stack allocation, probably not much but with all the U256 changes might be more than we think especially when lots of ops are in the pool

crates/pool/src/task.rs
```diff
@@ -68,18 +69,22 @@ pub struct PoolTask<P, E06, E07> {
...
 }
 
 #[async_trait]
 impl<P, E06, E07> Task for PoolTask<P, E06, E07>
 where
-    P: Provider,
-    E06: EntryPointProvider<UserOperationV0_6>,
-    E07: EntryPointProvider<UserOperationV0_7>,
+    P: EvmProvider + Clone + 'static,
+    E06: EntryPointProvider<UserOperationV0_6> + Clone + 'static,
```
0xfourzerofour, line 81: whats the reason this needs to be static?

0xfourzerofour (approved):
Looks good just a question about the static lifetime

crates/pool/src/task.rs
0xfourzerofour: I guess it doesnt really matter as it should be allocated for the whole runtime but we just wondering the differences between the task definition below

andysim3d (approved):

crates/pool/src/task.rs
dancoombs: At the moment the `Task` trait is static, so all of these need to be static. Might be able to remove that when we incorporate the Reth task crate.

PR #811, opened 2024-09-25
Fix replacement underpriced forget bug

dancoombs:
## Proposed Changes

  - Updates for rust 1.81
  - Fixes a bug in the builder state machine:
    - Builder receives "replacement underpriced", un-abandons its last known transaction in an attempt to unstick
    - Builder starts to build the next bundle, but the mempool is empty
    - (bug) Builder moves to initial state, forgetting any previous "replacement underpriced", but we remember the last known transaction. We have a mechanism if we know we are in an underpriced state to eventually reset after a configurable amount of blocks. However, we lost the underpriced state in this bug.
    - We get stuck with the priority fee from the last known transaction, with not reset mechanism, until a UO comes in that pays enough.
- Cleaned up a few of the state machine transitions to make it easier to reason about

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -241,11 +241,11 @@ where
             }
             Ok(SendBundleAttemptResult::NoOperationsInitially) => {
                 debug!("No operations available initially");
-                state.complete(Some(SendBundleResult::NoOperationsInitially));
+                state.no_operations();
             }
             Ok(SendBundleAttemptResult::NoOperationsAfterSimulation) => {
                 debug!("No operations available after simulation");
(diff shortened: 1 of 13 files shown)


crates/builder/src/bundle_sender.rs
```diff
@@ -241,11 +241,11 @@ where
             }
             Ok(SendBundleAttemptResult::NoOperationsInitially) => {
                 debug!("No operations available initially");
-                state.complete(Some(SendBundleResult::NoOperationsInitially));
+                state.no_operations();
```
dancoombs, line 244: This is the bug

dancoombs:
TODO: add a unit test for this

andysim3d (approved):

PR #815, opened 2024-09-26
fix(middleware): change mod visibility to pass build.

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/pool/Cargo.toml
+++ b/crates/pool/Cargo.toml
@@ -36,7 +36,6 @@ tokio-util.workspace = true
 tonic.workspace = true
 tonic-health.workspace = true
 tonic-reflection.workspace = true
-tower.workspace = true
 tracing.workspace = true
 url.workspace = true
 
--- a/crates/pool/Cargo.toml
+++ b/crates/pool/Cargo.toml
@@ -36,6 +36,7 @@ tokio-util.workspace = true
 tonic.workspace = true
 tonic-health.workspace = true
 tonic-reflection.workspace = true
+tower.workspace = true
 tracing.workspace = true
 url.workspace = true
 
--- a/crates/provider/src/traits/metrics.rs
+++ b/crates/provider/src/traits/metrics.rs
@@ -16,7 +16,7 @@ use rundler_types::task::traits::RequestExtractor;
 use alloy_json_rpc::RequestPacket;
 
 #[derive(Clone, Copy)]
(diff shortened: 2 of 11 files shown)


crates/provider/src/traits/metrics.rs
```diff
@@ -16,7 +16,7 @@ use rundler_types::task::traits::RequestExtractor;
 use alloy_json_rpc::RequestPacket;
 
 #[derive(Clone, Copy)]
-struct AlloyMethodExtractor;
+pub struct AlloyMethodExtractor;
```
dancoombs, line 19: This should be `pub(crate)`

crates/rpc/src/rpc_metrics.rs
```diff
@@ -15,10 +15,10 @@ use jsonrpsee::types::Request;
 use rundler_types::task::traits::RequestExtractor;
 
 #[derive(Copy, Clone)]
-struct RPCMethodExtractor;
+pub struct RPCMethodExtractor;
```
dancoombs, line 18: Same here

crates/provider/src/traits/metrics.rs
dancoombs: Hm actually I take that back, we might be creating the alloy middleware outside of the provider crate

dancoombs (approved):

crates/pool/Cargo.toml
```diff
@@ -36,6 +36,7 @@ tokio-util.workspace = true
 tonic.workspace = true
 tonic-health.workspace = true
 tonic-reflection.workspace = true
+tower.workspace = true
```
dancoombs, line 39: My bad, forgot to remove this, don't think its needed.

crates/task/src/metrics.rs
```diff
@@ -97,51 +127,37 @@ where
...
             method_name.as_str(),
-            self.protocal.as_str(),
+            self.protocol.as_str(),
         );
 
         let start = Instant::now();
         let mut svc = self.inner.clone();
         let service_name = self.service_name.clone();
-        let protocal = self.protocal.clone();
+        let protocal = self.protocol.clone();
```
dancoombs, line 143: Can you fix this typo? `protocol`

dancoombs (approved):

PR #816, opened 2024-09-26
feat(middleware): add response extractor.

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -


crates/types/src/task/traits.rs
```diff
@@ -17,3 +17,9 @@ pub trait RequestExtractor<R>: Copy + Sync + Send {
     /// Get method name.
     fn get_method_name(request: &R) -> String;
 }
+
+/// Trait to extract response code.
+pub trait ResponseExtractor<R>: Copy + Sync + Send {
```
dancoombs, line 22: Does this need to be `Copy`? Thats a pretty strict bound to put on a trait

crates/types/src/task/traits.rs
```diff
@@ -17,3 +17,9 @@ pub trait RequestExtractor<R>: Copy + Sync + Send {
...
+    /// Get response code.
+    fn get_response_code(response: &R) -> String;
```
dancoombs, line 24: Why is the code a string?

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,226 @@
...
+use alloy_json_rpc::{RequestPacket, ResponsePacket};
+use alloy_transport::{BoxFuture, TransportError};
+use futures_util::FutureExt;
+use rundler_types::task::{
+    metric_recorder::MethodSessionLogger,
+    status_code::{HttpCode, RpcCode},
+};
+use tower::{Layer, Service};
+
+#[allow(dead_code)]
```
dancoombs, line 25: Is this still dead?

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,226 @@
...
+                    }
+                },
             }
+            response
         }
+        .boxed()
     }
 }
+
+// #[pin_project]
```
dancoombs, line 143: Remove dead code

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,226 @@
...
+                        method_logger.record_rpc(RpcCode::InternalError);
+                    }
+                    alloy_json_rpc::RpcError::DeserError { .. } => {
+                        method_logger.record_http(HttpCode::FourHundreds);
+                        method_logger.record_rpc(RpcCode::ParseError);
+                    }
+                    alloy_json_rpc::RpcError::Transport(_) => {
+                        method_logger.record_http(HttpCode::FiveHundreds);
+                        method_logger.record_rpc(RpcCode::ServerError);
+                    }
```
dancoombs, line 134: I'd be surprised if there wasn't a way to eventually get the HTTP code from the `Transport` error here. Is that not exposed in the inner error type?

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,226 @@
...
+            method_logger.done();
+            match &response {
+                Ok(resp) => {
+                    method_logger.record_http(HttpCode::TwoHundreds);
+                    method_logger.record_rpc(get_rpc_status_code(resp));
+                }
+                Err(e) => match e {
+                    alloy_json_rpc::RpcError::ErrorResp(_) => {
+                        method_logger.record_http(HttpCode::FiveHundreds);
+                        method_logger.record_rpc(RpcCode::ServerError);
```
dancoombs, line 109: This seems off to me, you should be able to get the RPC error payload from this response.

I.e: https://github.com/alloy-rs/alloy/blob/a1733071f834d2d965fca21ebb7d7f68982608d1/crates/json-rpc/src/response/error.rs#L29

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,226 @@
...
+            let response = svc.call(request).await;
+            method_logger.done();
+            match &response {
+                Ok(resp) => {
+                    method_logger.record_http(HttpCode::TwoHundreds);
+                    method_logger.record_rpc(get_rpc_status_code(resp));
+                }
+                Err(e) => match e {
+                    alloy_json_rpc::RpcError::ErrorResp(_) => {
+                        method_logger.record_http(HttpCode::FiveHundreds);
```
dancoombs, line 108: How come we log this as a `5xx`? If we're getting an RPC response that means the HTTP layer passed, so wouldn't this be a 200?

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,226 @@
...
+                        method_logger.record_rpc(RpcCode::InvalidRequest);
+                    }
+                    alloy_json_rpc::RpcError::SerError(_) => {
+                        method_logger.record_http(HttpCode::FiveHundreds);
+                        method_logger.record_rpc(RpcCode::InternalError);
+                    }
+                    alloy_json_rpc::RpcError::DeserError { .. } => {
+                        method_logger.record_http(HttpCode::FourHundreds);
+                        method_logger.record_rpc(RpcCode::ParseError);
+                    }
```
dancoombs, line 130: Are all of these actually making network requests? It seems off to record codes that don't actually happen

crates/rpc/src/rpc_metrics.rs
```diff
@@ -11,13 +11,106 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct RpcMetricsMiddleware<S> {
+    service: S,
+    service_name: String,
+}
+
+impl<'a, S> RpcServiceT<'a> for RpcMetricsMiddleware<S>
```
dancoombs, line 56: For the RPC server I think I'd expect there to be 2 separate middleware implementations, one for RPC and one for HTTP. My understanding is that the RPC middleware won't even be called if the HTTP middleware encounters an error, but we should double check that.

crates/task/src/grpc/grpc_metrics.rs
```diff
@@ -11,15 +11,120 @@
...
+        match &res {
+            Poll::Ready(response) => {
+                this.method_logger.done();
+                match response {
+                    Ok(_) => {
+                        this.method_logger.record_http(HttpCode::TwoHundreds);
+                    }
+                    Err(_) => {
+                        // extract the error message form the error trait
+                        this.method_logger.record_http(HttpCode::FiveHundreds);
```
dancoombs, line 121: It seems surprising that the HTTP code isn't exposed somewhere in the error type in the grpc service. Do you know what the type is? May make sense to hardcode to that type so you can access it.

crates/types/src/task/metric_recorder.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// create a session logger.
+    pub fn new(service_name: String, method_name: String, protocol: String) -> Self {
+        Self {
+            start_time: Instant::now(),
+            method_name: method_name.clone(),
+            service_name: service_name.clone(),
+            protocol: protocol.clone(),
+        }
+    }
+    /// start the session. time will be initialized.
```
dancoombs, line 37: nit: newline above

crates/types/src/task/metric_recorder.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+impl MethodSessionLogger {
+    /// create a session logger.
+    pub fn new(service_name: String, method_name: String, protocol: String) -> Self {
+        Self {
+            start_time: Instant::now(),
+            method_name: method_name.clone(),
+            service_name: service_name.clone(),
+            protocol: protocol.clone(),
```
dancoombs, line 34: nit: none of these need to be clone

crates/rpc/src/rpc_metrics.rs
```diff
@@ -11,13 +11,160 @@
...
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
 
-impl RequestExtractor<Request<'static>> for RPCMethodExtractor {
-    fn get_method_name(req: &Request<'static>) -> String {
-        req.method_name().to_string()
+    fn call(&mut self, req: httpRequest<R>) -> Self::Future {
+        let uri = req.uri().clone();
+        let method_name = uri.path().split('/').last().unwrap_or("unknown");
```
0xfourzerofour, line 151: The method name is probably within the json body not the URI path, has this been appended somewhere else?

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,156 @@
...
+    S::Future: Send,
+{
+    type Response = ResponsePacket;
+    type Error = TransportError;
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+    fn call(&mut self, request: RequestPacket) -> Self::Future {
```
dancoombs, line 84: nit: newline between functions

crates/provider/Cargo.toml
```diff
@@ -30,8 +30,12 @@ anyhow.workspace = true
 async-trait.workspace = true
 auto_impl.workspace = true
 thiserror.workspace = true
+futures-util.workspace = true
+tower.workspace = true
 tracing.workspace = true
 url.workspace = true
+pin-project.workspace = true
```
dancoombs, line 37: I might be missing it, but is this used anywhere?

crates/provider/src/alloy/metrics.rs
```diff
@@ -11,22 +11,156 @@
...
+        x if (200..=299).contains(&x) => HttpCode::TwoHundreds,
+        x if (400..=499).contains(&x) => HttpCode::FourHundreds,
+        x if (500..=599).contains(&x) => HttpCode::FiveHundreds,
+        _ => HttpCode::Other,
+    }
+}
+
+fn get_rpc_status_code(response_packet: &ResponsePacket) -> RpcCode {
+    let response: &alloy_json_rpc::Response = match response_packet {
+        ResponsePacket::Batch(resps) => &resps[0],
```
dancoombs, line 158: Is this safe? Is there guaranteed to be one resp in the array?

crates/provider/src/alloy/mod.rs
```diff
@@ -34,8 +36,16 @@ pub fn new_alloy_evm_provider(rpc_url: &str) -> anyhow::Result<impl EvmProvider
...
     rpc_url: &str,
-) -> anyhow::Result<impl AlloyProvider<Http<Client>> + Clone> {
+) -> anyhow::Result<
+    impl AlloyProvider<AlloyMetricMiddleware<RetryBackoffService<Http<Client>>>> + Clone,
+> {
     let url = Url::parse(rpc_url).context("invalid rpc url")?;
-    let client = ClientBuilder::default().http(url);
-    Ok(ProviderBuilder::new().on_client(client))
+    let metric_layer = AlloyMetricLayer::default();
+    let retry_layer = alloy_transport::layers::RetryBackoffLayer::new(10, 500, 0);
```
dancoombs, line 44: Just making sure, are these the same parameters we used to use?

crates/provider/src/alloy/mod.rs
```diff
@@ -34,8 +36,16 @@ pub fn new_alloy_evm_provider(rpc_url: &str) -> anyhow::Result<impl EvmProvider
...
+    let client = ClientBuilder::default()
+        .layer(metric_layer)
+        .layer(retry_layer)
```
dancoombs, line 47: I _think_ these should be switched. We want the retry layer to call through the metric layer so that we can see the 429 errors in our statistics. Worth double checking if that is the case, but based on the type above it seems we'd want to switch.

crates/rpc/src/rpc_metrics.rs
dancoombs, line 151: Is this ever expected to be there? This is for jsonprsee correct? Won't the method name always be in the json blob?

crates/rpc/src/rpc_metrics.rs
```diff
@@ -11,13 +11,160 @@
...
+        let mut method_logger = MethodSessionLogger::new(
+            self.service_name.clone(),
+            method_name.to_string(),
+            "http".to_string(),
+        );
+        method_logger.start();
+        let mut svc = self.service.clone();
+        async move {
+            let rp = svc.call(req).await;
+            method_logger.done();
```
dancoombs, line 161: Seems like overkill to have the latency metrics on both the RPC and HTTP middlewares. Due to the high cardinality of those, I think its best we only capture one. Thoughts on how to do that?

crates/task/src/grpc/grpc_metrics.rs
```diff
@@ -11,15 +11,119 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let res = this.response_future.poll(cx);
+        match &res {
+            Poll::Ready(response) => {
+                this.method_logger.done();
 
-/// http request method extractor.
-pub struct HttpMethodExtractor;
+                let http_status = response.as_ref().ok().map(|response| response.status());
```
dancoombs, line 118: When is this not `ok`?

crates/types/src/task/metric_recorder.rs
```diff
@@ -0,0 +1,122 @@
...
+pub struct MethodSessionLogger {
+    start_time: Instant,
+    service_name: String,
+    method_name: String,
+    protocol: String,
+}
+
+impl MethodSessionLogger {
+    /// create a session logger.
+    pub fn new(service_name: String, method_name: String, protocol: String) -> Self {
```
dancoombs, line 29: Maybe instead of having `new` and `start` we combine them into a single `start` static function that returns `Self`. Can you think of a reason why you wouldn't always call `start` after `new`?

crates/provider/src/alloy/mod.rs
andysim3d: yes, I reference to our exisiting retry config here: https://github.com/alchemyplatform/rundler/blob/6c3691afdee003eac3e5e0bba144dd327978945f/crates/provider/src/ethers/provider.rs#L304-L310.

andysim3d: https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html#order first added layer will be executed first. metric layer should on top of retry layer to capture all 429s.

dancoombs: I think the metric layer should be below the retry layer so that it sees all of the 429s. The retry layer should be calling through the metric layer.

andysim3d: ohhhh yes. sorry I was thinking we are the server.

Cargo.toml
```diff
@@ -66,15 +65,13 @@ serde = "1.0.210"
...
 reqwest = { version = "0.12.8", default-features = false, features = ["rustls-tls"] }
-rustls = "0.23.13"
 thiserror = "1.0.64"
 tokio = { version = "1.39.3", default-features = false, features = ["rt", "sync", "time"]}
 tokio-util = "0.7.12"
 tonic = "0.12.3"
 tonic-build = "0.12.3"
 tonic-health = "0.12.3"
 tonic-reflection = "0.12.3"
-tonic-types = "0.12.3"
```
dancoombs, line 77: How did you know to remove the crates in this file?

Cargo.toml
```diff
@@ -52,7 +52,6 @@ anyhow = "1.0.89"
 async-trait = "0.1.83"
 auto_impl = "1.2.0"
 aws-config = { version = "1.5.6", default-features = false }
-cargo-husky = { version = "1", default-features = false, features = ["user-hooks"] }
```
dancoombs, line 55: We've used this tool before to add the hooks from the `.cargo-husky` folder

dancoombs: Looks like we need to set it as a dev-dependency instead https://github.com/rhysd/cargo-husky?tab=readme-ov-file#usage

bin/rundler/Cargo.toml
```diff
@@ -47,3 +47,4 @@ tracing.workspace = true
 tracing-appender = "0.2.3"
 tracing-log = "0.2.0"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter", "fmt", "json"] }
+cargo-husky = "1.5.0"
```
dancoombs, line 50: This is now added as a dependency for our bin crate. It should be a `dev-dependency`, likely on one of our library crates. Something like this:

https://github.com/rhysd/cargo-husky/issues/18#issuecomment-1500989739

andysim3d: I tried, and got error 
```
error: failed to parse manifest at `/Users/andy/project/rundler/Cargo.toml`

Caused by:
  this virtual manifest specifies a `dev-dependencies` section, which is not allowed
error: failed to push some refs to 'github.com:alchemyplatform/rundler.git'
```

andysim3d: my approach is the same as https://github.com/rhysd/cargo-husky/issues/18#issuecomment-1500989739. add husky in the root cargo.toml's workspace.dependencies, and reference in type/cargo.toml as dev-dependencies.

dancoombs: Leave `cargo-husky` as it was in our workspace `Cargo.toml` (with the same feature flags), add `cargo-husky.workspace = true` to the types crate `[dev-dependencies]`

andysim3d: hum I think it is a switching workspace issue. fixed now.

crates/types/Cargo.toml
```diff
@@ -27,12 +27,14 @@ serde.workspace = true
...
 strum.workspace = true
 thiserror.workspace = true
+metrics.workspace = true
 
 mockall = {workspace = true, optional = true }
 
 [dev-dependencies]
 rundler-types = { workspace = true, features = ["test-utils"] }
 alloy-primitives = { workspace = true, features = ["rand"] }
+cargo-husky = { version = "1", default-features = false, features = ["user-hooks"] }
```
dancoombs, line 37: Can't this be `cargo-husky.workspace = true`

dancoombs (approved):

PR #817, opened 2024-09-27
WIP

dancoombs:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/task/src/metrics.rs
+++ b/crates/task/src/metrics.rs
@@ -24,12 +24,12 @@ use rundler_types::task::traits::RequestExtractor;
 use tower::{Layer, Service};
 
 /// tower network layer: https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct MetricsLayer<T, R> {
     service_name: String,
     protocal: String,
-    _request_extractor_: PhantomData<T>,
-    _request_type_: PhantomData<R>,
+    _request_extractor: PhantomData<T>,
+    _request_type: PhantomData<R>,
 }
 
 impl<T, R> MetricsLayer<T, R>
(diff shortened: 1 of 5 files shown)


crates/task/src/metrics.rs
```diff
@@ -97,43 +127,31 @@ where
...
             method_name.as_str(),
-            self.protocal.as_str(),
+            self.protocol.as_str(),
         );
 
         let start = Instant::now();
         let mut svc = self.inner.clone();
         let service_name = self.service_name.clone();
-        let protocal = self.protocal.clone();
+        let protocal = self.protocol.clone();
```
dancoombs, line 143: Typo here still `protocol`

crates/task/src/metrics.rs
```diff
@@ -41,8 +41,22 @@ where
...
             protocal,
-            _request_extractor_: PhantomData,
-            _request_type_: PhantomData,
+            _request_extractor: PhantomData,
+            _request_type: PhantomData,
+        }
+    }
+}
+
+impl<T, R> Clone for MetricsLayer<T, R>
```
dancoombs, line 50: Needed to manually implement clone because `derive(Clone)` requires that `T, R` be `Clone`, where they don't need to be since we're using phantom data.

crates/task/src/metrics.rs
```diff
@@ -52,42 +66,58 @@ where
...
     service_name: String,
-    protocal: String,
-    _request_extractor_: PhantomData<T>,
-    _request_type_: PhantomData<R>,
+    protocol: String,
+    _request_extractor: PhantomData<T>,
+    _request_type: PhantomData<R>,
+}
+
+impl<S, T, R> Clone for MetricsMiddleware<S, T, R>
```
dancoombs, line 84: Same reason here

crates/task/src/metrics.rs
```diff
@@ -52,42 +66,58 @@ where
...
+impl<S, T, R> Service<R> for MetricsMiddleware<S, T, R>
 where
-    S: Service<Request> + Send + Sync + Clone + 'static,
-    S::Future: Send + Sync + 'static,
-    T: RequestExtractor<Request> + 'static,
-    Request: Send + Sync + 'static,
+    S: Service<R> + Send + Clone + 'static,
+    S::Future: Send + 'static,
+    T: RequestExtractor<R> + 'static,
+    R: Send + 'static,
```
dancoombs, line 120: Removed all of the `Sync` from these bounds as its not required by `Service`. Tbh I'm unsure why but this made the compiler happy, example here https://github.com/hyperium/tonic/blob/master/examples/src/tower/server.rs

crates/task/src/metrics.rs
```diff
@@ -97,43 +127,31 @@ where
...
+        let protocal = self.protocol.clone();
         async move {
             let rsp = svc.call(request).await;
             MethodMetrics::record_request_latency(
-                method_name.as_str(),
-                service_name.as_str(),
-                protocal.as_str(),
+                &method_name,
+                &service_name,
+                &protocal,
```
dancoombs, line 149: Very common to convert String -> &str with a dereference

andysim3d (approved):

Issue #789, opened 2024-09-10
[alloy] Add unit tests to alloy providers where applicable


dancoombs:
**Describe the feature**
The alloy provider module contains zero unit tests. Unit tests are difficult as the majority of the logic requires making network calls.

We should evaluate the module and add unit tests where applicable.

We should be careful not to over re-factor the code in order to make it unit testable. I'd prefer easily readable, modifiable, code vs. code with extra interfaces introduced just for testing. We can rely on integration testing to catch some of these cases.

I'd also like to avoid spinning up anvil instances just for unit tests, as that feels very heavy handed.

I'm okay if this ticket is closed with a no-op after evaluation.

dancoombs:
See #820 for a few examples of where unit tests would be helpful

andysim3d:
#826

PR #820, opened 2024-10-02
feat(bin): update binary to alloy, pass spec tests w/ fixes

dancoombs:


crates/builder/src/bundle_proposer.rs
```diff
@@ -470,7 +470,7 @@ where
         let mut context = ProposalContext::<UO>::new();
         let mut paymasters_to_reject = Vec::<EntityInfo>::new();
 
-        let mut gas_spent = self.settings.chain_spec.transaction_intrinsic_gas;
+        let mut gas_spent = self.settings.chain_spec.transaction_intrinsic_gas as u128;
```
dancoombs, line 473: This is annoying, but since serde doesn't support u128 I made all of these chain_spec settings u64.

I think in Alloy v0.4 we're going to be changing everything to u64, so will hold off on improving this until then.

crates/provider/src/alloy/evm.rs
```diff
@@ -152,12 +152,10 @@ where
...
-        Ok(Self::get_block(self, BlockId::pending())
-            .await?
-            .context("pending block should exist")?
-            .header
-            .base_fee_per_gas
-            .context("pending block should have a nonempty base fee")?)
+        let fee_history = self.fee_history(1, BlockNumberOrTag::Pending, &[]).await?;
+        Ok(fee_history
+            .next_block_base_fee()
+            .context("should have a next block base fee")?)
```
dancoombs, line 158: This is the most concerning change. We're now requiring every RPC to support `eth_feeHistory`. I have a feeling this won't fly, so will need to figure out another approach here. Will be doing that in a follow up PR.

crates/provider/src/alloy/entry_point/v0_7.rs
```diff
@@ -421,13 +421,16 @@ where
 }
 
 fn add_simulations_override(state_override: &mut StateOverride, addr: Address) {
-    state_override.insert(
-        addr,
-        AccountOverride {
+    // Do nothing if the caller has already overridden the entry point code.
```
dancoombs, line 424: This was a bug, consider adding a unit test

crates/provider/src/alloy/evm.rs
```diff
@@ -290,6 +288,7 @@ where
         };
         overrides.insert(helper_addr, account);
 
+        addresses.sort();
```
dancoombs, line 291: This was a bug, consider adding a unit test

dancoombs: Should do as part of https://github.com/alchemyplatform/rundler/issues/789

crates/provider/src/alloy/evm.rs
dancoombs: https://github.com/alchemyplatform/rundler/issues/821

crates/builder/src/bundle_proposer.rs
dancoombs: https://github.com/alchemyplatform/rundler/issues/822

crates/pool/src/chain.rs
```diff
@@ -161,7 +161,7 @@ impl<P: EvmProvider> Chain<P> {
                     .cloned()
                     .collect::<Vec<_>>(),
             )
-            .events(events.iter());
+            .event_signature(events);
```
dancoombs, line 164: This was a bug, might want to add to unit tests

0xfourzerofour (approved):
Looks good, the type casting is a little annoying but I can see its been added to new tickets for a follow up so we are good

andysim3d (approved):

PR #824, opened 2024-10-03
feat: replace existing tasks with reth tasks

### LINKED ISSUE #380: task: Incorporate Reth task management crate
**Describe the feature**
Reth has a great task management crate that we should use instead of our hand-rolled one. We can contribute back to theirs for any improvements.

https://github.com/paradigmxyz/reth/blob/main/crates/tasks/Cargo.toml

dancoombs:
Closes #380


crates/builder/src/signer/aws.rs
```diff
@@ -43,36 +42,32 @@ impl KmsSigner {
...
-                .to_owned();
+                .to_owned()
         };
 
         let signer = AwsSigner::new(client, key_id, Some(chain_id))
             .await
             .context("should create signer")?;
 
-        let monitor_guard = SpawnGuard::spawn_with_guard(monitor_account_balance(
+        task_spawner.spawn(Box::pin(monitor_account_balance(
```
0xfourzerofour, line 65: this is way cleaner nice

0xfourzerofour (approved):
Lets go! looks good to me

PR #830, opened 2024-10-07
fix(builder): don't forget underpriced info when no ops

dancoombs:
Cherry-pick of https://github.com/alchemyplatform/rundler/pull/811 to `feat/v0.4`

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -249,11 +249,11 @@ where
             }
             Ok(SendBundleAttemptResult::NoOperationsInitially) => {
                 debug!("No operations available initially");
-                state.complete(Some(SendBundleResult::NoOperationsInitially));
+                state.no_operations();
             }
             Ok(SendBundleAttemptResult::NoOperationsAfterSimulation) => {
                 debug!("No operations available after simulation");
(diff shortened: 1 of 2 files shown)


crates/builder/src/bundle_sender.rs
```diff
@@ -718,35 +727,47 @@ impl<T: TransactionTracker, TRIG: Trigger> SenderMachineState<T, TRIG> {
...
+        self.send_result(SendBundleResult::NoOperationsInitially);
+
+        self.inner = match &self.inner {
+            InnerState::Building(s) => InnerState::Building(BuildingState {
+                wait_for_trigger: true,
+                fee_increase_count: s.fee_increase_count,
+                underpriced_info: s.underpriced_info,
+            }),
+            _ => {
+                panic!("invalid state transition, no_operations called when not in building state")
```
0xfourzerofour, line 762: Will this exit rundler or be caught in the runtime and respawn?

0xfourzerofour (approved):
Looks good just a question on the panic

crates/builder/src/bundle_sender.rs
dancoombs: Exit rundler, this is a code issue that shouldn't happen

PR #834, opened 2024-10-08
feat: prep for da oracle with cache

dancoombs:
## Proposed Changes

- Move da oracle to its own module
- Use trait object instead of enum for da oracle
- Require a BlockHashOrNumber in calc_da_gas call, for use as future cache key
- Clean up some usage of block_hash in codebase


crates/sim/src/precheck.rs
```diff
@@ -43,8 +43,11 @@ pub trait Prechecker: Send + Sync {
     type UO: UserOperation;
 
     /// Run the precheck on the given operation and return an error if it fails.
-    async fn check(&self, op: &Self::UO) -> Result<(), PrecheckError>;
+    async fn check(&self, op: &Self::UO, block: BlockHashOrNumber) -> Result<(), PrecheckError>;
+
     /// Update and return the bundle fees.
+    ///
+    /// This MUST be called at block boundaries before checking any operations.
```
dancoombs, line 50: This should be abled to use the same block cache based approach.

I think we should break out a `FeeProvider` trait that can do this. Can do this at the same time as the pending base fee oracle.

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -433,12 +441,14 @@ where
...
-        self.prechecker.check(&versioned_op).await?;
+        self.prechecker
+            .check(&versioned_op, block_hash.into())
+            .await?;
 
         // Only let ops with successful simulations through
         let sim_result = self
             .simulator
-            .simulate_validation(versioned_op, None, None)
+            .simulate_validation(versioned_op, block_hash, None)
```
dancoombs, line 451: Ugh, this creates a gnarly race condition: the node may have already emitted a new block, but we haven't processed it yet, and the UO depends on something in that block.

For example in the spec tests:
1) Fund a SCA, wait for mine, minded in block X
2) Immediately submit a UO, pool is still on block X-1, UO fails.

dancoombs: Same issue applies to the prechecker

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -419,6 +433,18 @@ where
             );
         }
 
+        // NOTE: We get the latest block from the provider here to avoid a race condition
+        // where the pool is still processing the previous block, but the user may have been
+        // notified of a new block.
+        //
+        // This doesn't clear all race conditions, as the pool may need to update its state before
+        // a UO can be valid, i.e. for replacement.
+        let (block_hash, block_number) = self
```
dancoombs, line 442: This call is just moved up from the simulator into this function as the block is now needed by the prechecker as well.

Comment is there because I got burned by trying to use `state.block_hash` instead and ran into this race condition.

0xfourzerofour (approved):
LGTM

Issue #744, opened 2024-06-26
[pool] Insufficient preVerificationGas throws off latency tracking metric


dancoombs:
**Describe the bug**
On L2s that contain dynamic preVerificationGas, one of the filters for being eligible to mine is having a sufficient PVG value for the current network gas prices. If below, the UO will not get mined. This should not count against its time to mine latency in the mempool.

**To reproduce**
* Submit a UO that has an underpriced PVG
* Watch time to mine metrics

**Expected behavior**
A UO should only have its time to mine increased when it is eligible to be mined.

**Additional context**
One way to solve this would be to re-calculate PVG on every block. This is already done by the builder. The solution here is likely to move this calculation down into the pool in a batched manner, limiting the amount of `eth_call` requests.

Its possible that we can start calculating the dynamic PVG without making an `eth_call` per UO. The calculation is only based on network constants that need to be queried once per block. Until we move the PVG calculation into Rundler its unlikely we'll be able to efficiently serve this. This move is healthy for Rundler in general as it limits RPC calls.

For example on optimism we can do this calculation directly: https://github.com/ethereum-optimism/optimism/blob/7f403ea5f316a9085061bc63b6618afcdbca1278/packages/contracts-bedrock/src/L2/GasPriceOracle.sol#L226 by only querying a few constants that only change once per block.

dancoombs:
Arbitrum implemented here: https://github.com/OffchainLabs/nitro/blob/3ecd01ebcf768025ecadb96e50b90d89bc4c6bc5/execution/nodeInterface/NodeInterface.go#L509

dancoombs:
This not only throws off the latency tracking metric, but has a worse bug where it can stall the bundler.

If the mempool becomes full of UOs that have underpriced PVG, but have high enough priority fees to be in the top 128 (our default max bundle size) best operations, they can be crowding out UOs that have sufficiently priced PVG.

The fix here should fix both of these issues

PR #836, opened 2024-10-09
feat(provider): add a local bedrock DA provider

dancoombs:
Related to #744 

## Proposed Changes

  - Add a local DA provider for optimism bedrock chains. This requires a single network call per block to calculate the required DA gas for every user operation.
  - The DA provider contains an LRU cache to ensure that we only issue 1 network call update per block.
  - Compression happens locally now, via a `fastlz` binding at a specific commit. This commit is specifically the same used by [`LibZip.sol`](https://github.com/Vectorized/solady/blob/8b0601e1573ed17a583fdab2b2ebfb895507ec15/src/utils/LibZip.sol#L19), which is used by the optimism DA oracle contract.
 
TODO:
- [x] Utilize a helper contract to get all of the `BlockDAData` in a single `eth_call`
- [x] Consider the implications of running the `fastlz` compression on each block for each user operation. Its highly likely we need the following:
- [x] Cache the result of the `fastlz` compressed size for each UO by hash. This only needs to be done once.
- [x] Consider moving the `fastlz` compression to a specific threadpool for blocking operations.

### DIFF
--- a/crates/provider/src/alloy/da/local/bedrock.rs
+++ b/crates/provider/src/alloy/da/local/bedrock.rs
@@ -1,0 +1,172 @@
+// This file is part of Rundler.
+//
+// Rundler is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Lesser General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version.
+//
+// Rundler is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+// without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along with Rundler.
+// If not, see https://www.gnu.org/licenses/.
+
+use alloy_primitives::{Address, Bytes};
+use alloy_provider::Provider as AlloyProvider;
+use alloy_sol_types::sol;
+use alloy_transport::Transport;
+use rundler_utils::cache::LruMap;
+use tokio::sync::Mutex;
+use GasPriceOracle::GasPriceOracleInstance;
+
+use crate::{alloy::da::DAGasOracle, BlockHashOrNumber, ProviderResult};
+
+// From https://github.com/ethereum-optimism/optimism/blob/f93f9f40adcd448168c6ea27820aeee5da65fcbd/packages/contracts-bedrock/src/L2/GasPriceOracle.sol#L54
+sol! {
+    #[sol(rpc)]
+    interface GasPriceOracle {
+        bool public isFjord;
+
+        function baseFeeScalar() public view returns (uint32);
+        function l1BaseFee() public view returns (uint32);
+        function blobBaseFeeScalar() public view returns (uint32);
+        function blobBaseFee() public view returns (uint32);
+    }
+}
+
+const DECIMAL_SCALAR: u128 = 1_000_000_000_000;
+const COST_INTERCEPT: i128 = -42_585_600;
+const COST_FASTLZ_COEF: i128 = 836_500;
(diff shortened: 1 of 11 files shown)


crates/provider/src/alloy/da/local/bedrock.rs
```diff
@@ -0,0 +1,176 @@
...
+            .call()
+            .await
+            .map(|r| r.isFjord)
+            .unwrap_or(false)
+    }
+
+    async fn get_block_da_data(&self, block: BlockHashOrNumber) -> ProviderResult<BlockDAData> {
+        assert!(self.is_fjord().await);
+
+        let base_fee_scalar = self
```
dancoombs, line 121: TODO: use a helper contract to make this 1 network call.

crates/provider/src/alloy/da/local/bedrock.rs
```diff
@@ -0,0 +1,176 @@
...
+            base_fee_scalar,
+            l1_base_fee,
+            blob_base_fee_scalar,
+            blob_base_fee,
+        })
+    }
+
+    fn fjord_l1_fee(data: Bytes, block_da_data: &BlockDAData) -> u128 {
+        let mut buf = vec![0; data.len() * 2];
+        let compressed = rundler_bindings_fastlz::compress(&data, &mut buf);
```
dancoombs, line 154: TODO: consider doing this on a specific blocking thread pool

crates/provider/src/alloy/da/local/bedrock.rs
```diff
@@ -0,0 +1,176 @@
...
+        let compressed_length = compressed.len() as u64;
```
dancoombs, line 155: TODO: cache this value.

crates/provider/src/alloy/da/mod.rs
```diff
@@ -68,6 +70,82 @@ where
...
+    use alloy_provider::ProviderBuilder;
+    use alloy_sol_types::SolValue;
+    use rundler_contracts::v0_7::PackedUserOperation;
+
+    use super::*;
+
+    // This test may begin to fail if an optimism sepolia fork changes how the L1 gas oracle works.
+    // If that happens, we should update the local bedrock oracle to match the new fork logic in
+    // a backwards compatible way based on the fork booleans in the contract.
+    #[tokio::test]
```
dancoombs, line 93: Likely have to set these tests to `#[ignore]` due to the flakey-ness of the RPC endpoint.

crates/provider/src/alloy/da/local/bedrock.rs
```diff
@@ -0,0 +1,172 @@
...
+    #[sol(rpc)]
+    interface GasPriceOracle {
+        bool public isFjord;
+
+        function baseFeeScalar() public view returns (uint32);
+        function l1BaseFee() public view returns (uint32);
+        function blobBaseFeeScalar() public view returns (uint32);
+        function blobBaseFee() public view returns (uint32);
+    }
+}
```
moldy530, line 35: when `react! {...}` so we can do everything in rust XD

this is really cool

crates/provider/src/alloy/da/local/bedrock.rs
```diff
@@ -0,0 +1,172 @@
...
+{
+    async fn estimate_da_gas(
+        &self,
+        _to: Address,
+        data: Bytes,
+        block: BlockHashOrNumber,
+        gas_price: u128,
+    ) -> ProviderResult<u128> {
+        let block_da_data = {
+            let mut cache = self.block_da_data_cache.lock().await;
```
moldy530, line 86: lol these names are amazing "block da data cache" "estimate da gas" -- "do da thing". 

ok I'm done these comments are not at all helpful

dancoombs: lol I read it the same way. Got 50 cent stuck in my head writing this

PR #838, opened 2024-10-10
fix: update cargo.lock to compile project form main.

andysim3d:
[Closes/Fixes] #
fix compile error on my local box: 
```
➜  rundler git:(main) make build 
cargo build --all --all-features
   Compiling serde_derive v1.0.160
   Compiling serde v1.0.160
   Compiling time-core v0.1.0
   Compiling time-macros v0.2.8
   Compiling time v0.3.20
error[E0282]: type annotations needed for `Box<_>`
  --> /Users/andy/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.20/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
   = note: this is an inference error on crate `time` caused by an API change in Rust 1.80.0; update `time` to version `>=0.3.35` by calling `cargo update`
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1161,6 +1161,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "deranged"
+version = "0.3.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
+dependencies = [
+ "powerfmt",
+ "serde",
+]
+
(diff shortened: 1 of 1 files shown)


dancoombs:
Hm, this is likely because you are compiling with Rust v1.81 and we haven't yet updated support for that yet. Try compiling with v1.80

I think we should just update `main` to Rust v1.81 support

dancoombs:
This will be added in v0.4

PR #841, opened 2024-10-11
feat(rpc): log rpc error message.

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/provider/src/alloy/metrics.rs
+++ b/crates/provider/src/alloy/metrics.rs
@@ -85,7 +85,7 @@ where
         let method_name = get_method_name(&request);
         let method_logger = MethodSessionLogger::start(
             "alloy_provider_client".to_string(),
-            method_name,
+            method_name.clone(),
             "rpc".to_string(),
         );
         let mut svc = self.service.clone();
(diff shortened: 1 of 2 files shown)


crates/provider/src/alloy/metrics.rs
```diff
@@ -96,6 +96,13 @@ where
                 Ok(resp) => {
                     method_logger.record_http(HttpCode::TwoHundreds);
                     method_logger.record_rpc(get_rpc_status_code(resp));
+                    if resp.is_error() {
```
dancoombs, line 99: Should we also log the error below in the `Err(e)` branch?

crates/rpc/src/rpc_metrics.rs
```diff
@@ -75,6 +75,11 @@ where
                 method_logger.record_http(HttpCode::TwoHundreds);
                 method_logger.record_rpc(RpcCode::Success);
             } else if let Some(error) = rp.as_error_code() {
+                tracing::error!(
```
dancoombs, line 78: I don't think we need this. This is our server, so its actually expected that there are RPC errors here (i.e. when a user submits a bad request).

andysim3d (commented):
feat(logging): enrich error logs.

dancoombs (approved):

PR #842, opened 2024-10-11
feat(provider): add cached DA oracles for bedrock/nitro

dancoombs:
## Proposed Changes

  - Bedrock DA oracle with caching
    - Add a local DA provider for optimism bedrock chains. This requires a single network call per block to calculate the required DA gas for every user operation.
    - The DA provider contains an LRU cache to ensure that we only issue 1 network call update per block.
    - Compression happens locally now, via a fastlz binding at a specific commit. This commit is specifically the same used by [LibZip.sol](https://github.com/Vectorized/solady/blob/8b0601e1573ed17a583fdab2b2ebfb895507ec15/src/utils/LibZip.sol#L19), which is used by the optimism DA oracle contract.
  - Nitro DA oracle with caching
    - The oracle only makes 1 network call per new UO + 1 network call per block
    - Due to the nitro implementation, there is the possibility of rounding errors, which are reduced here by scaling. Errors should underestimate by a small margin


andysim3d (approved):

dancoombs:
Will add more tests in this ticket: https://github.com/alchemyplatform/rundler/issues/845

Want to get the mempool usage PR up quickly.

crates/provider/src/alloy/da/arbitrum.rs
```diff
@@ -63,17 +64,43 @@ where
...
+            DAGasUOData::Empty,
+            DAGasBlockData::Empty,
+        ))
+    }
+
+    async fn block_data(&self, _block: BlockHashOrNumber) -> ProviderResult<DAGasBlockData> {
+        Ok(DAGasBlockData::Empty)
+    }
+
+    async fn uo_data(
```
andysim3d, line 89: question: I get the idea of this method and above are for impl the trait. could we move those impl in the trait definition so they by default return empty?

andysim3d (approved):

crates/provider/src/alloy/da/arbitrum.rs
dancoombs: https://github.com/alchemyplatform/rundler/issues/850

Issue #845, opened 2024-10-14
[provider] Add more unit test for cached DA oracles


dancoombs:
**Describe the feature**
The cached DA oracle implementations, bedrock and nitro, are missing unit tests for their synchronous interfaces.

dancoombs:
Also add unit tests for DA gas tracking in pool/builder.

Issue #574, opened 2024-01-30
Move entrypoint filtering before it hits the mempool


0xfourzerofour:
When there is a chain update, each update will need to check that it was performed against the correct entrypoint. We should move this logic a step above so it cleans up the code on state changes

PR #847, opened 2024-10-15
feat(pool): add da tracking to pool

### LINKED ISSUE #744: [pool] Insufficient preVerificationGas throws off latency tracking metric

dancoombs:
Closes #744 

## Proposed Changes

  - Add `da_gas_tracking_enabled` CLI parameter.
  - Cache any returned DA UO data with each UO in the pool upon entry
  - If tracking is enabled, on each block, pull the block data, then sync calculate the required pvg of each UO to set eligibility
  - Only return eligible UOs to the builder when queried
  - Repeat the sync calc in the builder ~~(can this be removed?)~~

TODO
- [X] ~~Determine if we can remove the builder PVG check~~ Decided the builder must keep this, as it may raise the fees compared to the initial fees used by the mempool.
- [x] Performance check on the `do_maintenance` call on the pool when under heavy load
- [x] Doc updates
- [x] ~~More tests~~ https://github.com/alchemyplatform/rundler/issues/845

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -32,11 +32,12 @@ use rundler_provider::{
     BundleHandler, DAGasProvider, EntryPoint, EvmProvider, HandleOpsOut, SignatureAggregator,
 };
 use rundler_sim::{
-    gas, ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError, SimulationResult,
-    Simulator, ViolationError,
+    ExpectedStorage, FeeEstimator, PriorityFeeMode, SimulationError, SimulationResult, Simulator,
+    ViolationError,
 };
 use rundler_types::{
     chain::ChainSpec,
+    da::DAGasBlockData,
     pool::{Pool, PoolOperation, SimulationViolation},
     Entity, EntityInfo, EntityInfos, EntityType, EntityUpdate, EntityUpdateType, GasFees,
     Timestamp, UserOperation, UserOperationVariant, UserOpsPerAggregator, BUNDLE_BYTE_OVERHEAD,
@@ -44,7 +45,7 @@ use rundler_types::{
 };
 use rundler_utils::{emit::WithEntryPoint, math};
 use tokio::{sync::broadcast, try_join};
-use tracing::{error, info, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::emit::{BuilderEvent, ConditionNotMetReason, OpRejectionReason, SkipReason};
 
@@ -155,6 +156,7 @@ pub(crate) struct Settings {
     pub(crate) beneficiary: Address,
     pub(crate) bundle_priority_fee_overhead_percent: u32,
     pub(crate) priority_fee_mode: PriorityFeeMode,
+    pub(crate) da_gas_tracking_enabled: bool,
 }
 
 #[async_trait]
@@ -200,7 +202,7 @@ where
             return Err(BundleProposerError::NoOperationsInitially);
         }
 
-        tracing::debug!("Starting bundle proposal with {} ops", ops.len());
+        debug!("Starting bundle proposal with {} ops", ops.len());
 
         // (0) Determine fees required for ops to be included in a bundle
         // if replacing, just require bundle fees increase chances of unsticking
@@ -214,10 +216,30 @@ where
             .filter_map(|op| op.uo.paymaster())
             .collect::<Vec<Address>>();
 
+        let da_block_data = if self.settings.da_gas_tracking_enabled {
+            match self.entry_point.block_data(block_hash.into()).await {
+                Ok(block_data) => Some(block_data),
+                Err(e) => {
+                    error!("Failed to get block data for block hash {block_hash:?}, falling back to async da gas calculations: {e:?}");
+                    None
+                }
+            }
+        } else {
+            None
+        };
+
         // (1) Filter out ops that don't pay enough to be included
         let fee_futs = ops
             .into_iter()
-            .map(|op| self.check_fees(op, block_hash, base_fee, required_op_fees))
+            .map(|op| {
+                self.check_fees(
+                    op,
+                    block_hash,
+                    da_block_data.as_ref(),
+                    base_fee,
+                    required_op_fees,
+                )
+            })
             .collect::<Vec<_>>();
         let ops = future::join_all(fee_futs)
             .await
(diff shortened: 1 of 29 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -154,8 +165,19 @@ impl PoolInner {
         ret
     }
 
-    pub(crate) fn best_operations(&self) -> impl Iterator<Item = Arc<PoolOperation>> {
-        self.best.clone().into_iter().map(|v| v.po)
+    pub(crate) fn best_operations(&self) -> impl Iterator<Item = Arc<PoolOperation>> + '_ {
+        self.best.clone().into_iter().filter_map(|p| {
+            // Query into by_hash to get eligible status
```
dancoombs, line 170: TODO: this is bug prone.

crates/pool/src/mempool/pool.rs
```diff
@@ -181,36 +204,82 @@ impl PoolInner {
...
+                let required_pvg = op.uo().required_pre_verification_gas(
+                    &self.config.chain_spec,
+                    1,
+                    required_da_gas,
+                );
+                let actual_pvg = op.uo().pre_verification_gas();
+
+                if actual_pvg < required_pvg {
+                    if op.eligible {
+                        op.eligible = false; // TODO(danc): THIS ONLY SETS IT ON THE BY_HASH VERSION
```
dancoombs, line 238: TODO

crates/pool/src/mempool/pool.rs
```diff
@@ -551,9 +623,18 @@ impl PoolInner {
...
     submission_id: u64,
+    eligible: bool,
 }
 
 impl OrderedPoolOperation {
+    fn new(po: Arc<PoolOperation>, submission_id: u64) -> Self {
+        Self {
+            po,
+            submission_id,
+            eligible: true,
```
dancoombs, line 634: TODO: this is actually not correct. We allow UOs into the pool with a PVG < required. We should actually do the calculation when they are added.

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -669,7 +673,8 @@ where
         }
 
         // get the best operations from the pool
-        let ordered_ops = self.state.read().pool.best_operations();
+        let state = self.state.read();
+        let ordered_ops = state.pool.best_operations();
```
dancoombs, line 677: TODO: put this back

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -328,19 +337,22 @@ where
...
                 gas_fees,
                 base_fee,
-            );
+            )
+        };
+        let maintenance_time = start.elapsed();
+        tracing::debug!(
+            "Pool maintenance took {:?} µs",
+            maintenance_time.as_micros()
+        );
```
dancoombs, line 352: TODO: add a metric for this

dancoombs:
Tests to be added in this ticket https://github.com/alchemyplatform/rundler/issues/845

dancoombs:
Filled a mempool on my laptop on op-sepolia with 1000 UOs, the pool maintenance call took < 5ms.

I'd expect it to be closer to 1-2ms on good/isolated hardware. So I don't think we have a performance problem with the pool maintenance loops. Not going to do anything there for now. If we eventually see issues we have 2 levers to pull: (1) optimize the code (2) move to rayon blocking pool.

dancoombs:
Also closes https://github.com/alchemyplatform/rundler/issues/859

crates/builder/src/bundle_proposer.rs
```diff
@@ -369,32 +393,52 @@ where
...
-                    },
-                ));
-                return None;
+        if !self.settings.chain_spec.da_pre_verification_gas {
+            // Skip PVG check if no da pre-verification gas as this is checked on entry to the mempool.
+            return Some(op);
+        }
+
+        let required_da_gas = if self.settings.da_gas_tracking_enabled && da_block_data.is_some() {
+            let da_block_data = da_block_data.unwrap();
```
0xfourzerofour, line 402: is there some check that makes sure the da_block_data is Some if the tracking is enabled?

crates/pool/src/mempool/pool.rs
```diff
@@ -547,20 +655,41 @@ impl PoolInner {
...
+
     fn uo(&self) -> &UserOperationVariant {
         &self.po.uo
     }
 
     fn mem_size(&self) -> usize {
         std::mem::size_of::<Self>() + self.po.mem_size()
     }
+
+    fn eligible(&self) -> bool {
```
0xfourzerofour, line 682: have you double checked locally that these reads do not happen in the scope of a write? espsecially when its across multiple threads. only asking as its happened before and easy to overlook each deadlock scenario

0xfourzerofour (approved):
Looks good just a couple comments

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -169,130 +161,147 @@ where
...
 {
     async fn on_chain_update(&self, update: &ChainUpdate) {
-        {
-            let deduped_ops = update.deduped_ops();
-            let mined_ops = deduped_ops
-                .mined_ops
-                .iter()
-                .filter(|op| op.entry_point == self.config.entry_point);
+        let deduped_ops = update.deduped_ops();
+        let mined_ops = deduped_ops
```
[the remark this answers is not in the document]
dancoombs: We actually have a ticket for that https://github.com/alchemyplatform/rundler/issues/574

crates/builder/src/bundle_proposer.rs
dancoombs: Yeah, the line above

crates/pool/src/mempool/pool.rs
dancoombs: Don't think theres a deadlock possibility here since the write lock isn't held, its immediately released after mutating.

PR #851, opened 2024-10-15
style: add tomlfmt.toml to unify cargo-sort configuration.

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -


dancoombs:
Why is this needed? Do the defaults not work?

0xfourzerofour (approved):

andysim3d:
discussed offline. the default config it to make sure local and remote config consistent. but on the other hand, we could all use default config and keep the binary consistent.

PR #853, opened 2024-10-15
style: check pr title with commitlint

andysim3d:
[Closes/Fixes] #

## Proposed Changes

test pr title check.

  -
  -
  -

### DIFF
--- a/.github/workflows/pr-lint.yaml
+++ b/.github/workflows/pr-lint.yaml
@@ -1,0 +1,16 @@
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+
+name: PR title check
+jobs:
+  pr-semantic-check:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr-lint.yaml
+++ b/.github/workflows/pr-lint.yaml
@@ -4,6 +4,15 @@ on:
       - opened
       - edited
       - reopened
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+
+permissions:
+  pull-requests: read
 
 name: PR title check
 jobs:


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+name: PR title check
+jobs:
+  pr-semantic-check:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dancoombs, line 16: Should we just do this ourselves? So that we can make sure its the check that runs against the commits? Similar to how its done in AA-SDK.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,16 @@
+on:
```
dancoombs, line 1: Lets call this file `pr-lint.yaml` to make it very clear that its for lints that should run when PRs are edited. Don't want other, more expensive lints, to run alongside.

.github/workflows/lint.yaml
andysim3d: you mean check-in the binary to run the lint check? I find this one on action marketplace and thinking 3 benefits of using it:
- the github action get updates if in the future semantic pull request extended. 
- the binary checks commits semantic is not related to our main logic. we use cocogitto/cocogitto-action@v3 to check it.
- this repo is open sourced, granting read access of github action is low risk.

I have no objections to use our own tool to check semantic commit/pr title and merge 2 checks in one to keep the check consistent.

dancoombs: Got it. My concern here would be if `action-semantic-pull-request` and `cocogitto-action` diverge in their enforcement of conventional commits then we could get commits in that are invalid to `cocogitto-action` and thus cause downstream PRs to fail CI.

If we don't think thats a problem at the moment, we can just address it if it becomes one.

dancoombs (approved):

dancoombs:
Should we sync our config with the one from the action github?

https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#installation

andysim3d:
> Should we sync our config with the one from the action github?
> 
> https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#installation

the example is based on multi-forks repo. we use multi-branch so the trigger and read permission are slightly different.

Issue #858, opened 2024-10-16
[pool] Add unit tests for gas limit efficiency rejections


dancoombs:
**Describe the feature**
This feature is missing tests

PR #877, opened 2024-10-21
chore(pool): add tests for da/gas limit in pool

dancoombs:
Closes: #858, #845


crates/provider/src/alloy/da/local/bedrock.rs
```diff
@@ -147,6 +148,7 @@ where
             .call()
             .await
             .map(|r| r.isFjord)
+            .map_err(|e| error!("failed to check if fjord: {:?}", e))
             .unwrap_or(false)
```
dancoombs, line 152: TODO: should check this, if this ever errors we actually have an `assert!()` in it that causes panics. Not sure if that gets included in our builds or not.

andysim3d (approved):

crates/provider/src/alloy/da/local/bedrock.rs
dancoombs: Will look into this in another PR as this is not related to the tests (just added the log)

andysim3d (approved):

PR #893, opened 2024-10-30
fix(builder): remember last cancel txn fees on underpriced

dancoombs:


andysim3d (approved):

crates/builder/src/transaction_tracker.rs
```diff
@@ -357,31 +360,57 @@ where
...
-        self.update_metrics();
-        Ok(Some(cancel_info.tx_hash))
+                // Store this transaction as pending if last is empty or if it has higher gas fees than last
+                // so that we can continue to increase fees.
+                if self.transactions.last().map_or(true, |t| {
+                    gas_fees.max_fee_per_gas > t.gas_fees.max_fee_per_gas
+                        && gas_fees.max_priority_fee_per_gas > t.gas_fees.max_priority_fee_per_gas
+                }) {
+                    self.transactions.push(PendingTransaction {
+                        tx_hash: B256::ZERO,
```
0xfourzerofour, line 404: are we meant to store a zero hash here?

dancoombs: Yeah, its a marker that the transaction isn't actually pending (since it got rejected)

0xfourzerofour: I assume its just cause it was not succesful

0xfourzerofour (approved):

PR #894, opened 2024-10-31
fix: fix middleware service clone.

andysim3d:
[Closes/Fixes] #

reference: https://docs.rs/tower/latest/tower/trait.Service.html#be-careful-when-cloning-inner-services

## Proposed Changes

  - 
  -
  -

### DIFF
--- a/bin/rundler/chain_specs/opbnb_testnet.toml
+++ b/bin/rundler/chain_specs/opbnb_testnet.toml
@@ -1,0 +1,7 @@
+name = "Opbnb Testnet"
+id = 5611
+
+priority_fee_oracle_type = "USAGE_BASED"
+min_max_priority_fee_per_gas = 30000000000
+bloxroute_enabled = true
+max_transaction_size_bytes = 130000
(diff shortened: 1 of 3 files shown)


0xfourzerofour (approved):

bin/rundler/chain_specs/opbnb_testnet.toml
```diff
@@ -0,0 +1,7 @@
+name = "Opbnb Testnet"
+id = 5611
+
+priority_fee_oracle_type = "USAGE_BASED"
+min_max_priority_fee_per_gas = 30000000000
+bloxroute_enabled = true
```
dancoombs, line 6: Why is this true?

bin/rundler/chain_specs/opbnb_testnet.toml
```diff
@@ -0,0 +1,7 @@
+name = "Opbnb Testnet"
+id = 5611
+
+priority_fee_oracle_type = "USAGE_BASED"
+min_max_priority_fee_per_gas = 30000000000
```
dancoombs, line 5: Just wondering, where did this come from?

andysim3d: this is for local test only. I copied it from  polygon.

dancoombs: Got it, if just used for your local testing we should remove this from the repo.

PR #902, opened 2024-11-05
chore(ci): remove windows build

dancoombs:


0xfourzerofour (commented):
You can also remove the check for windows on the `Move cross-compiled binary` step that changes the extension to .exe

0xfourzerofour (approved):
🚢

Issue #904, opened 2024-11-05
[pool/builder] failed EREP-015 Paymaster opsSeen decrement test


shunsukew:
**Describe the bug**

Rundler fails test `test_paymaster_on_account_failure` in spec tests.
```
EREP-015 A `paymaster` should not have its opsSeen incremented on failure of factory or account
```

**To reproduce**

Run spec test.

**Expected behavior**

Paymaster opsSeen should decrement after the second validation checks if User Operations fail due to factory or account error.

**Logs**

```
+ pdm run pytest -rA -W ignore::DeprecationWarning --url http://localhost:3000/ --entry-point 0x0000000071727De22E5E9d8BAf0edAc6f37da032 --ethereum-node http://localhost:8545/ -k test_paymaster_on_account_failure -vv
======================================================================== test session starts =========================================================================
platform darwin -- Python 3.10.15, pytest-8.3.1, pluggy-1.5.0 -- /Users/shunsuke/alchemy/rundler/test/spec-tests/v0_7/bundler-spec-tests/.venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/shunsuke/alchemy/rundler/test/spec-tests/v0_7/bundler-spec-tests
configfile: pyproject.toml
plugins: web3-6.11.0
collected 185 items / 184 deselected / 1 skipped / 1 selected                                                                                                        

tests/single/reputation/test_erep.py::test_paymaster_on_account_failure FAILED                                                                                 [100%]

...

>       assert post == pre
E       AssertionError: assert Reputation(address='0x0a95d6f554faef4e2f804161037aa93e6e245ead', opsSeen=6, opsIncluded=2, status='ok') == Reputation(address='0x0a95d6f554faef4e2f804161037aa93e6e245ead', opsSeen=5, opsIncluded=2, status='ok')
E         
E         Matching attributes:
E         ['address', 'opsIncluded', 'status']
E         Differing attributes:
E         ['opsSeen']
E         
E         Drill down into differing attribute opsSeen:
E           opsSeen: 6 != 5
```

PR #905, opened 2024-11-05
fix: EREP-015 rule Paymaster opsSeen decrement

shunsukew:
[Closes/Fixes] #904 

## Proposed Changes

  - Paymaster opsSeen decrement when 2nd validation of User Operation fails due to factory or account revert, as defined in ERC4337.


crates/builder/src/bundle_proposer.rs
```diff
@@ -1450,12 +1464,28 @@ impl<UO: UserOperation> ProposalContext<UO> {
                         );
                     }
                 }
+                SimulationViolation::ValidationRevert(ValidationRevert::Operation {
+                    entry_point_reason,
+                    ..
+                }) => {
+                    // [EREP-015] If user operations error derived from factory or account, paymaster opsSeen amendment is required.
+                    let re = Regex::new(r"^AA[21]").unwrap();
```
shunsukew, line 1472: Same as what reference implementation does.
https://github.com/eth-infinitism/bundler/blob/8bc5028caaf2b380c12e057110f9bba950e49b09/packages/bundler/src/modules/BundleManager.ts#L353

dancoombs: Instead of bringing in regex, can we keep the same pattern that we use above? Something like

```paymaster_amendment_required |= matches!(&message[..3], "AA1" | "AA2")```

File: crates/builder/src/bundle_proposer.rs [hunk 2]
```diff
@@ -1517,6 +1547,23 @@ impl<UO: UserOperation> ProposalContext<UO> {
...
     }
+
+    fn add_erep_015_paymaster_amendment(&mut self, entity: Entity) {
+        if !entity.is_paymaster() {
+            warn!(
+                "Attempted to add EREP-015 paymaster amendment for non-paymaster entity: {:?}",
+                entity
+            );
+            return;
+        }
```
dancoombs, line 1558: I think we should use an `assert!` instead of a log here as this indicates a coding error.

crates/builder/src/bundle_proposer.rs
```diff
@@ -1517,6 +1547,23 @@ impl<UO: UserOperation> ProposalContext<UO> {
...
+            entity.address,
+            EntityUpdate {
+                entity,
+                update_type: EntityUpdateType::PaymasterAmendment,
+            },
+        );
+    }
 }
 
 #[cfg(test)]
```
dancoombs, line 1569: Can you add a test for this case?

crates/builder/src/bundle_proposer.rs
dancoombs:
```suggestion
   assert!(entity.is_paymaster(), "Attempted to add EREP-015 paymaster amendment for non-paymaster entity")
```

dancoombs (commented):
Great work! Thank you for the contribution. A few small comments and would love if you could add a unit test for this condition.

shunsukew:
> Great work! Thank you for the contribution. A few small comments and would love if you could add a unit test for this condition.

Thank you for reviewing, please let me address on those. Also, If you have better naming of EntityUpdateType`PaymasterAmendment`, please let me know.

shunsukew:
@dancoombs While writing some tests, realized that `entity_updates` is BTreeMap, and if having multiple EREP-015 applicable cases detected for the same paymaster address in the bundle, opsSeen decremented only 1.
By carrying value to decrement in EmptyUpdate, decrementing opsSeen by correct number.

Because we cannot define enum with custom data type in protocol buffer, I chose the way to have separate `Option<u64>` field for decrementing value in `EntityUpdate` struct.

dancoombs (approved):

dancoombs:
Thank you @shunsukew for the contribution!

Issue #912, opened 2024-11-22
Soneium Minato chainspec default support


shunsukew:
**Describe the feature**
Having Soneium Minato network chainspec as one of binary embedded default options.

PR #913, opened 2024-11-22
feat: chain spec soneium minato network

shunsukew:
Closes #https://github.com/alchemyplatform/rundler/issues/912

## Proposed Changes

  - Add Soneium Minato chain spec as embedded network

### DIFF
--- a/bin/rundler/chain_specs/soneium_minato.toml
+++ b/bin/rundler/chain_specs/soneium_minato.toml
@@ -1,0 +1,12 @@
+name = "Soneium Minato"
+id = 1946
+
+da_pre_verification_gas = true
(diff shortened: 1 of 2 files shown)


dancoombs:
Hey @shunsukew, we're not planning on adding any more L2s to our `chain_specs` folder. The plan at the moment is to only add L1s and L2 frameworks (i.e. optimism and arbitrum). For new L2s they can use optimism or arbitrum as a template and maintain their own chain spec. (We broke this rule with the Base networks and plan on deleting them).

You can pass in the chain spec file via a CLI flag or via ENV variables.

shunsukew:
I see, thank you so much for clarifying. I'm now currently using custom chain spec, I have no issue with it.
Thank you so mcuh!

PR #915, opened 2024-11-24
fix(pool): don't log an error during validation failure in gas check

dancoombs:
PR to main: https://github.com/alchemyplatform/rundler/pull/914


0xfourzerofour (approved):

crates/pool/src/mempool/uo_pool.rs
```diff
@@ -191,7 +191,10 @@ where
                     tracing::error!("Failed to simulate handle op for gas limit efficiency check, failing open: {:?}", e);
                 }
                 Ok(Err(e)) => {
-                    tracing::error!("Failed to simulate handle op for gas limit efficiency check, failing open: {:?}", e);
+                    tracing::debug!(
```
0xfourzerofour, line 194: did we wanna make this a warn? I only say because we do not get debug messaged in prod

dancoombs: No this is fine to leave as debug, if validation fails here then simulation will fail simultaneously and the UO will be rejected.

0xfourzerofour (approved):

PR #917, opened 2024-11-25
fix(builder): handle more tx sender errors

dancoombs:
## Proposed Changes

  - Handle a few more transaction sender errors that occur when the receiving node's mempool is full


dancoombs:
Pick to main: https://github.com/alchemyplatform/rundler/pull/918

0xfourzerofour (approved):
LGTM

PR #920, opened 2024-11-25
fix(builder): fix reth tx sender errors

dancoombs:


dancoombs:
Pick to https://github.com/alchemyplatform/rundler/pull/921

0xfourzerofour (approved):

PR #922, opened 2024-12-02
feat(chain): add a sync error count to reset in edge cases

0xfourzerofour:
## Proposed Changes
  - Add sync error counter to the chain tracker to catch the edge case where the chain will always get stopped at the `ensure!` macro where the current head is much larger than the new head

### DIFF
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -409,6 +409,29 @@ impl<P: EvmProvider> Chain<P> {
         Ok(added_blocks)
     }
 
+    async fn fetch_block_with_retries(&self, block_hash: B256) -> Option<Block> {
+        for attempt in 1..=self.settings.max_sync_retries {
+            match self.provider.get_block(block_hash.into()).await {
+                Ok(Some(block)) => return Some(block),
+                Ok(None) => warn!(
+                    "Block with hash {:?} not found. Retrying... (attempt {}/{})",
+                    block_hash, attempt, self.settings.max_sync_retries
+                ),
+                Err(err) => warn!(
+                    "Error fetching block with hash {:?}: {}. Retrying... (attempt {}/{})",
+                    block_hash, err, attempt, self.settings.max_sync_retries
+                ),
+            }
+            time::sleep(self.settings.poll_interval).await;
+        }
+
+        warn!(
+            "Failed to fetch block with hash {:?} after {} attempts.",
+            block_hash, self.settings.max_sync_retries
+        );
+        None
+    }
+
     async fn load_blocks_back_to_number_no_ops(
         &self,
         head: BlockSummary,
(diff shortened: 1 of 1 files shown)


crates/pool/src/chain.rs
```diff
@@ -240,10 +243,17 @@ impl<P: EvmProvider> Chain<P> {
...
-        ensure!(
-            current_block_number < new_block_number + self.settings.history_size,
+
+        if self.sync_error_count >= SYNC_ERROR_COUNT_MAX {
+            return self.reset_and_initialize(new_head).await;
+        }
+
+        if current_block_number > new_block_number + self.settings.history_size {
+            self.sync_error_count += 1;
+            bail!(
```
dancoombs, line 253: Any reason not to directly do 

```
if self.sync_error_count >= SYNC_ERROR_COUNT_MAX {
            return self.reset_and_initialize(new_head).await;
        }
```

after this increment and remove the code above?

crates/pool/src/chain.rs
```diff
@@ -240,10 +243,17 @@ impl<P: EvmProvider> Chain<P> {
         };
         let current_block_number = current_block.number;
         let new_block_number = new_head.number;
-        ensure!(
-            current_block_number < new_block_number + self.settings.history_size,
+
+        if self.sync_error_count >= SYNC_ERROR_COUNT_MAX {
```
dancoombs, line 247: How does this check interact with the `max_sync_retires` count that is going on in the `wait_for_update` loop? Could we instead just reset if we abandon a sync?

crates/pool/src/chain.rs
0xfourzerofour: I did this so it catches that specific error that we had where it was getting caught in the ensure section for thousands of blocks. But we could just reset after an abandon each time if we want

0xfourzerofour: oh right you mean handle it inside this if statement before the bail. Yeh I think that is fine I can change

crates/pool/src/chain.rs
```diff
@@ -408,16 +442,20 @@ impl<P: EvmProvider> Chain<P> {
...
+                blocks.push_front(BlockSummary::try_from_block_without_ops(
+                    parent,
+                    Some(blocks[0].number - 1),
+                )?);
+            } else {
+                warn!(
+                "Unable to backtrack chain history beyond block number {} due to missing parent block.",
+                blocks[0].number
+            );
+                break;
```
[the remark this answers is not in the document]
0xfourzerofour: this happens within the reset_and_initialize function which is what was causing the loop it would try to reset and then lead to a loop of resets. 

If we were to reset here do you think we should just reset to the head of the valid blocks? like this

```
self.reset_and_initialize(blocks.back())
```

crates/pool/src/chain.rs
0xfourzerofour: reseting more often just means that we need to loop through all the blocks again + call eth_getLogs for each of those blocks to load in the data so it just slows things down a little bit which is why I wanted to limit the resets.

The only places I could see that cause these loops are the ensure thing that I fixed and the block loading on reset

crates/pool/src/chain.rs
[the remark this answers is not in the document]
0xfourzerofour: that is what was causing the error before, if we error out it gets in an infinite loop of trying the same blocks and erroring again. We were getting a loop of the 

`block with parent hash of known block should exist`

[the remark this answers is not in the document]
0xfourzerofour: Yeh I guess it could miss some paymaster tracking info in the step after this function is called. In that case we just return an error like before so we do not miss paymaster ops

dancoombs (approved):

Issue #839, opened 2024-10-10
EIP-7702 Support


dancoombs:
**Describe the feature**
Add support for [EIP-7702 ](https://eips.ethereum.org/EIPS/eip-7702).

Likely set of changes:
* Modify RPC types to accept a 7702 authorization tuple alongside a user operation. The exact shape of this RPC should be agreed upon by other bundler providers.
* Modify preVerificationGas estimation and requirements to correctly charge for a EIP-7702 transaction.
* Use state overrides to set the account code while estimating gas, simulating the UOs, checking the bundle, etc.
* Hold onto the authorization tuple in the mempool alongside the UO.
* Send 7702 transactions in the bundler if any UO contains a authorization tuple.

The reference bundler currently has a branch with 7702 support: https://github.com/eth-infinitism/bundler/tree/AA-453-eip-7702-support

PR #923, opened 2024-12-02
feat: support 7702 txn with userOps 

andysim3d:
## Proposed Changes
it only works for unsafe mode now, other testnet lacking support of `debug_traceCall`. 
  -
  -
  -


andysim3d:
a step of #839

bin/rundler/chain_specs/mekong.toml
```diff
@@ -0,0 +1,2 @@
+name = "mekong"
+id = 7078815900
```
dancoombs, line 2: Remove this file, can be configured via CLI

bin/rundler/chain_specs/odessey.toml
```diff
@@ -0,0 +1,2 @@
+name="odessey"
+id = 911867
```
dancoombs, line 2: Same here, remove

crates/pool/src/server/remote/protos.rs
```diff
@@ -126,6 +155,18 @@ impl TryUoFromProto<UserOperationV07> for v0_7::UserOperation {
...
     ) -> Result<Self, ConversionError> {
+        let authorization_tuple = match &op.authorization_tuple {
+            Some(authorization) => Some(Authorization {
+                chain_id: from_bytes(&authorization.chain_id)?,
+                address: from_bytes(&authorization.address)?,
+                nonce: from_bytes(&authorization.nonce)?,
+                y_parity: from_bytes(&authorization.y_parity)?,
+                r: from_bytes(&authorization.r)?,
+                s: from_bytes(&authorization.s)?,
+            }),
```
dancoombs, line 166: Can you DRY this logic? Looks like its here twice. You can implement a `TryFrom` I think

crates/rpc/src/types/v0_6.rs
```diff
@@ -94,6 +101,7 @@ pub(crate) struct RpcUserOperationOptionalGas {
     max_priority_fee_per_gas: Option<U128>,
     paymaster_and_data: Bytes,
     signature: Bytes,
+    contract_address: Option<Address>,
```
dancoombs, line 104: I think we might need to be more specific than `contract_address` here. What about `delegation_contract_address` or something thats clearly 7702 specific?

crates/rpc/src/types/v0_7.rs
```diff
@@ -154,6 +160,7 @@ pub(crate) struct RpcUserOperationOptionalGas {
     paymaster_post_op_gas_limit: Option<U128>,
     paymaster_data: Option<Bytes>,
     signature: Bytes,
+    contract_address: Option<Address>,
```
dancoombs, line 163: Same comment here on naming

crates/sim/src/estimation/estimate_call_gas.rs
```diff
@@ -107,6 +107,10 @@ where
 
         let callless_op = self.specialization.get_op_with_no_call_gas(op.clone());
 
+        if let Some(authorization_tuple) = op.authorization_tuple().clone() {
+            let contract_address = authorization_tuple.address;
+            apply_7702_overrides(&mut state_override, op.sender(), contract_address);
```
dancoombs, line 112: nit: we typically leave at least som name spacing when importing functions. It makes it clear that the function lives in another file. So here `authoirzation_utils :: apply_7702_overrides `

crates/utils/src/authoirzation_utils.rs
```diff
@@ -0,0 +1,35 @@
+// This file is part of Rundler.
```
dancoombs, line 1: Typo in the file name

crates/utils/src/lib.rs
```diff
@@ -20,6 +20,7 @@
 
 //! Rundler utilities
 
+pub mod authoirzation_utils;
```
dancoombs, line 23: Typo here

crates/types/src/user_operation/v0_7.rs
```diff
@@ -360,6 +369,17 @@ impl UserOperationOptionalGas {
...
         }
+        if self.contract_address.is_some() {
+            builder = builder.authorization_tuple(Some(Authorization {
+                address: self.contract_address.unwrap(),
+                chain_id: chain_spec.id,
+                // fake value for gas estimation.
+                nonce: 0,
+                y_parity: 0,
+                r: U256::from(0),
+                s: U256::from(0),
```
dancoombs, line 380: What is this used for?

crates/types/src/user_operation/v0_6.rs
```diff
@@ -455,6 +475,14 @@ impl UserOperationOptionalGas {
             super::default_if_none_or_equal(self.verification_gas_limit, max_verification_gas, 0);
         let pvg = super::default_if_none_or_equal(self.pre_verification_gas, max_call_gas, 0);
 
+        let authorization_tuple = self.contract_address.map(|address| Authorization {
+            chain_id: chain_spec.id,
+            address,
+            nonce: 0,
+            y_parity: 0,
+            r: U256::from(0),
+            s: U256::from(0),
```
dancoombs, line 484: What is this used for? My understanding is that we don't need the authorization because during estimation we just use state overrides to set the code

crates/task/src/grpc/protos.rs
```diff
@@ -149,3 +167,15 @@ impl ToProtoBytes for u128 {
...
+    fn to_proto_bytes(&self) -> Vec<u8> {
+        self.to_le_bytes().into()
+    }
+}
+
+impl ToProtoBytes for u8 {
+    fn to_proto_bytes(&self) -> Vec<u8> {
+        self.to_le_bytes().into()
+    }
+}
```
dancoombs, line 181: Is there a reason why we need these? Could we instead just use `uint8` and `uint64` in the protobuf?

crates/sim/src/precheck.rs
```diff
@@ -228,7 +228,13 @@ where
             sender_exists,
             ..
         } = async_data;
+        // very hacky, should revisit later.
```
dancoombs, line 231: How so?

crates/sim/src/precheck.rs
```diff
@@ -228,7 +228,13 @@ where
...
+        let is_7702_uo = op.authorization_tuple().is_some();
         let mut violations = ArrayVec::new();
+
+        if is_7702_uo {
+            return violations;
+        }
         if op.factory().is_none() {
```
dancoombs, line 238: We should probably ensure that a 7702 UO doesn't have a factory set since they're not compatible.

crates/sim/src/estimation/v0_7.rs
```diff
@@ -71,6 +71,10 @@ where
     ) -> Result<GasEstimate, GasEstimationError> {
         self.check_provided_limits(&op)?;
 
+        let mut local_override = state_override.clone();
+        if let Some(au) = &op.contract_address {
+            apply_7702_overrides(&mut local_override, op.sender, *au);
```
dancoombs, line 76: Same nit here

crates/sim/src/estimation/v0_6.rs
```diff
@@ -71,6 +71,10 @@ where
         state_override: StateOverride,
     ) -> Result<GasEstimate, GasEstimationError> {
         self.check_provided_limits(&op)?;
+        let mut local_override = state_override.clone();
+        if let Some(au) = &op.contract_address {
+            apply_7702_overrides(&mut local_override, op.sender, *au);
```
dancoombs, line 76: Same nit

crates/sim/src/estimation/estimate_verification_gas.rs
```diff
@@ -73,8 +75,26 @@ where
         max_guess: u128,
         get_op_with_limit: F,
     ) -> Result<u128, GasEstimationError> {
+        let mut local_state_override = state_override.clone();
         let timer = std::time::Instant::now();
         let paymaster_gas_fee = self.settings.verification_estimation_gas_fee;
+        let authorization_list_gas = match op.authorization_tuple() {
```
dancoombs, line 81: I don't think this is the correct place for this. The authorization list setting doesn't happen during the verification phase, so its not subject to the verificationGasLimit.

This gas should be added to the preVerificationGas.

I'd guess that the bundler is losing tokens when its sending 7702 transactions without accounting for PVG this way.

crates/sim/src/estimation/estimate_verification_gas.rs
```diff
@@ -73,8 +75,26 @@ where
...
+            Some(au) => {
+                let prefix: FixedBytes<3> = fixed_bytes!("ef0100");
+                let code: FixedBytes<23> = prefix.concat_const(au.address.into());
+                // TODO(andy): if sca_address is 0x0, should remove code there.
```
dancoombs, line 85: What is this TODO?

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -233,13 +235,18 @@ where
...
             ops_per_aggregator,
             beneficiary,
             gas_limit,
         );
-        let res = self.i_entry_point.provider().call(&tx).await;
+        let res = self
+            .i_entry_point
+            .provider()
+            .call(&tx)
+            .overrides(&override_7702)
```
dancoombs, line 248: Why are we using an override here vs using a 7702 eth call?

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -518,25 +534,49 @@ fn get_handle_ops_call<AP: AlloyProvider<T>, T: Transport + Clone>(
...
-        entry_point
-            .handleOps(ops_per_aggregator.swap_remove(0).userOps, beneficiary)
-            .gas(gas)
-            .into_transaction_request()
+        (
+            entry_point
+                .handleOps(ops_per_aggregator.swap_remove(0).userOps, beneficiary)
+                .gas(gas)
+                .into_transaction_request(),
+            override_7702,
```
dancoombs, line 569: How come this one doesn't have `.with_authorization_list(authorization_list)` but the aggregator version below does?

Also, why do you need the override if you have the authorization list?

crates/provider/src/alloy/entry_point/v0_7.rs
```diff
@@ -11,10 +11,13 @@
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
 
+use std::vec;
```
dancoombs, line 14: Same questions in this file as in the v0.6 file

dancoombs (changes requested):
Great work! I think the biggest concern here is the handling of the 7702 gas should be moved to preVerificationGas. We should also check if the estimateGas RPC field naming conforms with what the 4337 team is implementing in their tests.

crates/rpc/src/types/v0_6.rs
andysim3d: humm I didn't notice their proposal change the name [authorizationContract](https://hackmd.io/WJtBdwvnT5yLq90F7UilIQ). how does it sounds?

crates/types/src/user_operation/v0_6.rs
andysim3d: We pass the authorization, at least the contract address(other fields are 0 by default) here, so estimator knows what 7702 override to apply.

crates/provider/src/alloy/entry_point/v0_6.rs
andysim3d: we can't, or at least shoulnd't use 7702 eath call at gas estimation phase. I assume user don't need to sign authorization tuple during gas estimation phase. without signed authorization tuple, we can't make eth_call with signed authorization tuple.

crates/provider/src/alloy/entry_point/v0_6.rs
andysim3d: oops, good catch, I missed the this one.

crates/task/src/grpc/protos.rs
andysim3d: well protos compiles all to Vec<u8>. impl those traits simplifies the work of serialize & deserialize logic.

andysim3d:
> Great work! I think the biggest concern here is the handling of the 7702 gas should be moved to preVerificationGas. We should also check if the estimateGas RPC field naming conforms with what the 4337 team is implementing in their tests.

for the 4337 - I left a comment to confirm the structure. https://github.com/eth-infinitism/bundler-spec-tests/pull/120#discussion_r1883202383. hope we could resolve the comment asap.

crates/rpc/src/types/v0_6.rs
dancoombs: I like that much better

crates/types/src/user_operation/v0_6.rs
dancoombs: Ah okay, this is a way to convert the user operation from the estimation RPC version to a normal UO for estimation - the signature is not actually needed during estimation. Understood!

crates/task/src/grpc/protos.rs
dancoombs: Does it? Don't we use `uint64` elsewhere in our protos? Would like to stay consistent.

crates/provider/src/alloy/entry_point/v0_6.rs
dancoombs: I don't believe this `call_handle_ops` function is used during estimation. In fact, it can't be because a normal handle ops call would fail if the signature isn't valid, so dummy signatures wouldn't work.

I believe this function is only used by the bundle builder to ensure validity of the final bundle. We should double check exactly where its used. But for `handleOps` its safe to assume that all signatures are valid, including authorization tuples.

dancoombs:
Any idea why compliance is failing?

andysim3d:
> Any idea why compliance is failing?

yes it is because I add the new field but not mark as optional, so some spec test failed. fixing now.

crates/task/src/grpc/protos.rs
andysim3d: Done.

crates/sim/src/estimation/estimate_verification_gas.rs
andysim3d: done

crates/pool/src/server/remote/protos.rs
```diff
@@ -69,11 +83,30 @@ pub trait TryUoFromProto<T>: Sized {
...
+}
 impl TryUoFromProto<UserOperationV06> for v0_6::UserOperation {
     fn try_uo_from_proto(
         op: UserOperationV06,
         chain_spec: &ChainSpec,
     ) -> Result<Self, ConversionError> {
+        let authorization_tuple = op
+            .authorization_tuple
+            .as_ref()
+            .map(|authorization| Authorization::try_from(authorization.clone()).unwrap());
```
dancoombs, line 108: Should this be `.unwrap()` or `?`

crates/pool/src/server/remote/protos.rs
```diff
@@ -126,6 +163,11 @@ impl TryUoFromProto<UserOperationV07> for v0_7::UserOperation {
         op: UserOperationV07,
         chain_spec: &ChainSpec,
     ) -> Result<Self, ConversionError> {
+        let authorization_tuple = op
+            .authorization_tuple
+            .as_ref()
+            .map(|authorization| Authorization::try_from(authorization.clone()).unwrap());
```
dancoombs, line 169: Should this be .unwrap() or ?

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -442,18 +444,27 @@ where
             .pre_verification_da_gas_limit(&self.chain_spec, Some(1))
             .try_into()
             .unwrap_or(u64::MAX);
+        let authorization_tuple = op.authorization_tuple.clone();
+
+        let mut local_state = state_override.clone();
```
dancoombs, line 449: Why is this clone needed? Can we just do `mut state_override` in the function signature?

crates/provider/src/alloy/entry_point/v0_7.rs
dancoombs, line 14: nit: what is this import used for? `vec!` shouldn't need to be imported

crates/task/src/grpc/protos.rs
```diff
@@ -107,7 +107,6 @@ impl FromFixedLengthProtoBytes for u128 {
         u128::from_le_bytes(int_bytes.try_into().unwrap())
     }
 }
-
```
dancoombs, line 110: nit: lets add this back so we don't have a single whitespace change in this file

crates/pool/src/server/remote/protos.rs
andysim3d: this is intended to keep as option<authorization>. if a UO w/o authorization_tuple, it should not fail here.

crates/pool/src/server/remote/protos.rs
andysim3d: same ^

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -297,14 +299,14 @@ where
...
             &self.i_entry_point,
             ops_per_aggregator,
             beneficiary,
             gas_limit,
-        )
-        .max_fee_per_gas(gas_fees.max_fee_per_gas)
-        .max_priority_fee_per_gas(gas_fees.max_priority_fee_per_gas)
+        );
+        tx.max_fee_per_gas(gas_fees.max_fee_per_gas)
+            .max_priority_fee_per_gas(gas_fees.max_priority_fee_per_gas)
```
dancoombs, line 309: nit: can we put this back to the  way it was?

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -436,12 +438,21 @@ where
...
         target_call_data: Bytes,
         block_id: BlockId,
-        state_override: StateOverride,
+        mut state_override: StateOverride,
     ) -> ProviderResult<Result<ExecutionResult, ValidationRevert>> {
         let da_gas: u64 = op
             .pre_verification_da_gas_limit(&self.chain_spec, Some(1))
             .try_into()
             .unwrap_or(u64::MAX);
+        let authorization_tuple = op.authorization_tuple.clone();
```
dancoombs, line 447: Do you need to clone here? Looks like you can do this by reference

crates/provider/src/alloy/entry_point/v0_7.rs
```diff
@@ -311,14 +312,17 @@ where
         block: BlockHashOrNumber,
         gas_price: u128,
     ) -> ProviderResult<(u128, DAGasUOData, DAGasBlockData)> {
-        let data = self
+        let au = user_op.authorization_tuple().clone();
```
dancoombs, line 315: Same question here, does this need to be a clone?

crates/provider/src/alloy/entry_point/v0_7.rs
```diff
@@ -350,6 +354,15 @@ where
         let mut override_ep = StateOverride::default();
         add_simulations_override(&mut override_ep, addr);
 
+        let authorization_tuple = user_op.authorization_tuple.clone();
```
dancoombs, line 357:
```suggestion
        if let Some(authorization) = &user_op.authorization_tuple {
            authorization_utils::apply_7702_overrides(
                &mut override_ep,
                user_op.sender(),
                authorization.address.clone(), // clone just the part you need here
            );
        }
```

crates/provider/src/alloy/entry_point/v0_7.rs
```diff
@@ -422,6 +435,16 @@ where
             .unwrap_or(u64::MAX);
 
         add_simulations_override(&mut state_override, *self.i_entry_point.address());
+
+        let authorization_tuple = op.authorization_tuple.clone();
```
dancoombs, line 439: Same quesetion

crates/sim/src/precheck.rs
```diff
@@ -260,6 +267,13 @@ where
             ..
         } = *async_data;
 
+        let authorization_gas = if op.authorization_tuple().is_some() {
+            alloy_eips::eip7702::constants::PER_AUTH_BASE_COST
+                + alloy_eips::eip7702::constants::PER_EMPTY_ACCOUNT_COST
+        } else {
+            0
+        };
```
dancoombs, line 275: Is there a reason why you did this here and not in the `estimate_pre_verification_gas` function? So we don't need to duplicate this logic in 3 places.

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -297,14 +299,14 @@ where
...
             beneficiary,
             gas_limit,
-        )
-        .max_fee_per_gas(gas_fees.max_fee_per_gas)
-        .max_priority_fee_per_gas(gas_fees.max_priority_fee_per_gas)
+        );
+        tx.max_fee_per_gas(gas_fees.max_fee_per_gas)
+            .max_priority_fee_per_gas(gas_fees.max_priority_fee_per_gas)
     }
 }
```
dancoombs, line 312: `calc_da_gas` below needs the `.with_authorization_list` portion if the UO contains an auth list, to ensure that we're charging for that DA calldata. Worth checking there rest of the codebase to see if there are other places this is missing.

crates/pool/src/server/remote/protos.rs
dancoombs: Got it, so we have a check for this elsewhere? There's no way for a UO that has an authorization tuple couldn't convert that tuple into a protobuf tuple? This `unwrap()` is on the protobuf conversion

crates/sim/src/precheck.rs
dancoombs: Fix LGTM

dancoombs (approved):
🚢 Great work!

PR #926, opened 2024-12-05
fix: impl timeout layer.

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -


crates/provider/src/alloy/mod.rs
```diff
@@ -39,15 +43,19 @@ pub fn new_alloy_evm_provider(rpc_url: &str) -> anyhow::Result<impl EvmProvider
...
-    impl AlloyProvider<RetryBackoffService<AlloyMetricMiddleware<Http<Client>>>> + Clone,
+    impl AlloyProvider<RetryBackoffService<AlloyMetricMiddleware<ProviderTimeout<Http<Client>>>>>
+        + Clone,
 > {
     let url = Url::parse(rpc_url).context("invalid rpc url")?;
     let metric_layer = AlloyMetricLayer::default();
     // TODO: make this configurable: use a large number for CUPS for now
     let retry_layer = alloy_transport::layers::RetryBackoffLayer::new(10, 500, 1_000_000);
+    // add a timeout layer here.
+    let timeout_layer = ProviderTimeoutLayer::new(Duration::from_secs(10));
```
dancoombs, line 54: Is this the correct layer to add the timeout? Doesn't alloy support adding a tower based client? We could use the tower timeout layer.

andysim3d: I tried the tower timeout layer, and that will make the metric layer very mess as they encapsulated the response and error type. 

I can add a todo here to research how to switch back to tower::timeout layer, but on the other side I think this is not the first time we see stalled chains, should we merge and deploy for now?

dancoombs: What about the metric layer will be messy?

crates/provider/src/alloy/metrics.rs
```diff
@@ -126,6 +126,11 @@ where
                             method_logger.record_http(HttpCode::TwoHundreds);
                             method_logger.record_rpc(RpcCode::Success);
                         }
+                        // for timeout error
+                        alloy_json_rpc::RpcError::LocalUsageError(_) => {
+                            method_logger.record_http(HttpCode::FourHundreds);
```
dancoombs, line 131: I think this is still a 5xx because its timing out the server. We'd want this separate from the 4xx's which are errors that need to be fixed by the client.

dancoombs:
Have you been able to test this? Would love to simulate a timeout and watch the error propagate

crates/provider/src/alloy/mod.rs
andysim3d: because the tower::timeout's trait bound of request and response are wider than metric layer, and we need metric layer to extract request and response info. use tower::timeout will cause the compiler complain metrics layer not satisfy trait bound somehow.

dancoombs: Okay, lets add a ticket to revisit maybe in v0.6

bin/rundler/Cargo.toml
```diff
@@ -41,6 +41,7 @@ reth-tasks.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 sscanf = "0.4.2"
+tiny_http.workspace = true
```
dancoombs, line 44: Is this needed here?

crates/provider/src/alloy/mod.rs
```diff
@@ -28,27 +31,79 @@ pub use da::new_alloy_da_gas_oracle;
...
+        }
+    }
+    #[tokio::test]
+    async fn test_timeout() {
+        thread::spawn(move || {
+            setup();
+        });
+        {
+            // Wait 11 seconds and get result
+            let provider = new_alloy_provider("http://localhost:8000", 11)
```
dancoombs, line 96: You may need to expand the tolerances to more than a second to avoid flaky tests

dancoombs (approved):

Issue #579, opened 2024-02-01
Add CORS config to the cli


0xfourzerofour:
Some rundler users may want to have a custom CORS configuration which we should be able to configure via the cli and pass that logic into the server

PR #927, opened 2024-12-09
feat(Rundler): Add cors to the server

Blu-J:
[Closes/Fixes] #579 

## Proposed Changes

* Implementing the cors that is already written for tower-http.

## Proof it works

This is before the cors implementation

image

Then adding the `RPC_CORSDOMAIN=*` fix we get

image

This is from 127.0.0.1:3000, notice no cors in response
image

This is from localhost:3000, notice again cors
image

Then we set the `RPC_CORSDOMAIN=http://127.0.0.1:3000`
image

Then we set the `RPC_CORSDOMAIN=http://127.0.0.1:3000,http://localhost:3000` Testing list of domains
image

And when we add a wildcard at the end `RPC_CORSDOMAIN=http://127.0.0.1:3000,http://localhost:3000,*`
we throw on reading the env
image

And with corsdomain as localhost
```js
fetch("http://127.0.0.1:3000/", {
            method: 'POST',
            mode: 'cors',
            headers: { 'Content-Type': 'application/json' },
            body: JSON.stringify({
                jsonrpc: '2.0',
                method: 'system_health',
                id: 1
            })
        }).then(res => {
            console.log("Response:", res);
            return res.text()
        }).then(body => {
            console.log("Response Body:", body)
        });
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,7 @@ tonic-health = "0.12.3"
 tonic-reflection = "0.12.3"
 tonic-types = "0.12.3"
 tower = { version = "0.4.13", features = ["timeout"] }
+tower-http = { version = "0.6.2", features = ["cors"] }
 tracing = "0.1.40"
 strum = { version = "0.26.3", features = ["derive"] }
 url = "2.5.2"
--- a/crates/rpc/src/task.rs
+++ b/crates/rpc/src/task.rs
@@ -15,6 +15,7 @@ use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use anyhow::Context;
 use futures_util::FutureExt;
+use http::{header::CONTENT_TYPE, HeaderValue};
 use jsonrpsee::{
     server::{middleware::http::ProxyGetRequestLayer, RpcServiceBuilder, ServerBuilder},
     RpcModule,
@@ -75,6 +76,8 @@ pub struct Args {
     pub entry_point_v0_6_enabled: bool,
     /// Whether to enable entry point v0.7.
     pub entry_point_v0_7_enabled: bool,
+    /// What domains to use in the corsdomain
+    pub corsdomain: Option<Vec<HeaderValue>>,
 }
 
 /// JSON-RPC server task.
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -87,6 +87,7 @@ tonic-health = "0.12.3"
 tonic-reflection = "0.12.3"
 tonic-types = "0.12.3"
 tower = { version = "0.4.13", features = ["timeout"] }
+tower-http = { version = "0.6.2", features = ["cors", "timeout"] }
```
andysim3d, line 90: just curious, do we need timeout of tower-http?

Blu-J: Good catch, I was just matching the features to the tower. And checked that w/o the timeout it still is good to go. Looks like the tower-http has a timeout layer, which we won't use because of tower already having it.

crates/rpc/src/task.rs
```diff
@@ -189,6 +192,26 @@ where
 
         // Set up health check endpoint via GET /health registers the jsonrpc handler
         let http_middleware = tower::ServiceBuilder::new()
+            .option_layer(self.args.corsdomain.map(|layers| {
+                use tower_http::cors::{AllowOrigin, Any};
+                // In the case where we pass '*', I want to be able to test the any domain.
+                // but without this change the list Origins will reject if there is a wildcard present.
+                // So in the case that there is just '*' passed in the args we will treat it like any
+                const WILDCARD: HeaderValue = HeaderValue::from_static("*");
+                let layers: AllowOrigin = if layers.contains(&WILDCARD) && layers.len() == 1 {
```
Blu-J, line 201: 🤷‍♀️ The reason why is that if our value for the env is something like 
image

It fails because the list doesn't allow the multi card. 
I would like to be able to use * because I haven't found out how to test locally with cors. 
So, in this test I'm testing in the url 127.0.0.1:4000  (A route that isn't our server)
And with the start we get 
image

but I have tried other origins like 127.0.0.1:4000 or 127.0.0.1:3000
And I always get this
image

Issue #822, opened 2024-10-02
[alloy] Update to alloy v0.4


dancoombs:
**Describe the feature**
This may require a repo wide update of u128 -> u64 for gas and fee values.

https://github.com/alloy-rs/alloy/releases/tag/v0.4.0

Of note:
https://github.com/alloy-rs/alloy/issues/1240

PR #930, opened 2024-12-10
feat: bump alloy to 0.7.3.

andysim3d:
[Closes/Fixes] #822 

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/rpc/src/eth/events/common.rs
+++ b/crates/rpc/src/eth/events/common.rs
@@ -13,6 +13,7 @@
 
 use std::{collections::VecDeque, marker::PhantomData};
 
+use alloy_consensus::Transaction;
 use alloy_primitives::{Address, Bytes, B256, U256};
 use alloy_sol_types::SolEvent;
 use anyhow::Context;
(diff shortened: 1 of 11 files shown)


crates/rpc/src/eth/events/common.rs
```diff
@@ -86,13 +86,44 @@ where
         if tx.block_hash.is_none() && tx.block_number.is_none() {
             return Ok(None);
         }
-        let to = tx
-            .to
-            .context("tx.to should be present on transaction containing user operation event")?;
+        let to_kind = match &tx.inner {
```
0xfourzerofour, line 89: we already spoke in person but just putting the comment here for other team members

these can be reduced into common matches by using an or match

andysim3d: fixed.

dancoombs:
Is there a reason why this is towards `feat/v0.5`? I thought we were using `main` for v0.5 development.

dancoombs (approved):

Issue #925, opened 2024-12-05
[builder] Handle standardized `eth_sendRawTransactionConditional` error codes 


dancoombs:
**Describe the feature**
We currently string match, we should check error codes instead. Codes are here https://eips.ethereum.org/EIPS/eip-7796

PR #931, opened 2024-12-10
feat(Rundler): Handle standardized eth_sendRawTransactionConditional error codes

niveda-krish:
[Closes/Fixes] # https://github.com/alchemyplatform/rundler/issues/925

## Proposed Changes

  - Use error codes -32003 and -32005 when condition is not met on eth_sendRawTransactionConditional https://eips.ethereum.org/EIPS/eip-7796 (instead of string matching on json rpc error message)
  - Tests
  
image
image

### DIFF
--- a/crates/builder/src/sender/mod.rs
+++ b/crates/builder/src/sender/mod.rs
@@ -240,7 +240,7 @@ impl From<ProviderError> for TxSenderError {
             ProviderError::RPC(e) => {
                 if let Some(e) = e.as_error_resp() {
                     // Client impls use different error codes, just match on the message
-                    if let Some(e) = parse_known_call_execution_failed(&e.message, &e.code) {
+                    if let Some(e) = parse_known_call_execution_failed(&e.message, e.code) {
                         e
                     } else {
                         TxSenderError::Other(value.into())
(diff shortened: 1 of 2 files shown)


0xfourzerofour:
Looks good just gotta update the PR title use a valid conventional commit message

crates/builder/src/sender/mod.rs
```diff
@@ -257,9 +257,15 @@ impl From<ProviderError> for TxSenderError {
 // Geth: https://github.com/ethereum/go-ethereum/blob/23800122b37695be50565f8221858a16ce1763db/core/txpool/errors.go#L31
 // Reth: https://github.com/paradigmxyz/reth/blob/8e4a917ec1aa70b3779083454ff2d5ecf6b44168/crates/rpc/rpc-eth-types/src/error/mod.rs#L624
 // Erigon: https://github.com/erigontech/erigon/blob/96fabf3fd1a4ddce26b845ffe2b6cfb50d5b4b2d/txnprovider/txpool/txpoolcfg/txpoolcfg.go#L124
-fn parse_known_call_execution_failed(e: &str) -> Option<TxSenderError> {
+fn parse_known_call_execution_failed(message: &str, code: &i64) -> Option<TxSenderError> {
```
andysim3d, line 260: nit: should pass i64 instead of &i64 here. https://stackoverflow.com/questions/77671849/when-and-why-its-better-to-pass-simple-types-by-value-or-reference-in-rust

crates/builder/src/sender/mod.rs
```diff
@@ -240,7 +240,7 @@ impl From<ProviderError> for TxSenderError {
             ProviderError::RPC(e) => {
```
andysim3d, line 240: super nit: could we make this code less nested?

niveda-krish: Yeah we could make this code look a neater by just using if statements everywhere

andysim3d (approved):

PR #933, opened 2024-12-11
chore: bump reth deps

0xOsiris:
[Closes/Fixes] #

## Proposed Changes

This PR bumps the `reth-tasks` dependency in `rundler` to `v1.1.4`. This makes `rundler` crates that have `reth-tasks` as a dependency compatible with projects depending on the latest version of Reth.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4628,37 +4628,26 @@ dependencies = [
 
 [[package]]
 name = "reth-metrics"
-version = "1.0.7"
-source = "git+https://github.com/paradigmxyz/reth.git?tag=v1.0.7#75b7172cf77eb4fd65fe1a6924f75066fb09fcd1"
+version = "1.1.4"
+source = "git+https://github.com/paradigmxyz/reth.git?tag=v1.1.4#15fac0873e91ea29ab2e605bfba17bedcd7a6084"
 dependencies = [
- "metrics 0.23.0",
- "reth-metrics-derive",
-]
-
-[[package]]
-name = "reth-metrics-derive"
-version = "1.0.7"
-source = "git+https://github.com/paradigmxyz/reth.git?tag=v1.0.7#75b7172cf77eb4fd65fe1a6924f75066fb09fcd1"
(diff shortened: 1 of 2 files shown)


dancoombs:
@0xOsiris is the large diff here because of a `cargo update`? Were there further conflicts in when trying to upgrade reth-tasks?

0xOsiris:
> @0xOsiris is the large diff here because of a `cargo update`? Were there further conflicts in when trying to upgrade reth-tasks?

Yes, `cargo build` initially failed but a `cargo update` patched the lock file

dancoombs:
Looks like the `cargo update` pulled in some changes that broke our lints.

It seems alloy has deprecated some fields, we're actively updating alloy in another PR, so this should get fixed once that merges https://github.com/alchemyplatform/rundler/pull/930.

For the Cargo Deny failure, its likely we just have to add `Unicode-3.0` to our allowlist. Let me check internally.

dancoombs:
Update here, we're still figuring out what we need to do (if anything) to change our license to incorporate these denied crates. Trying to get this done ASAP.

dancoombs:
This should clear up the deny issue, also includes the `cargo update` so you can just add the reth bump https://github.com/alchemyplatform/rundler/pull/942

test/spec-tests/v0_6/bundler-spec-tests
0xOsiris, line 1: Not sure how this happened 😅

0xOsiris:
> This should clear up the deny issue, also includes the `cargo update` so you can just add the reth bump #942

Awesome, just merged upstream!

dancoombs (changes requested):
Looks like you have changes to the git submodules. You should revert that change back to head and then run `git update submodules` to get your local updated and this diff should go away.

(I hate submodules and we're looking at ways to remove these, but they're currently very useful for testing)

dancoombs (approved):
Thanks for dealing with all the back-and-forth here!

PR #934, opened 2024-12-13
fix: reset metrics when builder re-initialized.

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -331,6 +331,7 @@ where
                 error!("Bundle send error {error:?}");
                 self.metrics.bundle_txns_failed.increment(1);
                 state.bundle_error(error);
+                state.transaction_tracker.reset().await;
             }
         }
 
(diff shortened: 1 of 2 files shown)


crates/builder/src/bundle_sender.rs
```diff
@@ -331,6 +331,7 @@ where
                 error!("Bundle send error {error:?}");
                 self.metrics.bundle_txns_failed.increment(1);
                 state.bundle_error(error);
+                state.transaction_tracker.reset().await;
```
dancoombs, line 334: Good catch. Should we put this in the `bundle_error` function so we don't forget it? Or do you think it makes more sense here?

dancoombs (approved):

dancoombs:
Do we want to backport this to v0.4?

0xfourzerofour (approved):

crates/builder/src/bundle_sender.rs
andysim3d: I tried but bundler_error is not an async function so I can't apply there.

andysim3d:
> Do we want to backport this to v0.4?

will do!

PR #937, opened 2024-12-16
fix: reset metrics when builder re-initialized. (#934)

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/builder/src/transaction_tracker.rs
+++ b/crates/builder/src/transaction_tracker.rs
@@ -295,6 +295,8 @@ where
         );
         let sent_tx = self.sender.send_transaction(tx, expected_storage).await;
 
+        self.update_metrics();
+
         match sent_tx {
             Ok(sent_tx) => {
                 info!(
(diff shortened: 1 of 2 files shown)


crates/builder/src/transaction_tracker.rs
```diff
@@ -295,6 +295,8 @@ where
         );
         let sent_tx = self.sender.send_transaction(tx, expected_storage).await;
 
+        self.update_metrics();
```
dancoombs, line 298: Is this needed? It looks like it gets called in both of the branches below

dancoombs: Oh jk, its missing a branch. Do we rely on logic from those branches? Would love to simplify

andysim3d: I prefer to merge now and simplify later.

dancoombs (approved):

PR #947, opened 2024-12-20
fix: ignore flaky tests

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/provider/src/alloy/mod.rs
+++ b/crates/provider/src/alloy/mod.rs
@@ -78,14 +78,15 @@ mod tests {
 
     use crate::new_alloy_provider;
     fn setup() {
-        let server = Server::http("0.0.0.0:8000").unwrap();
+        let server = Server::http("0.0.0.0:9009").unwrap();
         for request in server.incoming_requests() {
-            sleep(Duration::from_secs(10));
+            sleep(Duration::from_secs(5));
             let _ = request.respond(Response::from_string(
                 "{\"jsonrpc\": \"2.0\",	\"id\": 1,	\"result\": \"0x146b6d7\"}",
             ));
         }
     }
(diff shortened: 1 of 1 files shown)


alex-miao (approved):

crates/provider/src/alloy/mod.rs
```diff
@@ -78,29 +78,30 @@ mod tests {
 
     use crate::new_alloy_provider;
     fn setup() {
-        let server = Server::http("0.0.0.0:8000").unwrap();
+        let server = Server::http("0.0.0.0:9009").unwrap();
```
dancoombs, line 81: I'm concerned that this will still be flaky in case something is using this port. Is there a way to let the OS assign an open port and retrieve it after? Sometimes this is done by setting the port to `:0`

dancoombs: You may be able to use that and then do [.server_addr](https://docs.rs/tiny_http/latest/tiny_http/struct.Server.html#method.server_addr) to get the port it was bound to by the OS

dancoombs: You may need to use [bind](https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.bind) and then [from_listener](https://docs.rs/tiny_http/latest/tiny_http/struct.Server.html#method.from_listener) to create the server.

andysim3d: I mark the test to ignore. it should only run on local box.

dancoombs: Okay, we should eventually fix this

dancoombs (approved):

PR #951, opened 2024-12-23
fix(pool): fix pool max ops per sender during replacement

dancoombs:
Fixes: #958

## Proposed Changes

  - Consider replacements during max ops per sender enforcement
  - A few small tweaks to reputation logic

### DIFF
--- a/crates/pool/src/mempool/uo_pool.rs
+++ b/crates/pool/src/mempool/uo_pool.rs
@@ -292,7 +292,7 @@ where
 
             if let Some(po) = pool_op {
                 for entity_addr in po.entities().map(|e| e.address).unique() {
-                    self.reputation.remove_included(entity_addr);
+                    self.reputation.dec_included(entity_addr);
                 }
 
                 unmined_op_count += 1;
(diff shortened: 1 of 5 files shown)


crates/pool/src/mempool/uo_pool.rs
```diff
@@ -521,6 +518,8 @@ where
         // Check if op is already known or replacing another, and if so, ensure its fees are high enough
         // do this before simulation to save resources
         let replacement = self.state.read().pool.check_replacement(&op)?;
+        let to_replace = replacement.and_then(|r| self.state.read().pool.get_operation_by_hash(r));
```
andysim3d, line 521: replacement seems not used after declared. could shorten to 
```let to_replace = self.state.read().pool.check_replacement(&op)?.and_then(|r| self.state.read().pool.get_operation_by_hash(r));```

File: crates/pool/src/mempool/uo_pool.rs [hunk 2]
```diff
@@ -599,9 +599,16 @@ where
...
             for entity in pool_op
                 .unstaked_entities()
+                .unique()
                 .filter(|e| e.address != pool_op.entity_infos.sender.address())
             {
-                let ops_allowed = self.reputation.get_ops_allowed(entity.address);
+                let mut ops_allowed = self.reputation.get_ops_allowed(entity.address);
+                if let Some(to_replace) = &to_replace {
+                    if to_replace.entities().contains(&entity) {
+                        ops_allowed += 1;
```
andysim3d, line 608: not sure if I followed here: if op is a replacement, then the different should only be fees. do we want count the ops_allowed for the 'same' ops?

crates/pool/src/mempool/uo_pool.rs
dancoombs: Refactored this to only read the state once, don't like the double locking here

crates/pool/src/mempool/uo_pool.rs
dancoombs: > different should only be fees
Not true, a replacement is any new submission by the same sender/nonce pair. The rest of the UO can be different.

andysim3d (approved):

andysim3d (approved):

crates/pool/src/mempool/uo_pool.rs
dancoombs: Looking at this again, this really could have been a `continue` instead of adding to `ops_allowed`, would have been simpler

PR #966, opened 2025-01-03
feat(builder): log and metric for reverted mined bundle transactions

dancoombs:
## Proposed Changes

  - Add a metric to track reverted transactions

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -353,11 +353,19 @@ where
                     gas_used,
                     tx_hash,
                     nonce,
+                    is_success,
                     ..
                 } => {
-                    info!("Bundle transaction mined");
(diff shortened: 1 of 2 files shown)


andysim3d (approved):

crates/builder/src/bundle_sender.rs
```diff
@@ -353,11 +353,19 @@ where
                     gas_used,
                     tx_hash,
                     nonce,
+                    is_success,
                     ..
                 } => {
-                    info!("Bundle transaction mined");
+                    info!("Bundle transaction mined: block number {block_number}, attempt number {attempt_number}, gas limit {gas_limit:?}, gas used {gas_used:?}, tx hash {tx_hash}, nonce {nonce}, success {is_success}");
```
Blu-J, line 359: 👍 I like the information added for debugging
Usually I handled this with https://docs.rs/tracing/latest/tracing/attr.instrument.html as well, because certain errors like the one down below on the is_success, may not be tracked. In that case I would do probably still keep info because I like this pattern, especially with a unique string to look for. I might have added `#[instrument(skip_all, fields(id = tracker_update.id()))]` to decorate the fn. Also tracing works well with other tools, and it would be good to learn what patterns we use.

PR #967, opened 2025-01-06
feat: add maximum number of expected storage slots param

### LINKED ISSUE #924: [builder] Limit storage slots to check in `eth_sendRawTransactionConditional`
**Describe the feature**
Both Optimism and Arbitrum limit the number of storage slots or known accounts (respectively) that you can send to `eth_sendRawTransactionConditional`. Rundler needs to adhere to this limit by limiting the size of a bundle when it goes over this number of slots. We should, however, try to make the largest bundle possible.

dancoombs:
Closes #924 

## Proposed Changes

  - Limit the number of expected storage slots submitted in a bundle
  - Block UOs from the mempool that use more than the maximum number of expected storage slots

### DIFF
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -361,6 +361,7 @@ impl BuilderArgs {
             remote_address,
             da_gas_tracking_enabled,
             provider_client_timeout_seconds,
+            max_expected_storage_slots: common.max_expected_storage_slots.unwrap_or(usize::MAX),
         })
     }
 
--- a/crates/pool/proto/op_pool/op_pool.proto
+++ b/crates/pool/proto/op_pool/op_pool.proto
@@ -532,6 +532,7 @@ message MempoolError {
     OperationDropTooSoon operation_drop_too_soon = 16;
     PreOpGasLimitEfficiencyTooLow pre_op_gas_limit_efficiency_too_low = 17;
     ExecutionGasLimitEfficiencyTooLow execution_gas_limit_efficiency_too_low = 18;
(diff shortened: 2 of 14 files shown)


bin/rundler/src/cli/builder.rs
```diff
@@ -361,6 +361,7 @@ impl BuilderArgs {
             remote_address,
             da_gas_tracking_enabled,
             provider_client_timeout_seconds,
+            max_expected_storage_slots: common.max_expected_storage_slots.unwrap_or(usize::MAX),
```
0xfourzerofour, line 364: Have we had an issue with this before?

Just asking cause if we are setting the default to `18_446_744_073_709_551_615` that is a pretty high default

dancoombs: We haven't but we know some implementations, like optimism, do have limits. So we'd set this on a case by case basis. Default is off.

crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -594,6 +595,11 @@ message ExecutionGasLimitEfficiencyTooLow {
   float actual = 2;
 }
 
+message TooManyExpectedStorageSlots {
```
0xfourzerofour, line 598: nit: the naming convention here is a little messy. For example there are PreOpGasLimitEfficiencyTooLow, ExecutionGasLimitEfficiencyTooLow. If we wanted to keep this naming consistent it should probably be like ExpectedStorageSlotsTooHigh

0xfourzerofour (approved):

PR #972, opened 2025-01-09
fix: change authorization tuple use hexstirng

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/rpc/src/types/rpc_authorization.rs
+++ b/crates/rpc/src/types/rpc_authorization.rs
@@ -1,0 +1,47 @@
+use alloy_primitives::{Address, U256, U64, U8};
+use rundler_types::authorization::Authorization;
+use serde::{Deserialize, Serialize};
+
+/// authorization tuple for 7702 txn support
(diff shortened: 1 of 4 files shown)


crates/rpc/src/types/rpc_authorization.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// signed authorizzation tuple.
+    pub y_parity: U8,
+    /// signed authorizzation tuple.
+    pub r: U256,
+    /// signed authorizzation tuple.
+    pub s: U256,
+}
+
+impl From<RpcAuthorization> for Authorization {
+    fn from(val: RpcAuthorization) -> Self {
```
0xfourzerofour, line 24: can you implement try_from instead so that there are no panics at the RPC layer

andysim3d: Done

0xfourzerofour (approved):
LGTM

PR #977, opened 2025-01-22
fix(pool): fix execution gas eff check for v0.6 with paymasters

### LINKED ISSUE #976: [pool] Issue with gas efficiency check for v0.6 using a paymaster
**Describe the bug**
The check incorporates a post op estimated cost of 2*VGL, even if a post op isn't used, if VGL >> CGL, and a post op isn't used at all, then gas eff will be low.

**To reproduce**
Send a UO with the properties above and a gas limit eff of 15%, typically triggers and rejects the UO.

**Expected behavior**
This shouldn't reject the UO, for v0.6 we shouldn't include the post op gas at all in this calculation.

dancoombs:
Fixes #976

### DIFF
--- a/crates/pool/src/mempool/uo_pool.rs
+++ b/crates/pool/src/mempool/uo_pool.rs
@@ -168,7 +168,12 @@ where
                 return Ok(());
             }
 
-            let execution_gas_limit = op.execution_gas_limit();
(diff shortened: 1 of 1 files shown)


andysim3d (approved):

0xfourzerofour (approved):
LGTM

Issue #981, opened 2025-01-29
[builder/rpc] Bundle proxy configuration


dancoombs:
**Describe the feature**
Add the ability for the builder to conditionally submit through a proxy contract. Initially only support proxies that have the same interface as the entry point. This should initially condition on an aggregator address.

Additionally, the RPC read calls should understand proxies (with the same interface) for constructing UOs from transactions.

PR #986, opened 2025-01-31
feat(builder): add support for a bundle submission proxy

dancoombs:
Closes: #981


bin/rundler/src/cli/builder.rs
```diff
@@ -272,36 +267,60 @@ impl BuilderArgs {
...
+                    .map(|ep| ep.builders())
+                    .expect("should find entry point v0.7 builders")
+            } else {
+                builder_settings_from_cli(
+                    common.builder_index_offset_v0_7,
+                    common.num_builders_v0_7,
+                )
+            };
+
+            tracing::info!("entry point v0.7 builders: {:?}", builders);
```
dancoombs, line 314: remove

bin/rundler/src/cli/builder.rs
```diff
@@ -272,36 +267,58 @@ impl BuilderArgs {
...
         let rpc_url = common.node_http.clone().context("must provide node_http")?;
 
-        let mempool_configs = match &common.mempool_config_path {
-            Some(path) => get_json_config::<MempoolConfigs>(path)
-                .await
-                .with_context(|| format!("should load mempool configurations from {path}"))?,
-            None => MempoolConfigs::default(),
-        };
-
+        let mempool_configs = mempool_configs.unwrap_or_default();
```
0xfourzerofour, line 270: the default impl will store an empty address (assume Address::zero) + allowlist. will this mess with the current functionality if use the default None?

dancoombs: `MempoolConfigs::default()` is just an empty hashmap, when the simulator does mempool matching this just means that any violations would get rejected

0xfourzerofour (approved):
🚢

PR #997, opened 2025-02-06
fix: amend rpc name to eip7702auth

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/rpc/src/types/v0_6.rs
+++ b/crates/rpc/src/types/v0_6.rs
@@ -38,7 +38,7 @@ pub(crate) struct RpcUserOperation {
     max_priority_fee_per_gas: U128,
     paymaster_and_data: Bytes,
     signature: Bytes,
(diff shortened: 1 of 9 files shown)


crates/rpc/src/types/v0_6.rs
```diff
@@ -38,7 +38,7 @@ pub(crate) struct RpcUserOperation {
     max_priority_fee_per_gas: U128,
     paymaster_and_data: Bytes,
     signature: Bytes,
-    eip7702_auth: Option<RpcEip7702Auth>,
+    eip7702auth: Option<RpcEip7702Auth>,
```
dancoombs, line 41: Could you just do this with `serde(rename = "eip7702Auth")` instead of changing this everywhere?

andysim3d:
after sync with EF, eip7702Auth is the correct annotation.

PR #1013, opened 2025-02-12
feat(builder): introduce submission proxy logic

### LINKED ISSUE #1011: [builder] Submission proxy specific logic
**Describe the feature**
The builder needs to be able to run logic based on a type of submission proxy. For example:

- For the PBH entry point submission proxy we need to be able to catch reverts and remove UOs that caused the reverts either due to invalid nullifiers or failed proof checks.

We can follow a similar pattern here as the signature aggregators: add a registry to chain spec and create implementations in the `bin` folder. The proxy impl can live alongside the pbh aggregator.

dancoombs:
Closes #1011 

We needed a way to process proxy specific logic. Specifically to handle reverts during bundle simulation that require removing UOs from the mempool. This same interface could be used in the future to do things like bundle compression.


crates/aggregators/pbh/src/proxy.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn address(&self) -> Address {
+        self.address
+    }
+
+    fn process_revert(
+        &self,
+        _revert_data: Bytes,
+        _ops: UserOpsPerAggregator<UserOperationVariant>,
+    ) -> Vec<U256> {
+        vec![]
```
dancoombs, line 33: TODO in follow up PR

crates/builder/src/bundle_proposer.rs
```diff
@@ -158,7 +159,7 @@ pub(crate) struct Settings {
     pub(crate) priority_fee_mode: PriorityFeeMode,
     pub(crate) da_gas_tracking_enabled: bool,
     pub(crate) max_expected_storage_slots: usize,
-    pub(crate) submitter_proxy: Option<Address>,
+    pub(crate) submission_proxy: Option<Arc<dyn SubmissionProxy>>,
```
dancoombs, line 162: TODO in follow up PR: pass any reverts through the submission proxy and add tests

dancoombs: https://github.com/alchemyplatform/rundler/pull/1014

bin/rundler/src/cli/builder.rs
```diff
@@ -455,19 +460,34 @@ pub(crate) struct BuilderConfig {
     index_offset: Option<u64>,
     // Submitter proxy to use for builders
     proxy: Option<Address>,
+    // Type of proxy to use for builders
+    proxy_type: Option<String>,
```
0xfourzerofour, line 464: should proxy type be a deserializable enum instead of a string in case we are going to add more proxy types?

0xfourzerofour (dismissed):
looks good just a comment on the proxy type field

andysim3d (dismissed):

bin/rundler/src/cli/builder.rs
dancoombs: Good idea, added

0xfourzerofour (approved):

PR #1014, opened 2025-02-13
feat(builder): process reverts through proxies, removing ops

dancoombs:
## Proposed Changes

  - Pass non-entrypoint revert data to proxies, if they exist
  - Remove UOs from the pool if notified by proxy
  - Else, remove all UOs from the pool if the fault is not attributable to avoid infinite failure loops

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -756,6 +756,27 @@ where
         None
     }
 
+    async fn reject_bundle(&self, context: &mut ProposalContext<<Self as BundleProposer>::UO>) {
+        context.reject_all();
+    }
+
+    async fn reject_hash(
+        &self,
+        context: &mut ProposalContext<<Self as BundleProposer>::UO>,
+        hash: B256,
+    ) -> bool {
+        let Some(index) = context.get_op_index(
+            hash,
+            *self.ep_providers.entry_point().address(),
+            self.settings.chain_spec.id,
+        ) else {
+            return false;
+        };
+
+        self.reject_index(context, index, false).await;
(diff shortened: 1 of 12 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -756,6 +756,27 @@ where
         None
     }
 
+    async fn reject_bundle(&self, context: &mut ProposalContext<<Self as BundleProposer>::UO>) {
+        context.reject_all();
```
andysim3d, line 760: can't left comment at L787 so left here: 

it looks like each time we reject an uo in bundle, we immediately re-compute  signature aggregator will recompute again. if we reject_all of a size N bundle, that will recompute N times for nothing. should we add a flag in reject_index to avoid recompute aggregated signature?

andysim3d (dismissed):
left a minor comment. otherwise LGTM

crates/builder/src/bundle_proposer.rs
dancoombs: Some confusing naming here. There are 2 `reject_index` functions.

`BundleProposerImpl::reject_index` -> calls `ProposalContext::reject_index` and then computes the signature
`ProposalContext::reject_index` -> rejects the index from the bundle, does not compute signature

`reject_bundle` calls `ProposalContext::reject_all` which uses `ProposalContext::reject_index` and does not recompute the signature

andysim3d: ah got you.

andysim3d (approved):

0xfourzerofour (approved):

PR #1018, opened 2025-02-14
feat: log failed eth call

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/provider/src/alloy/entry_point/v0_6.rs
+++ b/crates/provider/src/alloy/entry_point/v0_6.rs
@@ -256,12 +256,12 @@ where
                 if let Some(revert) = resp.as_revert_data() {
                     Ok(Self::decode_handle_ops_revert(&resp.message, &revert))
                 } else {
-                    tracing::debug!("handle_ops failed with request: {:?}", tx);
+                    tracing::error!("handle_ops failed with request: {:?}", tx);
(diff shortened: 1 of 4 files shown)


crates/sim/src/estimation/estimate_verification_gas.rs
```diff
@@ -125,6 +125,11 @@ where
                 ))?;
             }
         } else if let Some(revert) = E::decode_simulate_handle_ops_revert(&gas_used.result)?.err() {
+            tracing::debug!(
```
dancoombs, line 128: This is going to be logged for every user operation that we validate gas for. I think its good to keep this as `debug` - but we probably won't want to enable debug logging on prod.

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -286,6 +286,7 @@ where
                             if r.reason.contains("return data out of bounds") {
                                 Ok(HandleOpsOut::PostOpRevert)
                             } else {
+                                tracing::debug!("handle_ops failed with request: {:?}", tx);
```
dancoombs, line 289: All of these errors are under `debug`. Do we want to be able to see them in prod? If we don't think these errors should happen we can make them `error!`

dancoombs (approved):

PR #1027, opened 2025-02-18
feat(monad): monad changes - DO NOT MERGE

0xfourzerofour:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -


crates/provider/src/alloy/cache.rs
```diff
@@ -0,0 +1,118 @@
...
+                if let Some(&block_number) =
+                    self.block_hash_cache.read().get(&block_hash.block_hash)
+                {
+                    return Ok(BlockId::Number(block_number));
+                }
+
+                match self
+                    .inner
+                    .get_block_by_hash(block_hash.block_hash, BlockTransactionsKind::Hashes)
+                    .await?
```
dancoombs, line 71: The issue here is that if a bunch of parallel threads get to this line at the same time, each of them will issue this request.

We get around this in the da providers by just using a regular lock on both reads and writes such that only 1 thread would ever need to issue the RPC call. The perf benefit of that will outweigh the lock overhead. 

https://github.com/alchemyplatform/rundler/blob/0dc72eca308ed94063242ca7689529233af4d70c/crates/provider/src/alloy/da/local/bedrock.rs#L97

crates/provider/src/alloy/cache.rs
```diff
@@ -0,0 +1,118 @@
...
+                    None => Ok(BlockId::Number(BlockNumberOrTag::Latest)),
+                }
+            }
+            BlockId::Number(number) => Ok(BlockId::Number(number)),
+        }
+    }
+}
+
+/// Cache to map block hash to the block number
+pub struct BlockNumberCache<AP, T>(BlockNumberCacheInner<AP, T>);
```
dancoombs, line 89: Whats the point of the inner cache if everything is a pass through?

crates/provider/src/alloy/metrics.rs
```diff
@@ -98,6 +100,10 @@ where
                     method_logger.record_rpc(get_rpc_status_code(resp));
                     if resp.is_error() {
                         let error = resp.as_error().unwrap();
+                        if get_method_name(&req_clone) == "eth_getLogs" {
+                            println!("{:?}", req_clone.serialize());
+                            println!("{:?}", error);
+                        }
```
dancoombs, line 106: nit: remove

PR #1030, opened 2025-02-20
feat: add otlp tracing

alex-miao:
## Proposed Changes

  - Add OTLP tracing layer to send spans
  - Add `LOG_OTLP_GRPC_ENDPOINT` environment variable
  - Instrument a basic span

### DIFF
--- a/bin/rundler/src/cli/tracing.rs
+++ b/bin/rundler/src/cli/tracing.rs
@@ -35,14 +35,14 @@ pub fn configure_logging(config: &LogsArgs) -> anyhow::Result<WorkerGuard> {
 
     let metadata = MetadataMap::new();
 
-    // let exporter = opentelemetry_otlp::SpanExporter::builder()
-    //     .with_http()
-    //     .with_endpoint("http://127.0.0.1:4318")
-    //     .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
-    //     .with_timeout(Duration::from_secs(3))
-    //     // .with_metadata(metadata)
-    //     .build()?;
-    let exporter = opentelemetry_stdout::SpanExporter::default();
+    let exporter = opentelemetry_otlp::SpanExporter::builder()
+        .with_tonic()
+        .with_endpoint("http://127.0.0.1:4317")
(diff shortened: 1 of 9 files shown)


bin/rundler/src/cli/tracing.rs
```diff
@@ -31,13 +35,34 @@ pub fn configure_logging(config: &LogsArgs) -> anyhow::Result<WorkerGuard> {
...
+        let subscriber = subscriber_builder
+            .json()
+            .finish()
+            .with(TargetBlacklistLayer);
+        if let Some(endpoint) = &config.otlp_grpc_endpoint {
+            let exporter = opentelemetry_otlp::SpanExporter::builder()
+                .with_tonic()
+                .with_endpoint(endpoint)
+                .with_timeout(Duration::from_secs(3))
+                .with_metadata(MetadataMap::new())
```
andysim3d, line 49: hum I am thinking maybe we want add network in metadata ...?

crates/rpc/src/eth/router.rs
```diff
@@ -85,6 +89,14 @@ pub(crate) struct EntryPointRouter {
     v0_7: Option<(Address, Arc<dyn EntryPointRoute>)>,
 }
 
+impl fmt::Debug for EntryPointRouter {
```
andysim3d, line 92: well you don't need this. it should be skipped. we don't care EntryPointRouter per reuqest.

crates/rpc/src/eth/server.rs
```diff
@@ -46,6 +47,7 @@ where
         .await
     }
 
+    #[instrument(name = "EthApiServer::estimate_user_operation_gas", skip(self))]
```
andysim3d, line 50: also skip state_override

andysim3d: for demo purpose, could you recursively instrument to calls with estimate_user_operation_gas? otherwise you won't see a full span.

crates/rpc/src/eth/router.rs
alex-miao: ah yeah missed deleting this

andysim3d (approved):

PR #1039, opened 2025-02-25
feat: add max uo per sender permission

dancoombs:
## Proposed Changes

  - Add a new permissions to allow more max allowed UOs in the pool for trusted senders

TODO:
- [x] Docs
- [x] Local testing


0xfourzerofour (approved):
LGTM

PR #1040, opened 2025-02-25
feat: add permission for relaxed fee checking

dancoombs:
## Proposed Changes

  - Add two new permissions: one for relaxing fee check for acceptance to the pool, one for relaxing fee check for bundling

TODO
- [x] Unit tests
- [x] Local tests
- [x] Docs

### DIFF
--- a/crates/pool/src/server/remote/protos.rs
+++ b/crates/pool/src/server/remote/protos.rs
@@ -621,6 +621,8 @@ impl From<UserOperationPermissions> for RundlerUserOperationPermissions {
             max_allowed_in_pool_for_sender: permissions
                 .max_allowed_in_pool_for_sender
                 .map(|c| c as usize),
+            underpriced_accept_pct: permissions.underpriced_accept_pct,
+            underpriced_bundle_pct: permissions.underpriced_bundle_pct,
         }
     }
 }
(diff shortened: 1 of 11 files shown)


crates/sim/src/precheck.rs
```diff
@@ -299,10 +306,12 @@ where
         // if preVerificationGas is dynamic, then allow for the percentage buffer
         // and check if the preVerificationGas is at least the minimum.
         if self.chain_spec.da_pre_verification_gas {
-            min_pre_verification_gas = math::percent(
-                min_pre_verification_gas,
+            let accept_pct = cmp::min(
```
dancoombs, line 309: Unit tests needed for this logic

crates/pool/src/server/remote/protos.rs
```diff
@@ -632,6 +634,8 @@ impl From<RundlerUserOperationPermissions> for UserOperationPermissions {
             max_allowed_in_pool_for_sender: permissions
                 .max_allowed_in_pool_for_sender
                 .map(|c| c as u64),
+            underpriced_accept_pct: permissions.underpriced_accept_pct,
```
andysim3d, line 637: just curious, why not use 
```
permissions.underpriced_accept_pct.map(|c| c as u32)
```
here to handle optional?

dancoombs: Not sure I understand? Both `underpriced_accept_pct` are already u32 so no mapping is required

PR #1041, opened 2025-02-25
feat: implement bundler sponsorship

dancoombs:
## Proposed Changes

  - Add a bundler sponsorship permission
  - This permission enforces that the bundler pays all of the gas for the user operation, up to a max cost and in a time range
  - Time range is best effort as its not enforced onchain.

TODO:
- [x] Unit tests
- [x] Local testing
- [x] Docs

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -44,8 +44,8 @@ use rundler_types::{
     pool::{Pool, PoolOperation, SimulationViolation},
     proxy::SubmissionProxy,
     Entity, EntityInfo, EntityInfos, EntityType, EntityUpdate, EntityUpdateType, GasFees,
-    Timestamp, UserOperation, UserOperationVariant, UserOpsPerAggregator, ValidationRevert,
-    BUNDLE_BYTE_OVERHEAD, TIME_RANGE_BUFFER, USER_OP_OFFSET_WORD_SIZE,
+    Timestamp, UserOperation, UserOperationVariant, UserOpsPerAggregator, ValidTimeRange,
+    ValidationRevert, BUNDLE_BYTE_OVERHEAD, TIME_RANGE_BUFFER, USER_OP_OFFSET_WORD_SIZE,
 };
 use rundler_utils::{emit::WithEntryPoint, guard_timer::CustomTimerGuard, math};
 use tokio::{sync::broadcast, try_join};
@@ -395,8 +395,10 @@ where
         }
 
         // filter by fees
-        if op.uo.max_fee_per_gas() < required_max_fee_per_gas
-            || op.uo.max_priority_fee_per_gas() < required_max_priority_fee_per_gas
+        if (op.uo.max_fee_per_gas() < required_max_fee_per_gas
+            || op.uo.max_priority_fee_per_gas() < required_max_priority_fee_per_gas)
+            && op.perms.bundler_sponsorship.is_none()
+        // skip if bundler sponsored
         {
             self.emit(BuilderEvent::skipped_op(
                 self.builder_tag.clone(),
@@ -420,6 +422,12 @@ where
         // TODO(bundle): assuming a bundle size of 1
         let bundle_size = 1;
 
+        let gas_price = if op.perms.bundler_sponsorship.is_some() {
+            required_op_fees.gas_price(base_fee)
+        } else {
+            op.uo.gas_price(base_fee)
+        };
+
         let required_da_gas = if self.settings.da_gas_tracking_enabled
             && da_block_data.is_some()
             && self.ep_providers.da_gas_oracle_sync().is_some()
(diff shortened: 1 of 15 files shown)


crates/builder/src/emit.rs
```diff
@@ -166,6 +166,8 @@ pub enum SkipReason {
         required_pvg: u128,
         actual_pvg: u128,
     },
+    /// Insufficient sponsorship cost for the operation
+    InsufficientSponsorshipCost { max_cost: U256, actual_cost: U256 },
```
dancoombs, line 170: nit fix name

crates/sim/src/precheck.rs
```diff
@@ -313,6 +326,8 @@ where
 
             min_pre_verification_gas = math::percent_ceil(min_pre_verification_gas, accept_pct);
         }
+        println!("min_pre_verification_gas: {}", min_pre_verification_gas);
+        println!("op.pre_verification_gas(): {}", op.pre_verification_gas());
```
dancoombs, line 330: remove

crates/builder/src/bundle_proposer.rs
```diff
@@ -598,6 +619,23 @@ where
...
+                    self.emit(BuilderEvent::rejected_op(
+                        self.builder_tag.clone(),
+                        op.hash(),
+                        OpRejectionReason::InvalidTimeRange {
+                            valid_range: valid_time_range,
+                        },
+                    ));
+                    context.rejected_ops.push((op.into(), po.entity_infos));
+                    continue;
+                }
```
dancoombs, line 638: Add unit tests for these checks

crates/pool/src/mempool/pool.rs
```diff
@@ -321,8 +348,10 @@ where
             self.best.insert(op.clone());
 
             // Check candidate status
-            if op.uo().max_fee_per_gas() < gas_fees.uo_fees.max_fee_per_gas
-                || op.uo().max_priority_fee_per_gas() < gas_fees.uo_fees.max_priority_fee_per_gas
+            if (op.uo().max_fee_per_gas() < gas_fees.uo_fees.max_fee_per_gas
+                || op.uo().max_priority_fee_per_gas() < gas_fees.uo_fees.max_priority_fee_per_gas)
+                && op.po.perms.bundler_sponsorship.is_none()
+            // skip if bundler sponsored
```
dancoombs, line 354: Add unit tests

crates/sim/src/precheck.rs
```diff
@@ -302,6 +302,19 @@ where
...
+            violations.push(PrecheckViolation::CallGasLimitTooLow(
+                op.call_gas_limit(),
+                MIN_CALL_GAS_LIMIT,
+            ));
+        }
+
+        // If the UO is bundler sponsored, skip the fee checks
+        if perms.bundler_sponsorship.is_some() {
+            return violations;
+        }
```
dancoombs, line 315: Unit test

crates/pool/src/mempool/pool.rs
```diff
@@ -268,10 +268,37 @@ where
                     },
                 });
                 expired.push(*hash);
+                continue;
+            } else if op
+                .po
+                .perms
+                .bundler_sponsorship
+                .as_ref()
+                .is_some_and(|s| Timestamp::from(s.valid_until) < block_timestamp)
```
0xfourzerofour, line 277: if the current block is the valid until wouldnt the next block be invalid so shouldnt it be <= ?

crates/rpc/src/eth/api.rs
```diff
@@ -69,6 +69,29 @@ where
             )));
         }
 
+        if permissions.bundler_sponsorship.is_some() {
+            if op.max_fee_per_gas() != 0 {
+                return Err(EthRpcError::InvalidParams(
```
0xfourzerofour, line 74: nit: these three if statements could be combined into one with | syntax. The message would not be as explicit but would remove some repeated code

crates/types/src/gas.rs
```diff
@@ -33,4 +34,12 @@ impl GasFees {
             ),
         }
     }
+
+    /// Get the gas price from these fees given a base fee
+    pub fn gas_price(self, base_fee: u128) -> u128 {
+        cmp::min(
+            self.max_fee_per_gas,
+            base_fee + self.max_priority_fee_per_gas,
```
0xfourzerofour, line 42: is there potential for an overflow here?

0xfourzerofour (dismissed):
Looks good just a couple nits

crates/rpc/src/eth/api.rs
dancoombs: Good call, that removed a bunch of code

crates/types/src/gas.rs
dancoombs: Yeah, someone would really have to pass in weird fees to get that, but we should protect against it since its behind an interface. Fixed

0xfourzerofour (approved):
lgtm

PR #1046, opened 2025-03-06
feat(provider): use anynetwork, add get full block/get balances

dancoombs:
## Proposed Changes

  - Add `get_full_block` and `get_balances` to `EvmProvider`
  - For `get_full_block`: alloy requires the use of `AnyNetwork` to deserialize blocks for any network, else we'd get errors deserializing arbitrum, optimism, etc blocks.
  - Modify the provider crate and types to use `AnyNetwork`

### DIFF
--- a/crates/provider/src/alloy/da/local/bedrock.rs
+++ b/crates/provider/src/alloy/da/local/bedrock.rs
@@ -12,7 +12,7 @@
 // If not, see https://www.gnu.org/licenses/.
 
 use alloy_primitives::{Address, Bytes};
-use alloy_provider::Provider as AlloyProvider;
+use alloy_provider::network::AnyNetwork;
 use alloy_rpc_types_eth::state::{AccountOverride, StateOverride};
 use alloy_transport::Transport;
 use anyhow::Context;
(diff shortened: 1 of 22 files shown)


crates/provider/src/alloy/da/local/bedrock.rs
```diff
@@ -42,15 +42,15 @@ const MIN_TRANSACTION_SIZE: i128 = 100_000_000;
...
+    oracle: GasPriceOracleInstance<T, AP, AnyNetwork>,
+    multicaller: MulticallInstance<T, AP, AnyNetwork>,
     block_data_cache: TokioMutex<LruMap<BlockHashOrNumber, BedrockDAGasBlockData>>,
     blocking_task_pool: BlockingTaskPool,
 }
 
 impl<AP, T> LocalBedrockDAGasOracle<AP, T>
 where
-    AP: AlloyProvider<T> + Clone,
+    AP: AlloyProvider<T>,
```
0xfourzerofour, line 53: can you not put the network at the base level of AlloyProvider? Its one of the generics on the trait

```
pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:

```

That way we do not need to put it everywhere, or is this feature dependent?

0xfourzerofour: actually wait the provider is already provided with Ethereum at the alloy level I think I am wrong

dancoombs: Are you suggesting something different than what I did here?

https://github.com/alchemyplatform/rundler/blob/d5e6d077956765f23f814886654b8811bea8989d/crates/provider/src/lib.rs#L82

The above code is only generic on the transport layer, which is a pretty complex type that changes if we add retries and such.

0xfourzerofour (approved):

PR #1047, opened 2025-03-06
feat(pool): add address tracking to pool

dancoombs:
## Proposed Changes

  - Modify new heads subscription to pass addresses to track
  - Track addresses during chain tracking, adding address updates from any of the tracked addresses
  - Address updates contains latest nonce, balance, and any mined transaction hashes

### DIFF
--- a/crates/pool/src/server/remote/client.rs
+++ b/crates/pool/src/server/remote/client.rs
@@ -88,6 +88,7 @@ impl RemotePoolClient {
     async fn new_heads_subscription_handler(
         client: OpPoolClient<Channel>,
         tx: mpsc::UnboundedSender<NewHead>,
+        to_track: Vec<Address>,
     ) {
         let mut stream = None;
 
@@ -98,7 +99,13 @@ impl RemotePoolClient {
                         "subscribe new heads",
                         || {
                             let mut c = client.clone();
-                            async move { c.subscribe_new_heads(SubscribeNewHeadsRequest {}).await }
+                            let to_track = to_track.clone();
+                            async move {
+                                c.subscribe_new_heads(SubscribeNewHeadsRequest {
+                                    to_track: to_track.iter().map(|a| a.to_proto_bytes()).collect(),
+                                })
+                                .await
+                            }
                         },
                         UnlimitedRetryOpts::default(),
                     )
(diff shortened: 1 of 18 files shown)


crates/pool/src/server/remote/client.rs
```diff
@@ -98,7 +99,13 @@ impl RemotePoolClient {
                         "subscribe new heads",
                         || {
                             let mut c = client.clone();
-                            async move { c.subscribe_new_heads(SubscribeNewHeadsRequest {}).await }
+                            let to_track = to_track.clone();
```
0xfourzerofour, line 102: not really related to this change however I was wondering about this new_head_subscriber_handler function. to me it does not seem like we need an outer loop and can just use a while let to read the next item in the stream. not sure if there was a specific reason for the loop while reading the stream iteratively in a loop but thought I would point it out 


```
        let mut stream = retry::with_unlimited_retries(
            "subscribe new heads",
            || {
                let mut c = client.clone();
                let to_track = to_track.clone();
                async move {
                    c.subscribe_new_heads(SubscribeNewHeadsRequest {
                        to_track: to_track.iter().map(|a| a.to_proto_bytes()).collect(),
                    })
                    .await
                }
            },
            UnlimitedRetryOpts::default(),
        )
        .await
        .into_inner();

        while let Ok(update) = stream.message().await {
            match update {
                Some(SubscribeNewHeadsResponse { new_head: Some(b) }) => match b.try_into() {
                    Ok(new_head) => {
                        if tx.send(new_head).is_err() {
                            return;
                        }
                    }
                    Err(e) => {
                        tracing::error!("error parsing new block: {:?}", e);
                        break;
                    }
                },
                Some(SubscribeNewHeadsResponse { new_head: None }) | None => {
                    tracing::debug!("block subscription closed");
                }
            }
        }

```

0xfourzerofour (approved):
Overall looks good just ad a comment about a semi unrelated structure

crates/pool/src/server/remote/client.rs
0xfourzerofour: Looks like tonic Streaming implements Stream + StreamExt so it could also be done using .next().await to make things consistent across other parts of the codebase that implements stream

dancoombs: Its been a while since I've looked at this code, but I think the outer loop is there to keep a persistent connection on the stream. If the stream ever disconnects it will loop around and issue a new subscription request.

PR #1048, opened 2025-03-06
feat(builder): consume address updates remove sender status

### LINKED ISSUE #887: [pool/builder] Track builder addresses via chain tracking
**Describe the feature**
Builders currently wait for new block updates and then issue a few RPC calls to determine if one of their own transactions has been mined.

Instead, we could have the builders register their addresses with the pool in their block subscription call. The pool can then track those addresses for transactions alongside the rest of its chain tracking logic. It can then send any receipts alongside the new block update that it sends via the block subscription.

dancoombs:
Closes #887 

## Proposed Changes

  - Consume address updates in builder
  - Remove sender-based transaction tracking
  - Side-effect is that we no longer process dropped transactions, but this has proven to be of little value in production and is turned off on most networks.


crates/pool/src/chain.rs
```diff
@@ -318,6 +358,22 @@ impl<P: EvmProvider> Chain<P> {
...
+        added_blocks
+            .iter()
+            .flat_map(|block| &block.address_updates)
+            .for_each(|update| {
+                let latest_update = address_updates
+                    .entry(&update.address)
+                    .or_insert(update.clone());
+                if update.nonce > latest_update.nonce {
+                    *latest_update = update.clone();
+                }
```
dancoombs, line 373: TODO: fix this to append transaction hashes

crates/builder/src/transaction_tracker.rs
```diff
@@ -241,39 +247,50 @@ where
...
+                gas_limit,
+                gas_used: Some(r.gas_used),
+                gas_price: Some(r.effective_gas_price),
+                is_success: r.inner.status(),
+            }),
             None => {
-                warn!("failed to fetch transaction receipt for tx: {}", tx_hash);
-                (None, None, false)
+                warn!("failed to find transaction receipt for tx: {}", tx_hash);
+                Ok(MinedTxInfo::default())
```
dancoombs, line 269: Why does this do something different than when fetching the tx?

crates/pool/src/chain.rs
```diff
@@ -573,10 +577,16 @@ impl<P: EvmProvider> Chain<P> {
...
+                let update = updates.entry(tx.from).or_insert(AddressUpdate {
+                    address: tx.from,
+                    nonce,
+                    balance: U256::ZERO, // placeholder
+                    mined_tx_hashes: vec![],
+                });
+                if nonce > update.nonce {
+                    update.nonce = nonce;
                 }
+                update.mined_tx_hashes.push(tx.inner.tx_hash());
```
dancoombs, line 589: Add test for this

crates/builder/src/transaction_tracker.rs
dancoombs: It doesn't.

0xfourzerofour (approved):
Nice, we may want to clean up the terraform code to conform to the removed env vars as well.

good to consolidate the tx status info across different senders

PR #1060, opened 2025-03-10
feat: track total 7702 uos

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/pool/src/mempool/pool.rs
+++ b/crates/pool/src/mempool/pool.rs
@@ -20,7 +20,7 @@ use std::{
 
 use alloy_primitives::{Address, B256};
 use anyhow::Context;
-use metrics::{Gauge, Histogram};
+use metrics::{Counter, Gauge, Histogram};
 use metrics_derive::Metrics;
 use parking_lot::RwLock;
 use rundler_provider::DAGasOracleSync;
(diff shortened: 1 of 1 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -793,6 +797,8 @@ struct PoolMetrics {
     time_to_mine: Histogram,
     #[metric(describe = "the duration distribution of a blocked mined.")]
     blocks_to_mine: Histogram,
+    #[metric(describe = "the number of ups with 7702 auth")]
```
dancoombs, line 800: nit: ops

crates/pool/src/mempool/pool.rs
```diff
@@ -793,6 +797,8 @@ struct PoolMetrics {
...
+    num_7702_ops: Counter,
```
dancoombs, line 801: What is this metric meant to capture - 7702 UOs added to the pool? We should probably have a similar metric for total ops.

Maybe:

num_ops_added
num_7702_ops_added

crates/pool/src/mempool/pool.rs
andysim3d: Done

crates/pool/src/mempool/pool.rs
andysim3d: Done

0xfourzerofour (approved):
LGTM

PR #1070, opened 2025-03-18
feat(signer): update how signer is configured from cli

### LINKED ISSUE #1069: [signer] Finalize how to configure KMS+mnemonic for funding
**Describe the feature**
The current way on the feature branch is just for testing, we should come up with a better method

dancoombs:
Closes #1069

### DIFF
--- a/crates/signer/src/aws.rs
+++ b/crates/signer/src/aws.rs
@@ -14,7 +14,7 @@
 use std::time::Duration;
 
 use alloy_consensus::SignableTransaction;
-use alloy_network::TxSigner;
+use alloy_network::{EthereumWallet, TxSigner};
 use alloy_primitives::{Address, PrimitiveSignature};
 use alloy_signer_aws::AwsSigner;
 use anyhow::Context;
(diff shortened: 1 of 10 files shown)


crates/signer/src/aws.rs
```diff
@@ -25,6 +25,24 @@ use tokio::{sync::oneshot, time::sleep};
 
 use crate::Result;
 
+pub(crate) async fn create_wallet_from_key_ids(
+    key_ids: &[String],
```
dancoombs, line 29: Pass by value

crates/signer/src/manager.rs
```diff
@@ -232,6 +232,15 @@ impl SignerManager for FundingSignerManager {
             Err(anyhow::anyhow!("Manager does not support funding"))?
         }
 
+        // force a funding update cycle of the balances
```
dancoombs, line 235: nit: bad comment

0xfourzerofour (approved):
no comments on this bad boi, looks good

PR #1084, opened 2025-03-27
fix(builder): don't get block number/fees unless there are UOs to bundle

dancoombs:


andysim3d (approved):

0xfourzerofour (approved):
nice little optimization

PR #1096, opened 2025-04-08
feat: limit bundle fee by signer balance, max uo cost

dancoombs:


crates/builder/src/bundle_proposer.rs
```diff
@@ -671,18 +681,21 @@ where
             }
 
             // if the bundle is at or past target, skip op and continue to finish processing any rejections
-            if gas_spent >= self.settings.target_bundle_gas {
+            if execution_gas_spent >= self.settings.target_bundle_gas {
```
dancoombs, line 684: TODO: unit test

crates/sim/src/precheck.rs
```diff
@@ -316,6 +320,14 @@ where
 
         // NOTE: only fee checks below this short circuit
 
+        // Check if the total gas fee of this op is too high
+        if op.max_gas_cost() > max_uo_cost {
```
dancoombs, line 324: TODO: unit test

crates/builder/src/bundle_proposer.rs
```diff
@@ -189,7 +190,8 @@ where
 
     async fn make_bundle(
         &mut self,
-        required_fees: Option<GasFees>,
+        max_bundle_fee: U256,
```
0xfourzerofour, line 193: nit: it could be easier to reason about to call this `bundler_sender_balance` especially if the error is OverSenderBalance

0xfourzerofour (dismissed):

crates/builder/src/bundle_proposer.rs
dancoombs: Updated the error name

0xfourzerofour (approved):

PR #1102, opened 2025-04-15
feat: add CHANGELOG.md

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  - Add a changelog.md file to consolidate all change logs. 
  -
  -


dancoombs:
I think we should hold off on this unless we automate it. https://github.com/alchemyplatform/rundler/releases is sufficient for looking at changelogs across releases.

PR #1103, opened 2025-04-16
feat: remove mempool shards, lock senders to builders

dancoombs:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/builder/src/assigner.rs
+++ b/crates/builder/src/assigner.rs
@@ -1,0 +1,135 @@
+// This file is part of Rundler.
+//
+// Rundler is free software: you can redistribute it and/or modify it under the
+// terms of the GNU Lesser General Public License as published by the Free Software
+// Foundation, either version 3 of the License, or (at your option) any later version.
+//
+// Rundler is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+// without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along with Rundler.
+// If not, see https://www.gnu.org/licenses/.
+
+use std::{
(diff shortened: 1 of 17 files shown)


crates/builder/src/assigner.rs
```diff
@@ -0,0 +1,135 @@
+// This file is part of Rundler.
```
dancoombs, line 1: TODO: unit tests

crates/builder/src/bundle_sender.rs
```diff
@@ -376,6 +383,9 @@ where
                 } => {
                     info!("Bundle transaction mined: block number {block_number}, attempt number {attempt_number}, gas limit {gas_limit:?}, gas used {gas_used:?}, tx hash {tx_hash}, nonce {nonce}, success {is_success}");
 
+                    // release all operations after the bundle is mined
+                    self.assigner.release_all_operations(self.sender_eoa);
```
dancoombs, line 387: TODO: consider a way to make this safer, we can't forget to do this

crates/builder/src/bundle_sender.rs
```diff
@@ -632,6 +657,10 @@ where
             .await;
         self.metrics.bundle_txns_sent.increment(1);
 
+        if send_result.is_err() {
+            self.assigner.release_all_operations(self.sender_eoa);
+        }
```
dancoombs, line 662: This is wrong - there may be outstanding transactions, consider what to do in this case

crates/builder/src/bundle_proposer.rs
```diff
@@ -321,12 +342,27 @@ where
                 self.metric
                     .bundle_build_ms
                     .record(timer.elapsed().as_millis() as f64);
+
+                // debug only, check that the total ops are equal to the initial ops length
```
dancoombs, line 346: TODO: consider a way to enforce this in a more strict manner

PR #1105, opened 2025-04-17
feat: remove mempool shards and add builder sender assigner

### LINKED ISSUE #291: [builder] Add immediate bundling mode
**Describe the feature**
When the P2P mempool is online there will be little incentive to create bundles of > size 1 due to issues competition with other bundlers invalidating ops by setting a higher priority fee.

Rundler should support a mode where a UO is immediately converted into a bundle transaction and sent to the builder endpoint.

This will likely require a completely separate `bundle_proposer` and `bundler_sender` class that optimizes for speed over bundle size.

dancoombs:
Closes #291

## Proposed Changes

  - Remove mempool shards from the pool and builder
  - Add the builder `Assigner`. It is responsible for:
    - Retrieving user operations from the mempool for a particular entry point and filter ID
    - Assigning user operations to builder's ensuring that no two builders are trying to bundle the same sender at the same time.
    - Sender is used because a sender can invalidate its own UOs, even unstaked senders can have 4 UOs in the pool they can invalidate.
    - Builders unlock senders after they are done bundling them (either successfully or in failure cases).
  - Add 2 new methods to the mempool to better work with the assigner
    - `get_ops_summaries`: Just query for the operation entrypoint, hash, and sender to reduce memory copies. Used when determining which UOs to assign
    - `get_ops_by_hashes`: Retrieve a set of UOs by hashes. Used after assigning UOs.


crates/builder/src/assigner.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+#[derive(Metrics)]
+#[metrics(scope = "builder_assigner")]
+struct PerBuilderMetrics {
+    #[metric(describe = "the count of senders assigned to a builder.")]
+    senders_assigned: Gauge,
+    #[metric(describe = "the count of senders confirmed to a builder.")]
+    senders_confirmed: Gauge,
+}
```
dancoombs, line 268: TODO: unit tests

crates/builder/src/bundle_sender.rs
```diff
@@ -965,9 +1000,10 @@ impl<T: TransactionTracker, TRIG: Trigger> SenderMachineState<T, TRIG> {
     }
 
     async fn get_balance(&mut self, provider: impl EvmProvider) -> anyhow::Result<U256> {
-        if let Some(balance) = self.balance {
-            return Ok(balance);
-        }
+        // TODO this balance can be stale if the sender has been recently funded
```
dancoombs, line 1003: Existing bug, will fix in follow up

dancoombs: https://github.com/alchemyplatform/rundler/pull/1106

crates/builder/src/assigner.rs
```diff
@@ -0,0 +1,504 @@
...
+        let ops = self
+            .pool
+            .get_ops_summaries(entry_point, self.max_pool_ops_per_request, filter_id)
+            .await?;
+        let mut return_ops_summaries = Vec::new();
+
+        {
+            let mut state = self.state.lock().unwrap();
+            for op in ops {
+                let entry = state
```
0xfourzerofour, line 89: nit: I think it is cleaner to dereference the tuple like `let (hash, sender) = ...

crates/builder/src/assigner.rs
```diff
@@ -0,0 +1,504 @@
...
+        builder_address: Address,
+        confirmed_senders: impl IntoIterator<Item = &'a Address>,
+    ) {
+        let mut state = self.state.lock().unwrap();
+        let per_builder_metrics =
+            PerBuilderMetrics::new_with_labels(&[("builder_address", builder_address.to_string())]);
+
+        // Confirm all of the senders in state
+        for confirmed_sender in confirmed_senders.into_iter() {
+            let entry = state
```
0xfourzerofour, line 166: same here

crates/builder/src/task.rs
```diff
@@ -191,17 +192,23 @@ where
...
+            MAX_POOL_OPS_PER_REQUEST,
+            self.args.max_bundle_size,
+        );
+
         for ep in &self.args.entry_points {
             match ep.version {
                 EntryPointVersion::V0_6 => {
                     let actions = self
-                        .create_builders_v0_6(&task_spawner, ep, &self.signer_manager)
+                        .create_builders_v0_6(&task_spawner, ep, &self.signer_manager, &assigner)
```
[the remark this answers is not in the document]
dancoombs: As written, `Assigner` uses internal `Arc` on the shared state, so not needed here.

However, I think this is different than how we typically write these shared state global singletons, I should move the `Arc` from the internal state to the Assigner itself, remove `Clone`, and pass `Arc<Assigner>`, updating.

crates/builder/src/assigner.rs
```diff
@@ -0,0 +1,504 @@
...
+                entry_point,
+                return_ops_summaries.iter().map(|op| op.hash).collect(),
+            )
+            .await?;
+
+        Ok(return_ops)
+    }
+
+    // This method confirms the locks for the confirmed senders and drops the locks for the senders that are not confirmed.
+    // If a sender is assigned to the builder, is in the "assigned" state, and is not confirmed, it will be dropped.
```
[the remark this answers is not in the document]
dancoombs: Reworded

crates/builder/src/bundle_sender.rs
```diff
@@ -225,6 +235,10 @@ where
         state: &mut SenderMachineState<T, TRIG>,
     ) -> anyhow::Result<()> {
         let tracker_update = state.wait_for_trigger().await?;
+        if tracker_update.is_some() {
```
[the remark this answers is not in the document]
dancoombs: Added more to the comment

andysim3d (approved):

PR #1106, opened 2025-04-17
fix(builder): fix sender balance tracking to capture funding events

dancoombs:
## Proposed Changes

  - Fixes a bug in `bundle_sender` where its view of its own balance becomes stale after a funding event. Prior to this change, it would only be notified of a balance change if its own nonce changed. This doesn't handle funding events.
  - The fix is to add an address update for every tracked address on every new block. This adds one new always-performed parallel call to the chain tracking logic.
  - Address update now has an optional nonce indicating that it nonce changed, else None.

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -13,7 +13,7 @@
 
 use std::{pin::Pin, sync::Arc, time::Duration};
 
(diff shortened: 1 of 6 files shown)


andysim3d (dismissed):

0xfourzerofour (dismissed):
🚢

PR #1111, opened 2025-04-18
feat(builder): for v0.7+ only run the validation portion during bundle sim

### LINKED ISSUE #903: [builder] Add a reverting UO to bundle simulation to exclude execution
**Describe the feature**
We run a final `eth_call` to `handleOps` help ensure that our bundle won't revert onchain. Currently we run the entire `handleOps`, which includes all of the UOs executions.

To improve this, we can add a UO that reverts at the end of the bundle so that we can check to see if any individual UOs revert during validation, without running the full bundles execution.

This should ideally improve latency for a large bundle submission.

dancoombs:
Closes #903

## Proposed Changes

  - Skip bundle simulation if v0.7+ and there is only one UO in the bundle. This simulation was just performed.
  - If v0.7+ add a reverting UO as the last index in the bundle to skip the execution portion during simulation as it doesn't impact bundle validity.

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -41,9 +41,9 @@ use rundler_types::{
     pool::{PoolOperation, SimulationViolation},
     proxy::SubmissionProxy,
     BundleExpectedStorage, Entity, EntityInfo, EntityInfos, EntityType, EntityUpdate,
-    EntityUpdateType, ExpectedStorage, GasFees, Timestamp, UserOperation, UserOperationVariant,
-    UserOpsPerAggregator, ValidTimeRange, ValidationRevert, BUNDLE_BYTE_OVERHEAD,
-    TIME_RANGE_BUFFER, USER_OP_OFFSET_WORD_SIZE,
+    EntityUpdateType, EntryPointVersion, ExpectedStorage, GasFees, Timestamp, UserOperation,
+    UserOperationVariant, UserOpsPerAggregator, ValidTimeRange, ValidationRevert,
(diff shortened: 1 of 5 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -965,11 +965,20 @@ where
...
             .context("estimated bundle gas limit is larger than u64::MAX")?;
 
-        // call handle ops with the bundle to filter any rejected ops before sending
-        // TODO: if EP v0.7
-        // - if only 1 op, skip this call as simulation has already run a similar check
-        // - if more than 1 op, add a reverting UO to the bundle and call handle ops, this will skip running the execution portion of the op
+        // if EP v0.7+ and only 1 op, skip this call as simulation has already run a similar check
+        // v0.6 cannot do this as we need to check for postOp reverts
+        if self.ep_providers.entry_point().version() != EntryPointVersion::V0_6
+            && context.iter_ops().count() == 1
```
andysim3d, line 971: so it is we skip simulation for v0.7+ with only 1 UO in the bundle, not we enforce v0.7+ bundle can only have one UO per bundle.

andysim3d (approved):
comment the pr description, otherwise LGTM.

crates/builder/src/bundle_proposer.rs
dancoombs: Yes skip simulation for v0.7 with only 1 UO in bundle

andysim3d (approved):

0xfourzerofour (approved):
lgtm

PR #1146, opened 2025-05-28
feat(errors): catch error and map to invalid params

0xfourzerofour:
## Proposed Changes

  - Catch error that is currently mapped into an internal error


dancoombs (approved):
🚢

PR #1150, opened 2025-05-30
feat: redact secrets from all logs

### LINKED ISSUE #1107: Mask sensitive info logs during start rundler
**Describe the feature**
Right after starting `rundler`, it will print an INFO log, containing all client options, the log starts with `"INFO rundler::cli: Parsed CLI options:"`, while these logs contain many sensitive info, e.g., private keys, mnemonic, etc. Any idea to mask this info? e.g., for private keys, can print account addresses instead of private keys.

Raising this concern because in a production environment, info-level logs are much more easily to be accessed in many cases, instead of the machine's terminal.

**Additional context**
N/A.

dancoombs:
Closes #1107 

## Proposed Changes

Logs now look like

```
private_keys: [
    SecretBox<str>([REDACTED]),
],
```


0xfourzerofour (approved):
looks good, you could also use the tracing::instrument(skip(field_name))) to redact these in the logs but this looks more secure

dancoombs:
> looks good, you could also use the tracing::instrument(skip(field_name))) to redact these in the logs but this looks more secure

I think that only removes the field from the instrumentation. But if you print the field as part of the log message (i.e. in a `{secret:?}` it would still show up.

PR #1156, opened 2025-06-10
fix(sim): return decoded validation errors during estimation

dancoombs:
## Proposed Changes

  - Return the decoded validation error during estimation as opposed to the opaque bytes


0xfourzerofour (approved):
🚢

PR #1166, opened 2025-07-02
feat: add revert check to mempool add

dancoombs:
## Proposed Changes

  - Removes execution gas limit efficiency check from the mempool, this hasn't been useful
  - Adds a revert check to the mempool. If either execution or postop reverts the UO will be rejected
  - Implement the revert check in the entry point in three ways:
    - using `eth_simulateV1` - this should be used on most networks as its much more efficient
    - using `debug_traceCall`
    - using `eth_call` - this does not catch postop reverts, but does catch full execution reverts, only useful on monad at the moment.
   
TODO
- [x] Switch to using `simulateHandleOp` and add support for aggregated user ops
- [ ] Lots of testing in difference scenarios
- [x] Determine correct settings for various networks

### DIFF
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -15,9 +15,9 @@ use std::{collections::HashMap, net::SocketAddr, time::Duration};
 
 use alloy_primitives::Address;
 use anyhow::Context;
-use clap::{builder::ValueParser, Args};
+use clap::Args;
 use rundler_pool::{LocalPoolBuilder, PoolConfig, PoolTask, PoolTaskArgs};
-use rundler_provider::{Providers, RevertCheckMode};
+use rundler_provider::Providers;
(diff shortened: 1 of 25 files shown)


dancoombs:
Looks like we need to update the geth version in the 06 compliance tests

bin/rundler/src/cli/pool.rs
```diff
@@ -174,6 +174,23 @@ pub struct PoolArgs {
...
     )]
     pub max_time_in_pool_secs: Option<u64>,
+
+    #[arg(
+        long = "pool.revert_check_mode",
+        name = "pool.revert_check_mode",
+        env = "POOL_REVERT_CHECK_MODE",
+        value_parser = ValueParser::new(parse_revert_check_mode)
+    )]
+    pub revert_check_mode: Option<RevertCheckMode>,
```
dancoombs: TODO: docs

Issue #1184, opened 2025-08-20
[pool] Support Flashblocks


andysim3d:
**Describe the feature**
Support [flashblocks](https://docs.base.org/base-chain/flashblocks/apps) 

**Additional context**

PR #1185, opened 2025-08-20
feat: support flashblock #1: incremental update

andysim3d:
[Closes/Fixes] #

part of #1184
## Proposed Changes

  - in chain.rs: rundler is able to fetch a pending block and incremental update ChainUpdate.
  - other file changes are just pass through the flag to rundler and pool.
  -


crates/task/src/block_watcher.rs
```diff
@@ -29,11 +29,17 @@ pub async fn wait_for_new_block(
     provider: &impl EvmProvider,
     last_block_hash: B256,
     poll_interval: Duration,
+    flashblocks: bool,
```
dancoombs, line 32: This could just be `BlockId` and we can do the flashblocks check a layer above. Can add a check that the passed blockId one of pending, latest, finalized.

crates/pool/src/chain.rs
```diff
@@ -230,60 +238,126 @@ impl<P: EvmProvider> Chain<P> {
...
+            if !self.settings.flashblocks {
+                let now_ms = SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_millis();
+                let block_timestamp_ms = (block.header.timestamp * 1000) as u128;
+                self.metrics
+                    .block_discovery_delay_ms
+                    .record(now_ms.saturating_sub(block_timestamp_ms) as f64);
+            }
```
dancoombs, line 262: So is this metric not used at all if we're using flashblocks?

bin/rundler/src/cli/pool.rs
```diff
@@ -229,6 +229,7 @@ impl PoolArgs {
                 .verification_gas_limit_efficiency_reject_threshold,
             max_time_in_pool: self.max_time_in_pool_secs.map(Duration::from_secs),
             max_expected_storage_slots: common.max_expected_storage_slots.unwrap_or(usize::MAX),
+            flashblocks: false,
```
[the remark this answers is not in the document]
andysim3d: eventually it will be pass from chain config, but pool need to aware if it is flashblock enable to decide how to parse the update.

crates/pool/src/chain.rs
```diff
@@ -230,60 +238,126 @@ impl<P: EvmProvider> Chain<P> {
...
+                        continue; // same block
+                    }
+                    if pending_block.parent_hash == header.parent_hash {
+                        return chain_update; // same parent
+                    }
+                }
+                self.pending_block = Some(summary.clone());
+                if header.parent_hash == full_block_hash {
+                    return chain_update;
+                }
```
dancoombs, line 289: Can you explain this and the above early return statements?

dancoombs: I think I get it, this is handling the case where there is a new flashblock with new pending ops, but the parent hasn't changed (thus `latest` hasn't changed) and we can skip the parent block processing.

crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -258,6 +258,12 @@ service OpPool {
   // Remove a UserOperation by its id
   rpc RemoveOpById(RemoveOpByIdRequest) returns (RemoveOpByIdResponse);
 
+  // Mark a set of UOs in "pending" state
+  rpc MarkUoPending(MarkUoPendingRequest) returns (MarkUoPendingResponse);
+
+  // Get the bundle hash of the preconfirmed UO by UO hash
+  rpc GetPreConfirmedUo(GetPreConfirmedUoRequest) returns (GetPreConfirmedUoResponse);
```
andysim3d, line 265: reuse getOpsByHashes

crates/pool/src/chain.rs
```diff
@@ -77,6 +81,16 @@ pub(crate) struct ChainSubscriber {
...
     pub(crate) to_track: Arc<RwLock<HashSet<Address>>>,
 }
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub(crate) enum UpdateType {
+    // Include at least one full block
+    #[default]
+    Full,
+    // only include pending block
+    Partial,
```
dancoombs, line 91: nit: slightly confused by the naming here, could these map to `Confirmed` and `Preconfirmed`?

crates/rpc/src/eth/api.rs
```diff
@@ -153,18 +153,36 @@ where
     pub(crate) async fn get_user_operation_receipt(
         &self,
         hash: B256,
+        preconfirmation: Option<bool>,
```
dancoombs, line 156: How do you feel about making this a string enum and passing in "pending" as a tag? I think that maps to the other methods well https://docs.base.org/base-chain/flashblocks/apps#eth-getbalance

crates/rpc/src/eth/events/common.rs
```diff
@@ -102,7 +102,28 @@ where
     }
 
     #[instrument(skip_all)]
-    async fn get_receipt(&self, hash: B256) -> anyhow::Result<Option<RpcUserOperationReceipt>> {
+    async fn get_receipt(
+        &self,
+        hash: B256,
+        bundle_transaction: Option<B256>,
```
dancoombs, line 108: Is this specifically a preconfirmed bundle transaction or is this now used for all receipts? It seems like when this arg is used then a fully different path is used within this function, might be cleaner to make it its own function.

crates/pool/src/chain.rs
```diff
@@ -230,45 +250,149 @@ impl<P: EvmProvider> Chain<P> {
...
+                .record(now_ms.saturating_sub(block_timestamp_ms) as f64);
+
+            self.pending_block = Some(summary);
+            if header.parent_hash == full_block_hash {
+                return chain_update;
+            }
+
+            let preconfirmed_txns = chain_update.preconfirmed_txns;
+
+            for attempt in 0..=self.settings.max_sync_retries {
```
dancoombs, line 312: We may consider removing retries for flashblocks. By the time we've gone through these retries there is probably already a new full block to process. Maybe we just retry on full blocks?

crates/pool/src/chain.rs
```diff
@@ -230,45 +250,149 @@ impl<P: EvmProvider> Chain<P> {
 
     #[instrument(skip_all)]
     async fn wait_for_update(&mut self) -> ChainUpdate {
-        let mut block_hash = self
-            .blocks
-            .back()
-            .map(|block| block.hash)
-            .unwrap_or_default();
+        if self.settings.flashblocks && !self.blocks.is_empty() {
```
dancoombs, line 253: What is the reason for only doing this when the blocks list is empty?

crates/pool/src/chain.rs
```diff
@@ -230,45 +250,149 @@ impl<P: EvmProvider> Chain<P> {
...
-                block_hash,
+                latest_block_hash,
                 self.settings.poll_interval,
+                BlockId::pending(),
             )
             .await;
-            block_hash = hash;
+            latest_block_hash = hash;
 
+            let summary = match self.load_flash_block_summary(&block).await {
```
dancoombs, line 275: How do we know if this block is a flashblock or a confirmed block at this point?

PR #1188, opened 2025-08-28
feat(record): add metric to record bundle size in bytes

0xfourzerofour:
## Proposed Changes

  - add histogram metric to view the distribution of tx sizes sent to the nodes/sequencers.
 
 Does not look like alloy provides a built in tx::size interface so I create a pretty simple calculation function to optimistically calculate the size.

### DIFF
--- a/crates/utils/src/eth.rs
+++ b/crates/utils/src/eth.rs
@@ -50,3 +50,146 @@ pub fn calculate_transaction_size(bundle_data_size: usize, auth_list_count: usiz
 
     size
 }
+
+#[cfg(test)]
+mod tests {
+    use alloy_consensus::{Signed, TxEip7702, TxLegacy};
+    use alloy_eips::eip7702::{Authorization, SignedAuthorization};
+    use alloy_primitives::{Address, Bytes, Signature, U256};
+
+    use super::*;
+
+    #[test]
+    fn test_calculate_transaction_size_without_auth_list() {
+        let bundle_data_size = 1000;
+        let auth_list_count = 0;
+
+        let calculated_size = calculate_transaction_size(bundle_data_size, auth_list_count);
+
+        let tx = TxLegacy {
+            chain_id: Some(1),
+            nonce: 42,
+            gas_price: 20_000_000_000u128,
+            gas_limit: 21_000,
+            to: alloy_primitives::TxKind::Call(Address::ZERO),
+            value: U256::ZERO,
+            input: Bytes::from(vec![0u8; bundle_data_size]),
+        };
+
+        let signature = Signature::from_bytes_and_parity(&[0u8; 64], false);
+        let signed_tx = Signed::new_unchecked(tx, signature, Default::default());
+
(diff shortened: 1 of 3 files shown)


crates/utils/src/eth.rs
```diff
@@ -19,3 +19,29 @@ use alloy_primitives::Address;
 pub fn format_address(address: Address) -> String {
     format!("{:#x}", address).to_string()
 }
+
+/// Calculate the size of a bundle transaction in bytes including transaction overhead
+/// This function estimates the complete transaction size for a bundle before it's fully constructed
+/// Note: Bundle transactions are sent via EOA, so EIP-7702 authorization lists are not included
+pub fn calculate_bundle_transaction_size(bundle_data_size: usize) -> usize {
```
dancoombs, line 26: nit: this function doesn't have anything to do with bundles, its just the transaction size on to of any calldata

crates/utils/src/eth.rs
```diff
@@ -19,3 +19,29 @@ use alloy_primitives::Address;
 pub fn format_address(address: Address) -> String {
     format!("{:#x}", address).to_string()
 }
+
+/// Calculate the size of a bundle transaction in bytes including transaction overhead
+/// This function estimates the complete transaction size for a bundle before it's fully constructed
+/// Note: Bundle transactions are sent via EOA, so EIP-7702 authorization lists are not included
```
dancoombs, line 25: Why are authorization lists not included? We send type 4 transactions that include authorization lists

crates/utils/src/eth.rs
```diff
@@ -19,3 +19,29 @@ use alloy_primitives::Address;
...
+    size += bundle_data_size;
+
+    // Signature
+    size += 65; // signature (r, s, v)
+
+    // RLP encoding overhead
+    size += 50; // headers, lengths, etc.
+
+    size
+}
```
dancoombs, line 47: Can we add a test for this? It should be pretty easy to use Alloy to create a serialized transaction that we can compare against? With and without 7702 auth lists

crates/builder/src/bundle_proposer.rs
```diff
@@ -1647,11 +1650,13 @@ impl<UO: UserOperation> ProposalContext<UO> {
...
     fn get_bundle_transaction_size(&self, chain_spec: &ChainSpec) -> u128 {
-        self.bundle_overhead_bytes(chain_spec)
+        let bundle_data_size = (self.bundle_overhead_bytes(chain_spec)
             + self
                 .iter_ops_with_simulations()
                 .map(|sim_op| sim_op.op.abi_encoded_size() as u128)
-                .sum::<u128>()
+                .sum::<u128>()) as usize;
+
+        calculate_bundle_transaction_size(bundle_data_size) as u128
```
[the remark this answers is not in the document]
0xfourzerofour: updated

dancoombs (approved):
🚢 , nice work!

PR #1191, opened 2025-09-02
feat(monad): add argument to optionally charge total gas limit

0xfourzerofour:
## Proposed Changes

  - Add configuration option to state whether or not to charge the full gas limit on a UO

### DIFF
--- a/crates/sim/src/estimation/v0_7.rs
+++ b/crates/sim/src/estimation/v0_7.rs
@@ -177,17 +177,6 @@ where
                     .verification_gas_limit(verification_gas_limit)
                     .paymaster_verification_gas_limit(paymaster_verification_gas_limit)
                     .call_gas_limit(call_gas_limit)
-                    .build();
-                op_with_limits.required_pre_verification_gas(
-                    &self.chain_spec,
-                    bundle_size,
-                    da_gas,
-                    None,
-                )
-            } else {
-                // Use original op baseline to match validation behavior, but override
-                // paymaster_post_op_gas_limit to avoid using max_fill's inflated value
-                let op_with_limits = UserOperationBuilder::from_uo(base_op, &self.chain_spec)
                     .paymaster_post_op_gas_limit(op.paymaster_post_op_gas_limit.unwrap_or(0))
                     .build();
                 op_with_limits.required_pre_verification_gas(
@@ -196,6 +185,8 @@ where
                     da_gas,
                     None,
                 )
+            } else {
+                base_op.required_pre_verification_gas(&self.chain_spec, bundle_size, da_gas, None)
             }
         };
 


bin/rundler/src/cli/mod.rs
```diff
@@ -526,6 +526,14 @@ pub struct CommonArgs {
...
     )]
     pub aggregator_options: Vec<(String, String)>,
+
+    #[arg(
+        long = "charge_total_gas_limit",
+        env = "CHARGE_TOTAL_GAS_LIMIT",
+        default_value = "false",
+        global = true
+    )]
+    pub charge_total_gas_limit: bool,
```
dancoombs, line 536: I think we should add this as part of chain spec instead, given that its only needed on one chain.

Also, how about the name `charge_gas_limit_via_pvg`

crates/sim/src/estimation/v0_6.rs
```diff
@@ -166,6 +166,10 @@ where
             )
         };
 
+        if self.settings.charge_total_gas_limit {
+            pre_verification_gas += gas_limit;
```
dancoombs, line 170: Can you double check that this `gas_limit` doesn't have any extra gas added to it? I think `bundle_computation_gas_limit` also includes some gas from PVG already (just not the L1 part).

dancoombs (commented):
nit: PR title for this should be `feat:`

bin/rundler/src/cli/mod.rs
dancoombs: I don't recall wanting to deprecate, but we haven't always been super clean about what goes where.

Rule of thumb is that if its something chain related that should basically never change it goes in chain spec. If its a true configuration option or not chain related it goes in CLI.

dancoombs: The useful thing for `chain_spec` is that its the easiest way to get configuration variables into the most places as we pass it all over the place. Typically if you add something to it that config should be available to you in the class/function you're modifying. For example the function that needs to be modified for this feature already takes a chain spec https://github.com/alchemyplatform/rundler/blob/835b9bb23b82556032251b4765e3d8fbdb125c7e/crates/types/src/user_operation/mod.rs#L282

0xfourzerofour:
> > I think you should just need to change these two functions https://github.com/alchemyplatform/rundler/blob/main/crates/sim/src/gas.rs
> 
> Which I think just boils down to updating this
> 
> https://github.com/alchemyplatform/rundler/blob/835b9bb23b82556032251b4765e3d8fbdb125c7e/crates/types/src/user_operation/mod.rs#L282

yeh that is a lot more simple

crates/types/src/user_operation/mod.rs
```diff
@@ -225,8 +225,12 @@ pub trait UserOperation: Debug + Clone + Send + Sync + 'static {
     ///
     /// `bundle_size` is the size of the bundle if applying shared gas to the gas limit, otherwise `None`.
     fn bundle_gas_limit(&self, chain_spec: &ChainSpec, bundle_size: Option<usize>) -> u128 {
-        self.pre_verification_bundle_gas_limit(chain_spec, bundle_size)
-            + self.total_verification_gas_limit()
+        self.pre_verification_bundle_gas_limit(chain_spec, bundle_size) + self.operation_gas_limit()
+    }
+
+    /// Returns the combined gas limit for a user operation.
+    fn operation_gas_limit(&self) -> u128 {
```
dancoombs, line 232: We have so many different flavors of these functions its getting hard to name them 😆 . I've found just being super specific helps me remember what they do.

Lets call this `bundle_gas_limit_without_pvg`.

I'm also realizing that this function seems to currently miss paymaster post op in the `bundle_gas_limit`...

dancoombs: Oh jk it looks like its in `required_pre_execution_buffer` (the naming here is getting so confusing).

crates/sim/src/estimation/v0_6.rs
```diff
@@ -114,23 +114,41 @@ where
             );
         }
 
-        let pre_verification_gas_future = self.estimate_pre_verification_gas(&op, block_hash);
```
dancoombs, line 117: Is this dead code now?

0xfourzerofour: yeh still used in tests, have not finished the clean up

crates/sim/src/gas.rs
```diff
@@ -12,10 +12,158 @@
...
+        let gas_limits = verification_gas_limit
+            + paymaster_verification_gas_limit
+            + op.required_pre_execution_buffer()
+            + call_gas_limit;
+        pvegl = pvegl.saturating_add(gas_limits);
+    }
+
+    pvegl
+}
+
```
dancoombs, line 166: Is `estimate_pre_verification_gas` used anywhere now?

crates/sim/src/estimation/v0_6.rs
```diff
@@ -114,23 +114,57 @@ where
...
+                    .build();
+                op_with_limits.required_pre_verification_gas_with_limits(
+                    &self.chain_spec,
+                    bundle_size,
+                    da_gas,
+                    None,
+                    Some(verification_gas_limit),
+                    Some(0), // v0_6 doesn't have paymaster verification gas
+                    Some(call_gas_limit),
+                )
```
dancoombs, line 161: Since we're already setting the limits on the UO, how come we're still passing them to `required_pre_verification_gas_with_limits`? Couldn't we just use `required_pre_verification_gas` with a special handling for `charge_gas_limit_via_pvg`?

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -167,12 +167,10 @@ where
         let ops_len = ops.len();
         let da_gas: u64 = ops
             .iter()
-            .map(|op: &UserOperation| {
-                op.pre_verification_da_gas_limit(&self.chain_spec, Some(ops_len))
+            .map(|op| {
+                entry_point::calculate_da_gas_for_simulation(op, &self.chain_spec, Some(ops_len))
```
dancoombs, line 171: I'm not sure I understand these changes, it looks like `calculate_da_gas_for_simulation` is the same code as `pre_verification_da_gas_limit`. Or am I missing a small difference?

0xfourzerofour: ah yes this was left over from me testing stuff that was another issue

crates/sim/src/gas.rs
```diff
@@ -52,10 +49,55 @@ pub async fn estimate_pre_verification_gas<UO: UserOperation, E: DAGasProvider<U
...
+    fee_estimator: &F,
+    random_op: &UO,
+    max_fee_per_gas: Option<u128>,
+    max_priority_fee_per_gas: Option<u128>,
+    block: BlockHashOrNumber,
+    pvg_timer: Histogram,
+) -> anyhow::Result<u128> {
+    let _timer = CustomTimerGuard::new(pvg_timer);
+    let gas_price = if !chain_spec.da_pre_verification_gas {
+        0
```
dancoombs, line 79: nit: can return early here since there's no need to call `estimate_da_gas_only`

crates/types/src/user_operation/mod.rs
```diff
@@ -226,7 +229,12 @@ pub trait UserOperation: Debug + Clone + Send + Sync + 'static {
     /// `bundle_size` is the size of the bundle if applying shared gas to the gas limit, otherwise `None`.
     fn bundle_gas_limit(&self, chain_spec: &ChainSpec, bundle_size: Option<usize>) -> u128 {
         self.pre_verification_bundle_gas_limit(chain_spec, bundle_size)
```
dancoombs, line 231: Hm, I'm realizing now that this will cause us to over-estimate the gas limit needed to be set on bundles where the PVG already accounts for this overcharge.

Any ideas on how to handle cleanly?

We have the same problem for the `bundle_computation_gas_limit` call.

crates/types/src/user_operation/mod.rs
```diff
@@ -326,7 +335,39 @@ pub trait UserOperation: Debug + Clone + Send + Sync + 'static {
...
+        &self,
+        chain_spec: &ChainSpec,
+        bundle_size: usize,
+        da_gas: u128,
+        verification_gas_limit_efficiency_reject_threshold: Option<f64>,
+        verification_gas_limit: Option<u128>,
+        paymaster_verification_gas_limit: Option<u128>,
+        call_gas_limit: Option<u128>,
+    ) -> u128 {
+        let base_pvg = if let Some(thresh) = verification_gas_limit_efficiency_reject_threshold {
```
dancoombs, line 370: This can all be reverted right?

0xfourzerofour: yeh some more remaining stuff

dancoombs: Yeah, it doesn't look like `required_pre_verification_gas_with_limits` is called anywhere else and the limits are always set to `None` in the one call spot

crates/types/src/user_operation/mod.rs
dancoombs: This may already be correct. It looks like `pre_verification_execution_gas_limit` recalculates the `static_pre_verification_gas` instead of using the user-provided value. So I think this is fine as is? We should verify though

crates/types/src/user_operation/mod.rs
```diff
@@ -316,6 +329,9 @@ pub trait UserOperation: Debug + Clone + Send + Sync + 'static {
     ///
     /// `bundle_size` is the size of the bundle
     /// `da_gas` is the DA gas cost for the user operation, calculated elsewhere
+    /// `verification_gas_limit` is the verification gas limit (if available from estimation)
+    /// `paymaster_verification_gas_limit` is the paymaster verification gas limit (if available from estimation)
+    /// `call_gas_limit` is the call gas limit (if available from estimation)
```
dancoombs, line 334: nit: remove

crates/types/src/user_operation/mod.rs
0xfourzerofour: yeh it does a subtraction in pre_verification_da_gas_limit if charge_gas_limit_via_pvg is enabled

crates/sim/src/estimation/v0_7.rs
```diff
@@ -139,22 +149,47 @@ where
...
         let paymaster_verification_gas_limit = paymaster_verification_gas_limit_result?;
         let call_gas_limit = call_gas_limit_result?;
 
+        // Calculate the final PVG now that we have all gas limits
+        let pre_verification_gas = if op.pre_verification_gas.is_some_and(|pvg| pvg != 0) {
+            op.pre_verification_gas.unwrap()
+        } else {
+            // TODO(bundle): assuming a bundle size of 1
+            let bundle_size = 1;
+            let base_op = op.max_fill(&self.chain_spec);
```
dancoombs, line 174: I think this maxes out the `paymaster_post_op_gas_limit`. I'd actually expect this to result in massive PVGs because the maxed `paymaster_post_op_gas_limit` would contribute to the total gas limit. If its not, we may have missed it in the op calculations.

crates/types/src/user_operation/mod.rs
```diff
@@ -336,6 +338,16 @@ pub trait UserOperation: Debug + Clone + Send + Sync + 'static {
             )
         } else {
             self.pre_verification_execution_gas_limit(chain_spec, Some(bundle_size)) + da_gas
+        };
+
+        if chain_spec.charge_gas_limit_via_pvg {
+            let total_vgl = self.total_verification_gas_limit();
+            let cgl = self.call_gas_limit();
+
+            let gas_limits = total_vgl + self.required_pre_execution_buffer() + cgl;
```
dancoombs, line 347: Looks like `required_pre_execution_buffer` contains `paymaster_post_op_gas_limit`, so if its maxed out those PVGs might be huge.

dancoombs: https://github.com/alchemyplatform/rundler/pull/1191/files#r2320245651

`base_op` is a `max_fill` and the post op isn't overwritten (its never estimated, only just provided - which is why we need to change paymaster logic to pass it during gas estimation).

0xfourzerofour: ah yeh right damn

dancoombs: > this gas calculation logic is the bane of my existence

So many edge cases. I wish I saw a way to make it cleaner but there are just so many different variations of how we need to add gas up for various purposes and chains.

0xfourzerofour: > its never estimated, only provided - which is why we need to change paymaster logic to pass it during gas estimation

Do you mean actually update the estimation code to estimate the post op instead of relying on input fields

If so wouldnt the postOp might on state changes from the main user operation execution which could be a pretty hefty change. 

Or do you mean just override the max_fill to have something like `.paymaster_post_op_gas_limit(op.paymaster_post_op_gas_limit.unwrap_or(0))`

dancoombs: Oh yeah definitely the latter. That override should work.

Thats exactly why we don't estimate postOp ourselves btw, its super complicated.

dancoombs (approved):
🚢 - thanks for dealing with all the back and forth on this. The existing code is terrible with so many edge cases.

Lets make sure to test this thoroughly when turned on and off, gotta make sure that bundles are profitable on Monad but are not overcharding on chains where its off.

0xfourzerofour:
> Lets make sure to test this thoroughly when turned on and off, gotta make sure that bundles are profitable on Monad but are not overcharding on chains where its off.

Yeh I am gonna do another deploy on staging for monad testnet + other testnets when the command is turned off

PR #1192, opened 2025-09-03
feat: block no builder to pool calls

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -409,13 +409,13 @@ impl<P: EvmProvider> Chain<P> {
         pending_block
             .transactions
             .txns()
-            .filter(|tx| {
-                tx.inner.to().is_some()
-                    && self
-                        .settings
-                        .entry_point_addresses
-                        .keys()
-                        .contains(&tx.inner.to().unwrap())
+            .filter(|tx| match tx.inner.to() {
+                Some(tx_inner) => self
+                    .settings
+                    .entry_point_addresses
+                    .keys()
+                    .contains(&tx_inner),
+                None => false,
             })
             .map(|tx| tx.inner.tx_hash())
             .collect()
@@ -1115,6 +1115,7 @@ mod tests {
     use std::ops::DerefMut;
 
     use alloy_consensus::{transaction::Recovered, SignableTransaction, TypedTransaction};
+    use alloy_eips::BlockNumberOrTag;
     use alloy_network_primitives::BlockTransactions;
     use alloy_primitives::{address, Log as PrimitiveLog, LogData, Signature};
     use alloy_rpc_types_eth::{Block as AlloyBlock, Transaction as AlloyTransaction};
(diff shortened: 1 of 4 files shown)


crates/pool/src/mempool/uo_pool.rs
```diff
@@ -230,6 +235,56 @@ where
...
+            .iter()
+            .map(|txn| self.get_transaction_mapping(*txn))
+            .collect();
+
+        let results = future::join_all(futures).await;
+
+        let txn_to_uos: Vec<(B256, Vec<B256>)> = results.into_iter().flatten().collect();
+
+        self.per_confirm_uos(txn_to_uos);
+    }
```
dancoombs, line 287: I traced the logic from `chain` to here and it seems like this is calling `get_transaction_receipt` on every transaction in each flashblock? That's going to be a ton of RPC calls. I'd suggest the following:

1. Move this logic into `chain` to not complicate `uo_pool` any further
2. In `chain`, while loading the flashblock, filter by transactions that are sent by our senders
3. Get all of the receipts and extract the UO hashes

crates/pool/proto/op_pool/op_pool.proto
```diff
@@ -258,6 +258,9 @@ service OpPool {
   // Remove a UserOperation by its id
   rpc RemoveOpById(RemoveOpByIdRequest) returns (RemoveOpByIdResponse);
 
+  // Get the bundle hash of the preconfirmed UO by UO hash
+  rpc GetPreConfirmedUo(GetPreConfirmedUoRequest) returns (GetPreConfirmedUoResponse);
```
dancoombs, line 262: Is the current plan to keep this or remove it? If keeping it, why is it useful over `PreconfInfo`?

andysim3d: I just removed it.

crates/pool/src/mempool/uo_pool.rs
andysim3d: That's better!

Updated by 1 and 3 but didn't do 2 here: in case there is front runner, I'll just filter transaction(s) by to EP. then per transaction, I'll filter UO by receipt log, and update all UOs to uo_pool. if the UO is in uo_pool, we update the uo transaction map. otherwise, we ignore the uo, as it might from other bundler.

crates/pool/src/mempool/pool.rs
```diff
@@ -89,6 +89,14 @@ pub(crate) struct PoolInner<D> {
...
     /// when enough new blocks have passed.
     mined_hashes_with_block_numbers: BTreeSet<(u64, B256)>,
+
+    /// Unmined UO to bundle transaction mapping.
+    #[allow(dead_code)]
+    unmined_uos_bundle_mapping: HashMap<B256, B256>,
+    /// Preconfimed UO to preconfirmed bundle.
+    preconfirmed_uos: BTreeSet<B256>,
+    /// Unmined bundle to UO mapping.
+    unmined_bundle_to_uos: HashMap<B256, BTreeSet<B256>>,
```
dancoombs, line 99: It doesn't look like this is inserted into or read from, only removed from, is it needed?

crates/pool/src/mempool/pool.rs
```diff
@@ -89,6 +89,14 @@ pub(crate) struct PoolInner<D> {
     /// Removed operation hashes sorted by block number, so we can forget them
     /// when enough new blocks have passed.
     mined_hashes_with_block_numbers: BTreeSet<(u64, B256)>,
+
+    /// Unmined UO to bundle transaction mapping.
+    #[allow(dead_code)]
+    unmined_uos_bundle_mapping: HashMap<B256, B256>,
+    /// Preconfimed UO to preconfirmed bundle.
+    preconfirmed_uos: BTreeSet<B256>,
```
dancoombs, line 97: Do you need this? Can you just use the keys in `unmined_uos_bundle_mapping`?

dancoombs: Looked below, looks like its UO->txn, so you likely don't also need the UO set

crates/pool/src/chain.rs
```diff
@@ -277,13 +402,73 @@ impl<P: EvmProvider> Chain<P> {
...
+                tx.inner.to().is_some()
+                    && self
+                        .settings
+                        .entry_point_addresses
+                        .keys()
+                        .contains(&tx.inner.to().unwrap())
+            })
+            .map(|tx| tx.inner.tx_hash())
+            .collect()
+    }
```
dancoombs, line 428: This misses calls to the entry point that are made through another "proxy" - which is actually something we use in worldchain PBH.

I think the filter should be updated to be either (1) txn to the entrypoint (2) txn from our senders

This does miss if another sender sends through a proxy, but I think thats fine to skip preconfirmation in those cases.

crates/pool/src/chain.rs
```diff
@@ -230,45 +251,149 @@ impl<P: EvmProvider> Chain<P> {
...
+                        continue;
+                    }
+                };
+
+                match result {
+                    Ok(mut update) => {
+                        self.metrics
+                            .block_sync_time_ms
+                            .record(start.elapsed().as_millis() as f64);
+                        update.preconfirmed_txns = preconfirmed_txns;
```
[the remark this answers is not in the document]
andysim3d: well, for 'eth_getBlockByNumber(pending)', we'll always get the preconfirmed block. if a block A is confirmed, when you call 'eth_getBlockByNumber(pending)', you'll get the next block A+1 or null if a new pending block is not ready, but we'll not get A again once it is confirmed.

crates/pool/src/chain.rs
```diff
@@ -230,45 +251,149 @@ impl<P: EvmProvider> Chain<P> {
...
-                block_hash,
+                latest_block_hash,
                 self.settings.poll_interval,
+                BlockId::pending(),
             )
             .await;
-            block_hash = hash;
+            latest_block_hash = hash;
 
+            let summary = match self.load_flash_block_summary(&block).await {
```
dancoombs, line 276: Flashblocks question - what happens when the sequencer is emitting its full block? Does it also emit a flashblock? I'm concerned that if we're only using the `pending` tag, which you mentioned only returns flashblocks, then we'll actually have to wait an extra 200ms to discover a new full block (when the next flashblock gets emitted).

crates/rpc/src/eth/events/common.rs
```diff
@@ -102,7 +102,28 @@ where
     }
 
     #[instrument(skip_all)]
-    async fn get_receipt(&self, hash: B256) -> anyhow::Result<Option<RpcUserOperationReceipt>> {
+    async fn get_receipt(
+        &self,
+        hash: B256,
+        bundle_transaction: Option<B256>,
```
dancoombs, line 108: Is this specifically a preconfirmed bundle transaction or is this now used for all receipts? It seems like when this arg is used then a fully different path is used within this function, might be cleaner to make it its own function.

crates/rpc/src/eth/events/common.rs
```diff
@@ -134,6 +155,11 @@ where
             return Ok(None);
         };
 
+        // ignore events from other entry points.
+        if event.address() != E::address(&self.chain_spec) {
+            return Ok(None);
+        }
```
dancoombs, line 161: Is this flashblocks related or was this a bugfix?

andysim3d: reverted, no longer needed.

crates/rpc/src/rundler.rs
```diff
@@ -64,7 +65,11 @@ pub trait RundlerApi {
 
     /// Gets the status of a user operation by user operation hash
     #[method(name = "getUserOperationStatus")]
-    async fn get_user_operation_status(&self, uo_hash: B256) -> RpcResult<RpcUserOperationStatus>;
+    async fn get_user_operation_status(
+        &self,
+        uo_hash: B256,
+        preconfirmation: Option<bool>,
```
dancoombs, line 71: I actually think we can skip adding this to this endpoint, since its only internal. We can just always assume that if flashblocks are enabled we should allow for the pending status to be returned.

crates/pool/src/chain.rs
```diff
@@ -277,13 +396,73 @@ impl<P: EvmProvider> Chain<P> {
...
     }
 
+    #[instrument(skip_all)]
+    // get all 4337 txns
+    async fn get_user_operations_from_pending_block(&self, pending_block: &Block) -> Vec<B256> {
+        pending_block
+            .transactions
+            .txns()
+            .filter(|tx| {
+                tx.inner.to().is_some()
```
0xfourzerofour, line 413: might be cleaner to do an `if let` here instead of the is_some and unwrap

crates/pool/src/chain.rs
```diff
@@ -993,43 +1219,84 @@ mod tests {
...
 
         fn get_block(&self, id: BlockId) -> Option<Block> {
-            let BlockId::Hash(RpcBlockHash {
-                block_hash: hash,
-                require_canonical: _,
-            }) = id
-            else {
-                panic!("get_block only supports hash ids");
-            };
+            if id == BlockId::pending() {
```
0xfourzerofour, line 1222: you could do a match here that might clean this up a bit

crates/pool/src/chain.rs
```diff
@@ -993,43 +1219,84 @@ mod tests {
...
+                let pending_block = self.pending_block.read();
+                if pending_block.is_none() {
+                    return None;
+                }
+                let pending_block_inner = self.pending_block.read().clone().unwrap();
```
0xfourzerofour, line 1227: do you need to call self.pending_block.read() here if you already have the pending block 3 lines above it?

crates/rpc/src/eth/api.rs
```diff
@@ -153,18 +153,32 @@ where
...
+        tag: Option<BlockTag>,
     ) -> EthResult<Option<RpcUserOperationReceipt>> {
         if hash == B256::ZERO {
             return Err(EthRpcError::InvalidParams(
                 "Missing/invalid userOpHash".to_string(),
             ));
         }
+        let tag = tag.unwrap_or(BlockTag::Latest);
+
+        let bundle_transaction: Option<B256> = if tag == BlockTag::Pending {
```
0xfourzerofour, line 165: this would also be a little cleaner using a `match tag`

0xfourzerofour (commented):
overall the logic looks fine to me but could be cleaned up a bit

dancoombs (approved):
🚢 - thanks for working through this! Lots of difficult edge cases and minimal documentation on how things are supposed to work. The result here is going to be huge for users doing non-sequential operations.

PR #1193, opened 2025-09-04
feat(chain): add transaction gas limit and modify CLI to use it for bundle limits

dancoombs:
## Proposed Changes

  - Add a new transaction gas limit field to chain spec
  - Add a function to chain spec to resolve the maximum transaction gas between the txn limit and the block limit
  - [BREAKING] Modify the names of the CLI parameters to control the bundle gas limits


0xfourzerofour (approved):
LGTM, will need to change helm + tf

PR #1194, opened 2025-09-05
feat: use latest block to update canonical head, not pending - 1

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -104,6 +104,7 @@ pub(crate) struct ChainUpdate {
     pub mined_ops: Vec<MinedOp>,
     pub unmined_ops: Vec<MinedOp>,
     pub preconfirmed_txns: Vec<(B256, Vec<B256>)>,
+    pub preconfirmed_block_number: Option<u64>,
     /// List of on-chain entity balance updates made in the most recent block
     pub entity_balance_updates: Vec<BalanceUpdate>,
     /// List of entity balance updates that have been unmined due to a reorg
@@ -294,7 +295,9 @@ impl<P: EvmProvider> Chain<P> {
                 .as_millis();
             let block_timestamp_ms = (pending_block.header.timestamp * 1000) as u128;
 
-            let chain_update = self.process_pending_block(&summary).await;
+            let chain_update = self
+                .process_pending_block(&summary, pending_block.header.number)
+                .await;
             let header = &pending_block.inner.header;
             if let Some(pending_block) = &self.pending_block {
                 if pending_block.hash == header.hash && latest_block_hash == full_block_hash {
@@ -312,6 +315,7 @@ impl<P: EvmProvider> Chain<P> {
             }
 
             let preconfirmed_txns = chain_update.preconfirmed_txns;
+            let preconfirmed_block_number = chain_update.preconfirmed_block_number;
 
             for attempt in 0..=self.settings.max_sync_retries {
                 if attempt > 0 {
@@ -328,6 +332,7 @@ impl<P: EvmProvider> Chain<P> {
                             .block_sync_time_ms
                             .record(start.elapsed().as_millis() as f64);
                         update.preconfirmed_txns = preconfirmed_txns;
+                        update.preconfirmed_block_number = preconfirmed_block_number;
                         return update;
                     }
                     Err(error) => {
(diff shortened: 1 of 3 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -229,8 +228,6 @@ where
     }
 
     pub(crate) fn preconfirm_txns(&mut self, txn_to_uos: Vec<(B256, Vec<B256>)>) {
-        // each call of this method should clear the preconfirmed uos before adding new ones.
-        self.preconfirmed_uos_bundle_mapping.clear();
```
dancoombs, line 233: Now that this is removed, where do UOs move out of this and into confirmed status?

crates/pool/src/chain.rs
```diff
@@ -261,19 +260,27 @@ impl<P: EvmProvider> Chain<P> {
...
                 self.settings.poll_interval,
                 BlockId::pending(),
-            )
-            .await;
-            latest_block_hash = hash;
+            );
+            let latest_block_fut = block_watcher::wait_for_new_block(
+                &self.provider,
+                // force it always get the latest block hash.
+                B256::ZERO,
```
dancoombs, line 274: I'm curious why this is needed when our other usage of it passes the latest block hash?

crates/pool/src/mempool/pool.rs
andysim3d: it will be removed once the uo is mined. https://github.com/alchemyplatform/rundler/blob/main/crates/pool/src/mempool/pool.rs#L693

crates/pool/src/chain.rs
andysim3d: if we pass the latest block hash, wait_for_new_block will block until it find a new block with different hash. consider that takes 2s to discover next block, then the whole "wait_for_update_flashblocks" will be blocked by 2s. 

I made it fetch whatever block it find, so there are 4 cases:
- both flashblock and fullblock updates
- flashblock updates, fullblock stays
- flashblock stays, full updates
- both stays

if we use latest block hash, case 2 and 4 are blocking to up to 2 seconds to reflect the change.

dancoombs: Ah okay makes sense.

Isn't case 3 blocked by flashblocks now though? Do we not care about that case?

I wonder if the cleaner way to do this is remove the `wait_for_new_block` function entirely and just handle the sleep inside of this function when there is neither a new flashblock nor a new fullblock

crates/pool/src/mempool/pool.rs
dancoombs: What if the UO never gets mined?

andysim3d: so that map updated the same as we handle uo track. if a uo never mined and out of date, it will removed from that map too. 

we maintain the map of (uo, txn).

assuming a reorg happened: 
- if the txn still mined but move to other block, then it will reflected when we call get_transaction_receipt.
- if the txn becomed unmined, get_transaction_receipt will return us null then we assume we the uo is unmined.

crates/rpc/src/eth/events/common.rs
```diff
@@ -134,6 +134,10 @@ where
             return Ok(None);
         };
 
+        if event.address() != E::address(&self.chain_spec) {
+            return Ok(None);
+        }
```
dancoombs, line 139: Why is this needed? Did flashblocks change something in the events that are returned?

crates/pool/src/mempool/pool.rs
dancoombs: I'm not following. What happens when:

1. A UO gets preconfirmed in a flashblock, and added to this list
2. The next flashblock, that UO is no longer in the list, so we never call `get_transaction_receipt` for that hash again.
3. When the next full block is mined, the UO is not in the list

crates/pool/src/chain.rs
andysim3d: ah I add a line `if pending_block.hash == header.hash && latest_block_hash == full_block_hash` to make 3 not early continue, so it will move forward to check 4.

crates/pool/src/chain.rs
```diff
@@ -311,16 +319,7 @@ impl<P: EvmProvider> Chain<P> {
...
-                let parent_hash = self.pending_block.as_ref().unwrap().parent_hash;
-                let result = match self.provider.get_full_block(parent_hash.into()).await {
-                    Ok(Some(parent_block)) => self.sync_to_block(parent_block).await,
-                    Ok(None) | Err(_) => {
-                        warn!("failed to get parent block, retry...");
-                        time::sleep(self.settings.poll_interval).await;
-                        continue;
-                    }
-                };
+                let result = self.sync_to_block(latest_block.clone()).await;
```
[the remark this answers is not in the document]
dancoombs: Discussed offline - will make this change as an optimization later

dancoombs (approved):
:shipit:

PR #1200, opened 2025-09-25
feat: support multiple mode of get balance

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -


dancoombs (changes requested):
Lets discuss this offline. We should avoid multiple ways of doing the same thing and the fan-out of `eth_call` if the balances list is large.

PR #1203, opened 2025-09-30
fix: flashblock will not retry

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/pool/src/chain.rs
+++ b/crates/pool/src/chain.rs
@@ -37,9 +37,7 @@ use rundler_contracts::{
         Withdrawn as WithdrawnV07,
     },
 };
-use rundler_provider::{
-    Block, BlockId, EvmProvider, Filter, Log, ProviderResult, TransactionTrait,
-};
+use rundler_provider::{Block, BlockId, EvmProvider, Filter, Log, TransactionTrait};
 use rundler_task::{block_watcher, GracefulShutdown};
 use rundler_types::{pool::AddressUpdate, EntryPointVersion, Timestamp, UserOperationId};
 use tokio::{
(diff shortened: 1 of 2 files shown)


crates/task/src/block_watcher.rs
```diff
@@ -32,19 +32,28 @@ pub async fn wait_for_new_block(
...
+
+/// get a block by tag.
+///
+/// This function polls the provider for the `block_id` block.
+pub async fn get_block(provider: &impl EvmProvider, block_id: BlockId) -> Option<(B256, Block)> {
+    let block = retry::with_unlimited_retries(
+        "watch latest block",
+        || provider.get_full_block(block_id),
+        UnlimitedRetryOpts::default(),
+    )
```
dancoombs, line 54: I don't think we want unlimited retries in the flashblocks case. If the `pending` tag stops working we shouldn't block here

crates/pool/src/chain.rs
dancoombs, line 271: Since we're doing this already, can this also just be `get_block`?

crates/pool/src/chain.rs
```diff
@@ -279,36 +274,39 @@ impl<P: EvmProvider> Chain<P> {
             );
             let (pending_block_res, latest_block_res) =
                 future::join(pending_block_fut, latest_block_fut).await;
```
dancoombs, line 276: Looking at these two futures it seems we need to add back the `sleep` if neither of them return a new block, else we'll be in a hot polling loop.

crates/pool/src/chain.rs
```diff
@@ -264,12 +264,7 @@ impl<P: EvmProvider> Chain<P> {
 
         loop {
             // TODO: parallelize the two block watcher calls. one for pending block and one for latest block.
```
dancoombs, line 266: Is this TODO still valid?

crates/task/src/block_watcher.rs
andysim3d: this 'unlimitedRetry' here is to make sure we get a response. the 'unlimited' logic was the infinity loop in the `wait_for_new_block`

dancoombs: Understood, but what happens if we're doing an unlimited retry for a flashblock and only getting 500s? That will block the entire chain progression even if the full blocks are getting 200s

0xfourzerofour (dismissed):
LGTM

crates/pool/src/chain.rs
```diff
@@ -260,96 +262,129 @@ impl<P: EvmProvider> Chain<P> {
...
                 future::join(pending_block_fut, latest_block_fut).await;
-            let (_, pending_block) = pending_block_res;
-            let (latest_block_hash, latest_block) = latest_block_res;
 
+            if latest_block_res.is_err() && pending_block_res.is_err() {
+                // if none of the call succeed, sleep and retry form the beginning.
+                time::sleep(self.settings.poll_interval).await;
+                continue;
+            }
+            let pending_update = self.handle_pending_block(pending_block_res).await;
```
dancoombs, line 277: nit: it would be cleaner to check if `pending_block_res` is a non-error prior to calling handle_pending_block to avoid needing the large nested if statement in `handle_pending_block`

dancoombs: Similar to what is done below for `handle_latest_blocks`

Issue #886, opened 2024-10-24
[pool] Improve chain tracking on inconsistent RPC providers


dancoombs:
**Describe the feature**
Inconsistent RPC provider: an RPC provider whose "consistency level" block height/block hash may be different for simultaneous requests.
Consistency level: On L1: "latest", "safe", "finalized". Typically tags that can be added to RPC requests.

Inconsistent RPCs cause issues for Rundler, but mostly in chain tracking. When the chain tracker sees a new block, it turns around and issues requests to get info about those logs, and then publishes that block to the pool and the builder. The pool and the builder then process the new state and my issue new RPC queries tagged with the new block hash that the chain tracker published.

Many of these subsequent requests fail with "unknown block" or a multitude of other RPC errors that are returned when the request is routed to a node that doesn't yet know about a block, or has processed a different fork of the chain.

There isn't much we can do about removing inconsistent RPCs. Many chains we support only have inconsistent RPC providers.

There are, however, things we can do to improve this.

1. In some sort of alloy middleware, we can catch errors that are related to inconsistent RPC providers, and issue retries. This will help hide inconsistency to the higher level logic, at the expense of higher latency.
2. Audit the codebase to find the places where we are tagging requests with a block number or a block hash. These are the areas that are going to be the most susceptible to issues here. We should ensure we're using the right logic in these places to get a consistent view of the chain.

andysim3d:
it make sense to define 3 categories:
- retry-able error(like resource exhausted),
- inconsistent error(like unknown blocks)
- non-retry-able error(like invalid argument),

and config different retry policies, e.g. interval, max retry count, etc.

0xfourzerofour:
In the chain tracker we can store the block numbers of the last N blocks in a ring buffer. check that the average discrepancy between the new heads and the current head is within an acceptable range or reset_and_initialize to account for these issues. So worse case we have this issue for N blocks (I would say maybe 50 would suffice)

PR #1204, opened 2025-09-30
feat: custom retry alloy middleware - handling server errors

shunsukew:
Partially resolve https://github.com/alchemyplatform/rundler/issues/886

## Proposed Changes

Retry requests when having defined providers' server errors with the cost of latency increase.
Proposed approach is modified version of alloy `RetryBackoffLayer` https://github.com/alloy-rs/alloy/blob/main/crates/transport/src/layers/retry.rs, instead of implementing another custom middleware. This is to avoid unnecessary `ResponsePacket` decodings for efficiency.


crates/provider/src/alloy/provider_retry.rs
```diff
@@ -0,0 +1,324 @@
...
+                        )));
+                    }
+                    trace!(%err, "retrying request");
+
+                    let current_queued_reqs = this.requests_enqueued.load(Ordering::SeqCst) as u64;
+
+                    // try to extract the requested backoff from the error or compute the next
+                    // backoff based on retry count
+                    let backoff_hint = this.policy.backoff_hint(&err);
+                    let next_backoff = backoff_hint.unwrap_or_else(|| this.initial_backoff());
```
shunsukew, line 257: perhaps should better set exponential backoff when `backoff_hint` is None.

crates/provider/src/alloy/provider_retry.rs
```diff
@@ -0,0 +1,324 @@
...
+use alloy_json_rpc::{ErrorPayload, RequestPacket, ResponsePacket};
+use alloy_transport::{RpcError, TransportError, TransportErrorKind, TransportFut};
+use once_cell::sync::Lazy;
+use regex::Regex;
+use serde::Deserialize;
+#[cfg(not(target_family = "wasm"))]
+use tokio::time::sleep;
+use tower::{Layer, Service};
+use tracing::trace;
+#[cfg(target_family = "wasm")]
```
andysim3d: we don't have plan to support rundler in wasm env, could you remove this, and only keep branch of not wasm?

crates/provider/src/alloy/provider_retry.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+}
+
+/// [ProviderRetryPolicy] implements [RetryPolicy] to determine whether to retry depending on the
+/// err.
+#[derive(Debug, Copy, Clone, Default)]
+#[non_exhaustive]
+pub(crate) struct ProviderRetryPolicy;
+
+/// [RetryPolicy] defines logic for which [TransportError] instances should
```
andysim3d: could you use https://github.com/alloy-rs/alloy/blob/main/crates/transport/src/layers/retry.rs#L115-L121 instead of redefine the trait?

crates/provider/src/alloy/provider_retry.rs
```diff
@@ -0,0 +1,324 @@
...
+        // Our middleware doesn't care about backpressure, so it's ready as long
+        // as the inner service is ready.
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: RequestPacket) -> Self::Future {
+        let inner = self.inner.clone();
+        let this = self.clone();
+        let mut inner = std::mem::replace(&mut self.inner, inner);
+        Box::pin(async move {
```
andysim3d: we use BoxFuture in rundler, you could find an example: https://github.com/alchemyplatform/rundler/blob/main/crates/provider/src/alloy/metrics.rs#L92

andysim3d (commented):
Thanks for the contribution! left some minor comments

shunsukew:
> Thanks for the contribution! left some minor comments

Thank you for the review! I changed to use `BoxFuture`.
It seems I added changes before you submit review comments, codes has changed since your last review, and now this PR contains separate `RetryBackoffLayer` for provider server errors.

This is for having different retry logics and configs from the one available in alloy middleware. Alloy one is solely focuses on rate limit `retry-able error (like resource exhausted)`, while custom layer from this PR focuses on retry-able server errors `inconsistent error(like unknown blocks)`.

Please let me know if prev version of middleware is more preferable. I can revert it back.

shunsukew:
> @shunsukew sorry for the long wait - your pr lgtm, just need you run cargo sort to make sure your cargo file is well sorted.

Thank you so much. sure, pushed commit after `cargo sort`

crates/provider/src/alloy/provider_error_retry.rs
```diff
@@ -0,0 +1,151 @@
...
+use futures_util::FutureExt;
+use once_cell::sync::Lazy;
+use regex::Regex;
+use tokio::time::sleep;
+use tower::{Layer, Service};
+use tracing::trace;
+
+static BLOCK_NOT_FOUND_RE: Lazy<Regex> = Lazy::new(|| {
+    // case-insensitive; 0x + 64 hex chars
+    Regex::new(r"(?i)^block 0x[0-9a-f]{64} not found$").unwrap()
```
dancoombs: In our recent testing we've also seen this error come up as "Unknown block" - would be great to add that to the regex here as well

shunsukew:
> Hey @shunsukew - we had an urgent need for this layer. (thanks for doing the upfront work!)
> 
> I refactored it a bit to look like our existing layer. And fixed a few things I found in testing on top of the alchemy RPC.
> 
> Going to close this in favor of: #1218 (cherry-picked your commits so you'll still have attribution 🙏 )

Sure, thank you so much!

PR #1207, opened 2025-10-08
feat: use feature-aware map to avoid async wait piling up

andysim3d:
[Closes/Fixes] #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/crates/provider/src/alloy/da/local/nitro.rs
+++ b/crates/provider/src/alloy/da/local/nitro.rs
@@ -63,7 +63,7 @@ where
         gas_price: u128,
         extra_bytes_len: usize,
     ) -> ProviderResult<(u128, DAGasData, DAGasBlockData)> {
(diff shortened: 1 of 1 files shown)


crates/provider/src/alloy/da/local/nitro.rs
```diff
@@ -154,6 +123,15 @@ impl<AP> CachedNitroDAGasOracle<AP>
...
 {
+    async fn get_map_data(&self, block: BlockHashOrNumber) -> ProviderResult<NitroDAGasBlockData> {
+        let data = self
+            .block_data_cache
+            .try_get_with(block, async { self.get_block_data(block).await })
+            .await
+            .context(format!("failed to get block data {:?}", block))?;
+
+        Ok(data)
+    }
```
dancoombs, line 134: nit: newline after function

PR #1273, opened 2026-03-25
feat(undelegation): add method for users to undelegate from 7702 using sponsorship

0xfourzerofour:
### Summary                                                                                                                                                                                                  
                                                                                                                                                                                                           
  - Adds a rundler_sendSponsoredUndelegation RPC method that lets users submit a bundler-sponsored EIP-7702 undelegation transaction (clearing an EOA's delegation) without paying gas themselves          
  - Refactors signer management from permanent per-sender ownership to dynamic on-demand leasing, enabling signers to be shared across bundle sending and undelegation
  - Replaces per-sender pool subscriptions with a single shared broadcast channel to eliminate O(N²) fan-out overhead at large signer counts                                                               
                                                                                                                                                                                                           
### Motivation                                                                                                                                                                                               
                                                                                                                                                                                                           
  In the prior architecture each bundle sender permanently owned one signer key for its entire lifetime. This made it impossible to use any signer for out-of-band transactions (like sponsored            
  undelegation) because the SignerManager had zero available keys at runtime.
                                                                                                                                                                                                           
  Fixing this required decoupling sender slots from signer keys: senders now lease a key only when they have active work, releasing it during the idle wait between blocks. This opens a window every ~2s  
  per sender where a key is available for other operations.
                                                                                                                                                                                                           
  The same refactor also exposed an O(N²) pool subscription problem: each of N senders was independently subscribing to all N signer addresses. At 64 signers this meant 4,096 address update deliveries   
  per block; at 256 it would be 65,536. This is replaced by a single shared subscription broadcast to all senders.
                                                                                                                                                                                                           
 ### Proposed Changes                                                                                                                                                                                       

  Dynamic signer leasing (transaction_tracker.rs, bundle_sender.rs, task.rs)                                                                                                                               
  - TransactionTrackerImpl now holds Arc<dyn SignerManager> and Option<SignerLease> instead of a permanent SignerLease
  - Added begin_cycle() / end_cycle() to the TransactionTracker trait — begin_cycle acquires a signer and fetches fresh nonce/balance; end_cycle returns it to the pool when no transactions are pending   
  - The bundle sender loop explicitly calls end_cycle before blocking on a trigger and begin_cycle after, keeping keys in the pool during the idle wait                                                  
  - TransactionTrackerImpl::new is now synchronous (no eager reset() at construction)                                                                                                                      
  - sender_eoa is now updated each cycle from the leased signer rather than being fixed at startup                                                                                                         
                                                                                                                                                                                                           
  Shared pool subscription (bundle_sender.rs, task.rs)                                                                                                                                                     
  - Replaced N independent pool.subscribe_new_heads(all_addresses) calls with one shared subscription                                                                                                      
  - A new-heads-fanout task forwards Arc<NewHead> to a broadcast::channel, eliminating redundant heap allocations and pool-side work                                                                       
  - Pool subscription work goes from O(N²) to O(N); heap allocations per block drop by ~256× at 256 signers                                                                                              
                                                                                                                                                                                                           
  Sponsored undelegation (local.rs, server.rs, protos.rs, client.rs, rundler.rs)                                                                                                                           
  - New rundler_sendSponsoredUndelegation(auth: Eip7702Auth) -> B256 JSON-RPC method                                                                                                                       
  - New SendSponsoredUndelegation gRPC method on the builder service                                                                                                                                       
  - RPC layer validates: address == 0x0 (clearing delegation, not setting one), chain ID matches, signature is recoverable, EOA currently has code                                                         
  - Builder layer re-validates address == 0x0 so gRPC callers cannot bypass the check                                                                                                                      
  - Proto deserialization now uses TryFrom with explicit errors for malformed bytes and out-of-range y_parity rather than silently defaulting to zero values                                               
  - Undelegation acquires a signer from the shared pool, builds a type-4 transaction to the EOA (zero value, zero calldata, authorization list clears delegation), and returns the signer immediately after
   broadcast


crates/builder/src/server/local.rs
```diff
@@ -289,13 +335,96 @@ impl LocalBuilderServerRunner {
...
+
+        // Recover the EOA from the authorization signature so we can address the tx.
+        let eoa = auth
+            .recover_authority()
+            .context("failed to recover EOA from authorization signature")?;
+
+        // Build the type-4 (EIP-7702) transaction.
+        // The tx is sent to the EOA itself (zero-value, no calldata).
+        let tx = TransactionRequest::default()
+            .to(eoa)
```
dancoombs: I think this should be the bundler's sender address. It doesn't matter for undelegations, but it does for delegations.

When you delegate, the EOA then gets a contract, and when you call into it the `fallback()` function could run and revert the transaction.

By setting to to be our addresses, where we know the address doesn't have code and won't revert.

crates/builder/src/server/local.rs
```diff
@@ -289,13 +335,96 @@ impl LocalBuilderServerRunner {
...
+            .get_transaction_count(signer.address())
+            .await
+            .context("failed to get bundler nonce")?;
+
+        // Fetch current gas fees.
+        let (gas_fees, _base_fee) = self
+            .fee_estimator
+            .latest_bundle_fees()
+            .await
+            .context("failed to get bundle fees")?;
```
dancoombs: I think at a minimum this needs to bump the fees by 10% on each subsequent request to not get underpriced errors.

This is why I _really_ want to modify this to use the `bundle_sender` state machine, which has all of this logic (but will need some abstraction).

crates/builder/src/server/local.rs
```diff
@@ -289,13 +335,96 @@ impl LocalBuilderServerRunner {
...
+            .context("failed to sign delegation transaction")?;
+
+        let tx_hash = self
+            .evm_provider
+            .send_raw_transaction(raw_tx)
+            .await
+            .context("failed to submit delegation transaction")?;
+
+        tracing::info!("sent sponsored delegation for eoa {eoa} tx_hash {tx_hash}");
+        Ok(tx_hash)
```
dancoombs: I don't see any logic that actually ensures that this `tx_hash` eventually gets mined, so if the transaction is underpriced the delegation will get stuck in the mempool and eventually dropped. Also the signer will become stuck, the lease will get dropped, and the next time the signer is leased it will likely replace the transaction (dropping the delegation).

crates/builder/src/task.rs
```diff
@@ -432,15 +442,42 @@ where
     where
         T: TaskSpawnerExt,
     {
-        let mut bundle_sender_actions = vec![];
+        // One pool subscription covers all signer addresses. A single broadcast channel
+        // fans out new-head events to every bundle sender, keeping pool-side work O(N)
+        // instead of O(N²) (one subscription per sender × N addresses each).
```
dancoombs, line 465: I like this change in general, but just wondering if it was required for the logic in this PR?

crates/builder/src/bundle_sender.rs
```diff
@@ -230,6 +269,8 @@ where
             .record(value);
     }
 
+    /// Drive one full step of the state machine: wait for a trigger, then act.
+    /// Used by tests which call this directly.
```
dancoombs: Ideally we don't leave this as test only code, can we just mock this in tests?

crates/sim/tracer/package.json
```diff
@@ -14,5 +14,6 @@
     "@types/node": "^18.14.0",
     "prettier": "^2.8.4",
     "typescript": "^4.9.5"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
```
dancoombs: Do we need to commit this? I think it comes from corepack which we don't use

dancoombs (commented):
Overall, I think this works for a PoC, but runs into some pretty serious issues if our first attempt to land a transaction results in a stuck tx due to fee estimation or other reasons.

Biggest concern is that since this doesn't wait for the tx to land, the signer will get re-leased and the transaction will get dropped and replaced by a bundle and lost.

We also don't get the "auto-retry" behavior that users would expect in BSO, and they'll need to handle resubmissions if their txns stay pending.

Lets discuss offline the expectations of our initial user

crates/builder/src/task.rs
0xfourzerofour: Not strictly required for the delegation logic itself — each sender could still subscribe individually. It was included here because the signer-sharing change (dynamic lease/release cycle) made the O(N²) fan-out more visible: with N signers each idle-releasing every ~2s, each pool subscription fires N times per block update across all senders. The broadcast channel fix keeps that at O(N). Happy to split it into a separate PR if you'd prefer to review them independently.

dancoombs:
I think the fastest way to improve this will be to create a `delegation_sender` class that runs a simplified version of the state machine that is in `bundle_sender`.

I think it needs to:

1. Wait for a tx to land using the event stream that we already have - need to make sure the chain tracker will send events for these delegation only transactions
2. If the tx doesn't land for N blocks (same as bundle sender) drop and replace it
3. Bail after a certain number of fee bumps

This can be a lot simpler than the `bundle_sender` state machine as it doesn't need to deal with the UO pricing dynamics.

crates/builder/src/server/local.rs
0xfourzerofour: added this into the delegation_sender

crates/builder/src/bundle_sender.rs
```diff
@@ -1058,6 +1108,19 @@ impl<T: TransactionTracker, TRIG: Trigger> SenderMachineState<T, TRIG> {
      * Reset moves
      */
 
+    /// Returns true when the sender is idle: no pending transactions and waiting
+    /// for the next block trigger before attempting another bundle. This is the
+    /// only state in which the signer can be safely released.
+    fn is_waiting_for_trigger(&self) -> bool {
```
dancoombs: nit: we should rename this, there are other states that wait for the trigger but that shouldn't drop the signer (i.e. when a tx is pending)

crates/builder/src/task.rs
dancoombs: I'll review in this PR and we can keep as long as we don't see any issues. I'm not super concerned with 64 signers doing something every 2 seconds so we can revert this if we see problems in testing.

crates/builder/src/bundle_sender.rs
dancoombs: Bumping this, looks like we just removed the comments, but is this function used outside of tests? If not, we should remove it and modify tests

crates/builder/src/delegation_sender.rs
```diff
@@ -0,0 +1,269 @@
...
+                    // Fetch the receipt to disambiguate before bailing.
+                    if update.nonce.is_some_and(|n| n >= nonce) {
+                        if self.receipt_exists(tx_hash).await {
+                            info!("delegation tx {tx_hash} confirmed (via receipt) for eoa {eoa}");
+                            return Ok(tx_hash);
+                        }
+                        anyhow::bail!(
+                            "signer nonce {nonce} was consumed by another transaction \
+                             while awaiting delegation for eoa {eoa}"
+                        );
```
dancoombs, line 406: We should make sure this logs an error as this is a bad state. It would mean something isn't working properly in our signer leasing.

dancoombs (commented):
A couple of nits, and a question still open on test-only code.

We discussed offline moving this to a async model where send returns an ID and we have a status method. The status method will query the pool, which will check if a transaction has been mined containing a matching delegation ID. The pool will hold onto these for N blocks. If not in the mined map, it will query the builder to get pending status.

crates/builder/src/bundle_sender.rs
0xfourzerofour: renamed to is_signer_releasable

crates/builder/src/bundle_sender.rs
0xfourzerofour: now using the step_after_trigger function in the tests and removed the wrapper function

crates/builder/proto/builder/builder.proto
```diff
@@ -78,3 +83,53 @@ message BuilderError {
...
+    oneof result {
+        GetDelegationStatusSuccess success = 1;
+        BuilderError failure = 2;
+    }
+}
+
+enum DelegationStatusKind {
+    DELEGATION_STATUS_KIND_UNSPECIFIED = 0;
+    DELEGATION_STATUS_KIND_PENDING = 1;
+    DELEGATION_STATUS_KIND_MINED = 2;
```
dancoombs, line 128: Should we have an unknown status?

crates/rpc/src/types/mod.rs
```diff
@@ -415,3 +415,14 @@ pub struct RpcUserOperationGasPrice {
     /// Suggested fees with buffer for faster inclusion
     pub suggested: RpcSuggestedGasFees,
 }
+
+/// Status of a sponsored delegation request returned by `rundler_delegationStatus`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcDelegationStatus {
+    /// Current status: "pending", "mined", or "unknown".
+    pub status: String,
```
dancoombs: Can we make this a enum

crates/rpc/src/rundler.rs
```diff
@@ -94,6 +100,18 @@ pub trait RundlerApi {
...
     #[method(name = "getUserOperationGasPrice")]
     async fn get_user_operation_gas_price(&self) -> RpcResult<RpcUserOperationGasPrice>;
+
+    /// Sponsor an EIP-7702 delegation on behalf of the caller.
+    ///
+    /// The bundler pays the gas cost and immediately returns a delegation ID.
+    /// Poll `rundler_delegationStatus` with the returned ID to get the transaction hash.
+    /// To undelegate, set `address` to the zero address.
+    #[method(name = "sendSponsoredDelegation")]
+    async fn send_sponsored_delegation(&self, auth: Eip7702Auth) -> RpcResult<String>;
```
dancoombs: The ID is a hash right? If so we should use B256 for the return type

crates/builder/proto/builder/builder.proto
0xfourzerofour: I choose this because the buf linting made it so one of them needs to be "_UNSPECIFIED" and it felt a little weird having an UNKNOWN and UNSPECIFIED

crates/rpc/src/rundler.rs
0xfourzerofour: done

crates/rpc/src/types/mod.rs
0xfourzerofour: done

crates/builder/src/delegation_sender.rs
```diff
@@ -0,0 +1,442 @@
...
+        auth: Eip7702Auth,
+        heads_rx: broadcast::Receiver<Arc<NewHead>>,
+    ) -> anyhow::Result<B256> {
+        let signer = self.acquire_signer().await?;
+        let result = self.run(&signer, auth, heads_rx).await;
+        self.signer_manager.return_lease(signer);
+        result
+    }
+
+    async fn acquire_signer(&self) -> anyhow::Result<rundler_signer::SignerLease> {
```
dancoombs: Now that we have this `DelegationSenderTask` persistent object, thoughts on just turning it into a "mempool"?

I don't think its far off to make the change to:

1. Have `send` just add the requests to a queue and the pending map. Attempt to acquire a sender right away, but if none is available leave the request in the queue
2. Attempt to acquire senders after each block if there are pending requests in the queue.
3. Kick off the `DelegationSenderCore` once a sender is acquired

dancoombs: That would actually give is the ability to batch these when under load, as we only need to acquire 1 sender to be able to transmit N delegations

dancoombs: Up to a gas limit

crates/builder/proto/builder/builder.proto
dancoombs: In proto `_UNSPECIFIED` should never actually be used in an enum value, something to do with serde and backwards compatibility. It would be ideal to have `_UNKNOWN` as well here to represent that state explicitly.

crates/builder/src/delegation_sender.rs
dancoombs: We could even consider delaying each delegation N ms so that we can get more batching. Which is basically what our bundle sender does where it has a trigger every N ms to create a new bundle.

0xfourzerofour: not opposed to it, it would also give us a small reduction in DOS vector due to being able to send multiple at a time and not lock up all the keys

crates/builder/proto/builder/builder.proto
dancoombs: By never used I mean never set in the code, it needs to be in the proto definition

crates/builder/src/delegation_sender.rs
dancoombs: I think its worth a shot, the diff shouldn't be huge based on what is already here

0xfourzerofour: ok I updated this to use a mempool like design and use the predefined max_bundle_gas calculation to figure out how many auths can fit in the transaction

crates/builder/src/delegation_sender.rs
```diff
@@ -0,0 +1,470 @@
...
+/// into a single type-4 transaction, reducing signer contention and cutting
+/// the per-delegation tx-overhead when the system is under load.
+pub(crate) struct DelegationSenderTask<E, F> {
+    action_rx: mpsc::Receiver<DelegationSenderAction>,
+    completion_tx: mpsc::Sender<CompletionEvent>,
+    completion_rx: mpsc::Receiver<CompletionEvent>,
+    /// Incoming delegations waiting for a free signer.
+    queue: VecDeque<(DelegationId, Eip7702Auth)>,
+    /// Delegations that have been submitted and are waiting to mine.
+    pending: HashMap<DelegationId, ()>,
```
dancoombs: Any reason this isn't a `HashSet`?

crates/builder/src/delegation_sender.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// Create the task and its associated handle.
+    pub(crate) fn new(
+        signer_manager: Arc<dyn SignerManager>,
+        provider: E,
+        fee_estimator: F,
+        settings: Settings,
+        heads_tx: broadcast::Sender<Arc<NewHead>>,
+    ) -> (Self, DelegationSenderHandle) {
+        let (action_tx, action_rx) = mpsc::channel(64);
+        let (completion_tx, completion_rx) = mpsc::channel(64);
```
dancoombs: Any reason not to make these a decent amount larger? We have a lot of memory to work with

dancoombs (commented):
Nice this is clean! Can you make sure the docs are updated (specifically the RPC methods doc)? Then this is ready to ship.

crates/builder/src/delegation_sender.rs
0xfourzerofour: updated to set

crates/builder/src/delegation_sender.rs
0xfourzerofour: changed to 1024

dancoombs (approved):
🚢

PR #1280, opened 2026-04-23
feat(delegation): add valid_until field to delegation

0xfourzerofour:
## Proposed Changes

  - add valid_until field to delegations to allow for expiry if they stay for too long in the pool


dancoombs (approved):
🚢

PR #1282, opened 2026-05-05
feat(fallback): add fallback to raw

0xfourzerofour:
## Proposed Changes

  - Add fallback provider so that alternate providers fallback to raw sender if there are issues on the active path

### DIFF
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -18,8 +18,9 @@ use anyhow::Context;
 use clap::Args;
 use rundler_builder::{
     self, BloxrouteSenderArgs, BuilderEvent, BuilderEventKind, BuilderSettings, BuilderTask,
-    BuilderTaskArgs, EntryPointBuilderSettings, FlashbotsSenderArgs, LocalBuilderBuilder,
-    PolygonPrivateArgs, RawSenderArgs, TransactionSenderArgs, TransactionSenderKind,
+    BuilderTaskArgs, EntryPointBuilderSettings, FallbackSenderArgs, FlashbotsSenderArgs,
+    LocalBuilderBuilder, PolygonPrivateArgs, RawSenderArgs, TransactionSenderArgs,
+    TransactionSenderKind,
 };
 use rundler_pbh::PbhSubmissionProxy;
 use rundler_pool::RemotePoolClient;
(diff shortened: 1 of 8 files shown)


bin/rundler/src/cli/builder.rs
```diff
@@ -309,22 +321,34 @@ impl BuilderArgs {
...
             })),
             TransactionSenderKind::PolygonPrivate => {
-                Ok(TransactionSenderArgs::PolygonPrivate(PolygonPrivateArgs {
+                let primary = TransactionSenderArgs::PolygonPrivate(PolygonPrivateArgs {
                     submit_url: self.submit_url.clone().unwrap_or_else(|| rpc_url.into()),
+                });
+                Ok(TransactionSenderArgs::Fallback(FallbackSenderArgs {
+                    primary: Box::new(primary),
+                    fallback: Box::new(TransactionSenderArgs::Raw(raw_fallback())),
+                    recovery_interval,
```
dancoombs: There should be a way to configure this on/off. Maybe make `BUILDER_SENDER_RECOVERY_INTERVAL_SECS` optional and use its presence to toggle the fallback behavior on?

crates/builder/src/sender/mod.rs
```diff
@@ -66,6 +68,13 @@ pub(crate) enum TxSenderError {
     /// Insufficient funds for transaction
     #[error("insufficient funds for transaction")]
     InsufficientFunds,
+    /// Sender is unavailable due to an outage or unrecognized error.
+    ///
+    /// When a fallback sender is configured this triggers failover. The original
+    /// error is logged at WARN level at the point of classification so it can be
+    /// investigated if the attribution turns out to be wrong.
+    #[error("sender unavailable")]
+    SenderUnavailable,
```
dancoombs: Probably makes sense to pass the inner error inside of this one so it could be logged at a higher level if needed

crates/builder/src/sender/fallback.rs
```diff
@@ -0,0 +1,374 @@
...
+            .primary
+            .send_transaction(tx.clone(), expected_storage, signer)
+            .await
+        {
+            Ok(hash) => {
+                self.last_tx_used_fallback.store(false, Ordering::Release);
+                Ok(hash)
+            }
+            Err(TxSenderError::SenderUnavailable) => {
+                self.activate_fallback();
```
dancoombs: Do you think its worth adding a few retires before activating the fallback?

dancoombs (commented):
This is slick, great idea implementing this as another sender type.

My biggest logical comment is if we should have some sort of retry or error tolerance before activating the fallback. Transient errors are likely to happen from time to time that can be fixed with retires.

bin/rundler/src/cli/builder.rs
0xfourzerofour: added this

crates/builder/src/sender/fallback.rs
0xfourzerofour: added retries

dancoombs (approved):
LGTM! Need to look into the CI issues though

PR #1285, opened 2026-05-07
feat(nonce): cache nonce and balance

0xfourzerofour:
## Proposed Changes

  - Remove unnecessary RPC queries every block and cache value 
  - Get updates from block stream

### DIFF
--- a/crates/signer/src/manager.rs
+++ b/crates/signer/src/manager.rs
@@ -12,7 +12,7 @@
 // If not, see https://www.gnu.org/licenses/.
 
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt::{self, Debug},
     sync::Arc,
 };
@@ -299,8 +299,7 @@ impl SignerManager for FundingSignerManager {
             self.funding_notify.notify_one();
         }
 
-        let known: std::collections::HashSet<Address> =
-            self.signer_statuses.read().keys().copied().collect();
+        let known: HashSet<Address> = self.signer_statuses.read().keys().copied().collect();
         let mut cache = self.address_states.write();
         for update in updates {
             if !known.contains(&update.address) {
--- a/crates/signer/src/manager.rs
+++ b/crates/signer/src/manager.rs
@@ -36,6 +36,30 @@ use crate::{
     utils,
 };
 
+/// A snapshot of an address's chain state, sourced from new-head events.
+///
+/// `nonce` is `None` until a transaction from the address is observed in a
+/// new-head update; once observed it is monotonically tracked at the latest
+/// committed value.
+#[derive(Clone, Copy, Debug)]
+pub struct AddressStateSnapshot {
+    /// The latest known balance of the address.
+    pub balance: U256,
+    /// The latest committed nonce of the address, if one has been observed.
+    pub nonce: Option<u64>,
+}
+
+/// An address-state update derived from a new-head event.
+#[derive(Clone, Copy, Debug)]
+pub struct AddressStateUpdate {
+    /// The address.
+    pub address: Address,
+    /// The latest balance.
+    pub balance: U256,
+    /// A new committed nonce, if a transaction from this address mined in this block.
+    pub nonce: Option<u64>,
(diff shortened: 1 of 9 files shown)


crates/signer/src/manager.rs
```diff
@@ -250,6 +285,45 @@ impl SignerManager for FundingSignerManager {
...
+            }
+            let entry = cache.entry(update.address).or_insert(AddressStateSnapshot {
+                balance: update.balance,
+                nonce: None,
+            });
+            entry.balance = update.balance;
+            // Nonce is monotonic; only advance when the new head reports a
+            // committed value at least as high as what we've already seen.
+            if let Some(new_nonce) = update.nonce
+                && entry.nonce.is_none_or(|cached| new_nonce >= cached)
```
dancoombs, line 307: This might have issues with reorgs

crates/signer/src/manager.rs
dancoombs: Do we still call this?

crates/signer/src/manager.rs
0xfourzerofour: Addressed in f4498f4f. `reorg_depth` is now plumbed through `NewHead` (proto + types + pool publisher). The builder local server invokes `SignerManager::invalidate_nonce_cache()` whenever a head arrives with `reorg_depth > 0`, which clears all cached nonces (balances stay). On the next `begin_cycle` for any signer, `cached_state` returns `nonce: None` and we fall through to `reset()` (RPC), repopulating from chain state. The monotonic guard is still correct for the normal-block case where most addresses get `nonce: None` updates.

crates/signer/src/manager.rs
0xfourzerofour: Good catch — the only remaining caller was `init_balances` at startup. Removed the trait method in f4498f4f and routed the bootstrap path through `update_address_states` with `nonce: None` so all balance/nonce flow now goes through one entry point.

crates/signer/src/manager.rs
dancoombs: I'm concerned this will still cause a lot of RPC churn on networks where we see a lot of reorgs.

I think we need to make `reset()` (or the caller) lazy, we should only actually repopulate the nonce/balance when we're about to build a bundle. `begin_cycle` seems to be called much more than that.

0xfourzerofour: Agreed. Sketching the lazy version — want to confirm a couple of choices before I push:

**Plan**: `begin_cycle()` becomes a pure lease (no RPC, no cache lookup) and sets a `state_dirty` flag. New `async fn ensure_state()` does the existing cache-or-RPC dance only when dirty. Called at the top of `handle_building_state` / `handle_cancelling_state`. On reorg, bundle sender calls `tracker.mark_state_dirty()` in addition to `invalidate_nonce_cache()` so held-lease trackers re-fetch before their next build/cancel.

**Open questions:**

1. **Signer change in `begin_cycle`**: when the leased address differs from last cycle, clear `self.nonce`/`self.balance` to zero (rely on `ensure_state` to repopulate), or leave them and gate solely on `state_dirty`? I'd default to clearing — avoids any chance of a stale other-signer value being read.

2. **`process_update` during pending**: today it reads `self.nonce` to compare against `update.nonce`. If we held the lease through a reorg that rolled back our tx, `self.nonce` is stale (this is pre-existing, not introduced here). Gate `process_update` on `state_dirty` too (auto-refresh at the top), or leave it as today and let the next build/cancel pick it up?

3. **Cache hit semantics**: keep cache lookup inside `ensure_state` (so a hit still avoids RPC), or just go straight to `reset()` and drop the cache entirely? I'd keep the cache — it still helps when a signer cycles in/out of leasing across blocks without a reorg.

Default I'd ship: 1=clear, 2=gate, 3=keep cache. Sound right?

0xfourzerofour: mind the AI reply from my claude instance but the questions still the same

dancoombs: I think the default makes sense

PR #1290, opened 2026-06-03
fix(builder): don't release in-flight sender locks on Building tracker update

pavelm:
## Context

A single userOp was observed being submitted on-chain by **two different builders** within ~0.5s on a multi-builder deployment; one tx succeeded and the other reverted with `AA25 invalid account nonce`.

### Root cause

The builder-assigner's per-**sender** lock (`uo_sender_to_builder_state` in `assigner.rs`) is the only cross-builder coordination that stops the same userOp from being bundled by more than one builder. An op is **not** removed from the shared pool when it's submitted — only when it mines — so while a bundle tx is in-flight, the sender lock is the sole thing preventing a re-submit.

`step_after_trigger` released **all** of a builder's locks unconditionally on any tracker update while in the `Building` state:

```rust
InnerState::Building(building_state) => {
    if has_tracker_update {
        // Building state has no tracker-dependent cleanup, so release immediately.
        self.assigner.release_all(self.sender_eoa);
    }
    self.handle_building_state(state, building_state).await?;
}
```

`release_all` drops `Confirmed` locks too — including the lock for a bundle that was already submitted but not yet mined. A `Building` state is routinely re-entered with a tx still in-flight (the underpriced / replacement / condition-not-met retry paths in `handle_building_state`, which transition back to `Building`). When a tracker update then arrives, the confirmed lock is dropped while the op is still in-flight and still pool-resident, so another builder's `assign_work` claims and re-submits it → duplicate on-chain submission → the loser reverts with `AA25 invalid account nonce`.

Every per-result handler in `handle_building_state` already guards its release with `num_pending_transactions() == 0`; the top-level `Building` release was the one place missing that guard.

## Proposed Changes

  - Guard the `Building`-arm `release_all` in `step_after_trigger` with `num_pending_transactions() == 0`, so locks for in-flight transactions are retained until the tx resolves (mined/dropped/nonce-used), matching the convention used throughout `handle_building_state`.
  - Add `test_building_step_with_pending_tx_keeps_confirmed_locks`, a regression test that drives a `Building` step with a tracker update while a tx is in-flight and asserts the confirmed sender lock is not handed to another builder. (Verified it fails on the previous unconditional release — the second builder gets `Assigned` the in-flight op.)

## Notes

- Scope: any deployment with `num_signers > 1`; worse on short-block-time / high-throughput chains where the build cadence outpaces mining.
- The `Pending` / `Cancelling` / `CancelPending` arms are intentionally unchanged — `Pending` already defers release until after the update is handled (terminal updates only), and the cancel arms release on terminal cancellation updates.
- `cargo test -p rundler-builder --lib` (92 passed) and `cargo clippy -p rundler-builder` are clean.

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -281,8 +281,16 @@ where
 
         match state.inner {
             InnerState::Building(building_state) => {
-                if has_tracker_update {
-                    // Building state has no tracker-dependent cleanup, so release immediately.
+                // Only release locks when no transaction is in-flight. A Building state can be
+                // (re-)entered while a previously sent bundle is still pending (e.g. the
+                // underpriced / replacement / condition-not-met retry paths in
+                // `handle_building_state`). Releasing a sender whose userOp is still in-flight —
+                // submitted but not yet mined, and therefore still served by the shared pool —
+                // lets another builder re-assign and re-submit the same op, producing a duplicate
+                // on-chain submission where one tx reverts with AA25 invalid account nonce. This
+                // mirrors the `num_pending_transactions() == 0` guard used by every per-result
+                // handler in `handle_building_state`.
+                if has_tracker_update && state.transaction_tracker.num_pending_transactions() == 0 {
                     self.assigner.release_all(self.sender_eoa);
                 }
                 self.handle_building_state(state, building_state).await?;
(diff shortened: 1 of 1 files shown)


dancoombs (approved):
I think this is a fine defensive check to merge, but unlikely to fix the issue we saw.

`has_tracker_update` is only true if we get an update for the EOA controlled by this sender. This should only happen when (1) tx mines (2) nonce changes - both of which should drop all pending txns because we only ever have on pending nonce.

That said, this check could prevent bugs in the future if this changes so is a reasonable thing to add.

PR #1292, opened 2026-06-03
fix(builder): recover from Cronos invalid nonce errors

jakehobbs:
[Closes/Fixes] #

## Proposed Changes

  - Treat Cronos/Cosmos SDK `invalid nonce; got ...` RPC errors as nonce-too-low in the builder sender classifier.
  - Route those failures through the existing nonce refresh and bundle retry path instead of the generic unexpected-error path.
  - Add regression tests for the Cronos message, case-insensitive matching, and avoiding a bare `invalid sequence` false positive.


dancoombs (dismissed):
🚢

pavelm (approved):

PR #1293, opened 2026-06-10
fix(builder): cache signer nonce/balance to avoid per-block RPC reads

pavelm:
## Problem

Internal Compute Unit usage on the rundler fleet spiked ~4x starting early April. Root cause traced to the signer-sharing rework in #1273 (`feat(undelegation)`, merged Apr 2):

- The bundle-sender loop now **releases its signer to the shared pool whenever idle** and re-leases it on the next head, so the key can be borrowed for sponsored undelegation.
- `begin_cycle()` → `reset()` unconditionally fetches **nonce + balance from chain on every re-lease**.
- Since the loop wakes on every new head, each **idle builder makes 2 RPC calls per block regardless of userop traffic**.
- Across ~167 prod builders (`num_builders_v0_6 + v0_7` in chain-config), that's millions of internal node calls/day of brand-new traffic — the dominant driver of the CU step.

## Fix

Add a per-address account-state cache to `SignerManager`:

- `begin_cycle()` seeds nonce/balance from the cache and **skips the chain `reset()`** when a valid entry exists. This is the per-block hot path — re-acquiring a signer is now free.
- `reset()` (chain read) and `process_update()` (confirmed nonce) write fresh state back to the cache.
- `update_balances` (already invoked per block from the head broadcast) keeps the cached balance current; it only refreshes existing entries, so it never fabricates a "known nonce".
- The delegation sender advances the nonce outside the tracker, so it **invalidates the cached state on lease return**.

Idle builders drop from 2 RPC calls/block to ~0. The chain read fires only on genuine first acquisition, an error-recovery reset, or right after an undelegation.

## Correctness

Only a lease holder can advance an account's nonce, and the bundle sender releases the signer **only when no transactions are pending** — so a cached nonce stays valid across release/re-acquire until another consumer leases the address and sends from it (which invalidates the cache). Balance self-heals per block via `update_balances`. Error-recovery resets (`requires_reset`, `NonceTooLow`, nonce mismatch) are untouched and still read from chain.

## Testing

- `cargo build` + `cargo clippy` clean on `rundler-signer` and `rundler-builder`.
- 93 builder + 7 signer unit tests pass.
- New `test_begin_cycle_uses_cache_after_release` asserts a second `begin_cycle` after `end_cycle` makes **zero** additional `get_transaction_count`/`get_balance` calls (enforced with `.times(1)` mock expectations).

⚠️ Not yet validated against a live node — recommend a staging soak watching internal CU before rollout.

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -281,8 +281,16 @@ where
 
         match state.inner {
             InnerState::Building(building_state) => {
-                if has_tracker_update {
-                    // Building state has no tracker-dependent cleanup, so release immediately.
+                // Only release locks when no transaction is in-flight. A Building state can be
+                // (re-)entered while a previously sent bundle is still pending (e.g. the
+                // underpriced / replacement / condition-not-met retry paths in
+                // `handle_building_state`). Releasing a sender whose userOp is still in-flight —
+                // submitted but not yet mined, and therefore still served by the shared pool —
+                // lets another builder re-assign and re-submit the same op, producing a duplicate
+                // on-chain submission where one tx reverts with AA25 invalid account nonce. This
+                // mirrors the `num_pending_transactions() == 0` guard used by every per-result
+                // handler in `handle_building_state`.
+                if has_tracker_update && state.transaction_tracker.num_pending_transactions() == 0 {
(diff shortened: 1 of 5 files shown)


dancoombs (approved):
LGTM

PR #1294, opened 2026-06-12
feat(rpc): add block option parameter to user operation lookup methods

dancoombs:
## Summary

Adds a non-standard optional positional parameter to `eth_getUserOperationByHash`, `eth_getUserOperationReceipt`, and `rundler_getUserOperationStatus` that scopes the `eth_getLogs` search for the `UserOperationEvent` to a caller-supplied block range (`{fromBlock, toBlock}`) or block hash.

This is useful when the caller already knows where the operation was mined (cheaper, targeted `getLogs`) or needs to look up operations older than the configured `user_operation_event_block_distance` window.

## Behavior

- **Omitted**: unchanged — search covers the last `user_operation_event_block_distance` blocks (all blocks if unset), with the existing fallback retry.
- **Supplied**: the block option is used as-is for the `getLogs` filter, and the `user_operation_event_block_distance_fallback` retry is disabled so caller-supplied ranges are never silently rewritten.
- `eth_getUserOperationReceipt` remains backwards compatible with its existing `LATEST`/`PENDING` block tag parameter via an untagged union (`RpcBlockOptionOrTag`).

## Notes

- Wire format uses `rename_all_fields = "camelCase"` so clients send standard `fromBlock`/`toBlock` fields.
- Added a unit test in the existing `api.rs` test module verifying the caller's block option reaches the `getLogs` filter (both range and block-hash variants) and that the latest-block lookup is skipped.
- Documented the new parameter in `docs/architecture/rpc.md`.

## Compatibility notes

- Adding a positional parameter means values in that slot are now parsed instead of ignored: a client that previously sent a surplus 2nd positional param to `eth_getUserOperationByHash` or `rundler_getUserOperationStatus` (previously silently dropped by jsonrpsee) will now either get `-32602` for unparseable values, or — if the value happens to parse (e.g. a 32-byte hex string is a valid block hash) — have it interpreted as a block option that scopes the search. This is inherent to adding any trailing parameter and matches the precedent of the existing `permissions`/`state_override` params.
- Everything else is verified unchanged for existing clients: the legacy `eth_getUserOperationReceipt` tag shapes (`"LATEST"`/`"latest"`/`"PENDING"`/`"pending"`/`null`/omitted, positional and named), the positional `permissions` wire format, the default search window and fallback retry, and full header inertness when `--rpc.permissions_enabled` is off.

### DIFF
--- a/crates/rpc/src/eth/events/common.rs
+++ b/crates/rpc/src/eth/events/common.rs
@@ -344,27 +344,8 @@ where
                     ));
                 };
 
-                let from = match from {
-                    BlockNumberOrTag::Number(from) => from,
-                    _ => {
-                        return Err(EventProviderError::InvalidRequest(format!(
-                            "max event block distance set to {max_distance}, fromBlock cannot be block tag"
-                        )));
-                    }
-                };
-
-                let to = match to {
-                    BlockNumberOrTag::Number(to) => to,
-                    _ => {
-                        let Some(block) = self.provider.get_block(BlockId::Number(to)).await?
-                        else {
-                            return Err(EventProviderError::Provider(ProviderError::Other(
-                                anyhow::anyhow!("provider returned null block"),
-                            )));
-                        };
-                        block.number()
-                    }
-                };
+                let from = self.resolve_block_number(from).await?;
+                let to = self.resolve_block_number(to).await?;
 
                 if to.saturating_sub(from) > max_distance {
                     return Err(EventProviderError::InvalidRequest(format!(
@@ -591,5 +572,19 @@ where
         }
 
         Ok(None)
+    }
+
+    async fn resolve_block_number(&self, block: BlockNumberOrTag) -> EventProviderResult<u64> {
+        match block {
+            BlockNumberOrTag::Number(block) => Ok(block),
+            _ => {
+                let Some(block) = self.provider.get_block(BlockId::Number(block)).await? else {
+                    return Err(EventProviderError::Provider(ProviderError::Other(
+                        anyhow::anyhow!("provider returned null block"),
+                    )));
+                };
+                Ok(block.number())
+            }
+        }
     }
 }
--- a/crates/rpc/src/eth/events/common.rs
+++ b/crates/rpc/src/eth/events/common.rs
@@ -28,7 +28,9 @@ use rundler_types::{
 use rundler_utils::log::LogOnError;
 use tracing::instrument;
 
-use super::{EventProviderError, EventProviderResult, UserOperationEventProvider};
+use super::{
+    EventBlockOptions, EventProviderError, EventProviderResult, UserOperationEventProvider,
+};
 use crate::types::{RpcUserOperationByHash, RpcUserOperationReceipt, UOStatusEnum};
 
 #[derive(Debug)]
@@ -73,12 +75,11 @@ where
     async fn get_mined_by_hash(
         &self,
         hash: B256,
-        block_option: Option<FilterBlockOption>,
-        max_block_range: Option<u64>,
+        block_options: EventBlockOptions,
     ) -> EventProviderResult<Option<RpcUserOperationByHash>> {
         // Get event associated with hash (need to check all entry point addresses associated with this API)
         let event = self
-            .get_event_by_hash(hash, block_option, max_block_range)
+            .get_event_by_hash(hash, block_options)
             .await
             .log_on_error("should have successfully queried for user op events by hash")?;
 
@@ -115,11 +116,10 @@ where
     async fn get_receipt(
         &self,
         hash: B256,
-        block_option: Option<FilterBlockOption>,
-        max_block_range: Option<u64>,
+        block_options: EventBlockOptions,
     ) -> EventProviderResult<Option<RpcUserOperationReceipt>> {
         let event = self
-            .get_event_by_hash(hash, block_option, max_block_range)
+            .get_event_by_hash(hash, block_options)
             .await
             .log_on_error("should have successfully queried for user op events by hash")?;
         let Some(event) = event else { return Ok(None) };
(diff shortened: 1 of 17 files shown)


jakehobbs (dismissed):

crates/rpc/src/eth/events/common.rs
```diff
@@ -314,30 +318,52 @@ where
...
-            Some(distance) => to_block.saturating_sub(distance),
-            None => 0,
+    async fn get_event_by_hash(
+        &self,
+        hash: B256,
+        block_option: Option<FilterBlockOption>,
+    ) -> EventProviderResult<Option<Log>> {
+        let has_block_option = block_option.is_some();
+        let block_option = if let Some(bo) = block_option {
+            bo
```
dancoombs, line 382: This needs to enforce a max range of event_block_distance when set. Probably should happen in the RPC handler for a nice error message.

jakehobbs (dismissed):

crates/rpc/src/eth/events/common.rs
```diff
@@ -314,30 +318,98 @@ where
...
+                let (Some(from), Some(to)) = (from_block, to_block) else {
+                    return Err(EventProviderError::InvalidRequest(
+                        "fromBlock and toBlock must both be populated in event request".to_string(),
+                    ));
+                };
+
+                let from = match from {
+                    BlockNumberOrTag::Number(from) => from,
+                    _ => {
+                        return Err(EventProviderError::InvalidRequest(format!(
```
dancoombs: For consistency we might want to resolve this the same way we resolve `to`

crates/rpc/src/eth/events/common.rs
```diff
@@ -314,30 +321,83 @@ where
...
+        let has_block_option = block_option.is_some();
+        let block_option = if let Some(bo) = block_option {
+            if let FilterBlockOption::Range {
+                from_block,
+                to_block,
+            } = bo
+                && let Some(max_distance) = event_block_distance
+            {
+                let (Some(from), Some(to)) = (from_block, to_block) else {
+                    return Err(EventProviderError::InvalidRequest(
```
jakehobbs: When this caller-supplied range is rejected here, the error currently comes back to the client as a JSON-RPC internal error because `From<EventProviderError>` only special-cases `Provider` and maps every other variant through `EthRpcError::Internal`. Since this branch is validating user input, `InvalidRequest` should be converted to invalid params so too-wide or incomplete block ranges return `-32602` instead of `-32603`.

posted by claude 🤖

crates/rpc/src/types/mod.rs
```diff
@@ -430,3 +431,46 @@ pub enum RpcDelegationStatus {
     /// Delegation status is unknown.
     Unknown,
 }
+
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[serde(untagged, rename_all_fields = "camelCase")]
+pub(crate) enum RpcBlockOption {
+    Range {
+        /// The block number or tag this filter should start at.
+        from_block: BlockNumberOrTag,
```
jakehobbs: The docs for this new parameter say `fromBlock` and `toBlock` are optional, but this variant requires both fields for `eth_getUserOperationByHash` and `rundler_getUserOperationStatus`. A caller using a documented one-sided range like `{ "fromBlock": "0x64" }` will fail deserialization before the provider sees it, so either these fields need to be `Option` here too or the docs should not promise partial ranges. 

posted by claude 🤖

crates/rpc/src/eth/events/common.rs
dancoombs: Fixed in b225335f — `EventProviderError::InvalidRequest` now maps to `EthRpcError::InvalidParams`, so invalid caller-supplied ranges (too wide, missing a bound while a max is enforced, or reversed) return `-32602`. Added a test covering all three cases.

crates/rpc/src/types/mod.rs
dancoombs: Fixed in b225335f — `RpcBlockOption::Range` fields are now `Option<BlockNumberOrTag>`, matching the docs and `RpcBlockOptionOrTag`. Partial ranges work when no max block distance is enforced; when one is enforced they're rejected with invalid params, and the docs now state that both bounds are required in that case.

crates/rpc/src/eth/events/common.rs
```diff
@@ -72,10 +73,12 @@ where
     async fn get_mined_by_hash(
         &self,
         hash: B256,
+        block_option: Option<FilterBlockOption>,
+        max_block_range: Option<u64>,
```
dancoombs: Should we consider moving max_block_range into FilterBlockOption?

crates/rpc/src/eth/events/common.rs
```diff
@@ -314,30 +321,89 @@ where
...
+                && let Some(max_distance) = event_block_distance
+            {
+                let (Some(from), Some(to)) = (from_block, to_block) else {
+                    return Err(EventProviderError::InvalidRequest(
+                        "fromBlock and toBlock must both be populated in event request".to_string(),
+                    ));
+                };
+
+                let from = super::resolve_block_number(&self.provider, from).await?;
+                let to = super::resolve_block_number(&self.provider, to).await?;
```
dancoombs: Should we be using resolve block option here?

crates/rpc/src/eth/events/common.rs
dancoombs: Done in 48b6c8b3 — `FilterBlockOption` is alloy's type so we can't extend it, but the pair is now bundled into a single `EventBlockOptions { block_option, max_block_range }` struct threaded through the router and event provider traits. `max_block_range` also moved out of `RpcPermissions` into its own `MaxBlockRange` request extension, since it's lookup configuration rather than a UO permission — this also stops the header from wiping out positional permissions on `eth_sendUserOperation`.

crates/rpc/src/eth/events/common.rs
dancoombs: Done in 48b6c8b3 — `get_event_by_hash` now calls `resolve_block_option` on the whole option up front (a no-op when the handler already resolved tags), then validates against the resolved numbers. As a bonus this makes the `fromBlock > toBlock` check apply whether or not a max distance is enforced.

pavelm (approved):

PR #1297, opened 2026-06-25
refactor(provider): use tower timeout for provider requests

### LINKED ISSUE #944: use tower::timeout instead a copied one
**Describe the feature**
Follow up of #926, we now use a copied of tower::timeout due to type trait restriction is different. should revisit and use the tower::timeout as much as possible. 

**Additional context**

pengqima:
## Summary

- replace the copied provider timeout future with `tower::timeout::TimeoutLayer`
- keep the existing `ProviderTimeoutLayer` wrapper so provider construction remains unchanged
- map `tower::timeout::Elapsed` back to the existing client-side timeout `TransportError`
- remove the now-unused `pin-project` dependency from `rundler-provider`

## Motivation

This addresses #944 by reusing Tower's timeout middleware instead of maintaining a local copy of the timeout future. The small wrapper is kept only to preserve Rundler's existing `TransportError` mapping and metrics behavior for client-side provider timeouts.

## Testing

- `cargo +nightly fmt --all --check`
- `git diff --check`
- compiled the timeout adapter in a minimal check crate pinned to `alloy-json-rpc = 1.3.0`, `alloy-transport = 1.3.0`, and `tower = 0.4.13`

Fixes #944


pengqima:
@pavelm @dancoombs Hi, Could you please review this PR at your convenience? Thank you very much.

dancoombs:
Thanks for the cleanup! Since the only reason the wrapper still exists is alloy's `Transport` contract (`Error = TransportError`) vs. tower's `Timeout` erasing to `BoxError`, I think we can go one step further and drop the custom `Layer`/`Service` types entirely — tower already ships the adapter as `map_err`, and `ServiceBuilder` itself implements `Layer`, so alloy's `ClientBuilder::layer()` accepts it as-is:

```rust
// in new_alloy_provider (mod.rs) — the four-arm match is unchanged
let timeout_layer = ServiceBuilder::new()
    .map_err(map_timeout_error)
    .timeout(Duration::from_secs(config.client_timeout_seconds));
```

with just the error-mapping function kept (can live in `mod.rs`, deleting `provider_timeout.rs`):

```rust
/// Maps errors from a [`tower::timeout::Timeout`]-wrapped transport back to
/// [`TransportError`], as required by alloy's `Transport` contract.
fn map_timeout_error(error: BoxError) -> TransportError {
    match error.downcast::<Elapsed>() {
        Ok(_) => TransportError::local_usage_str("provider request timeout from client side"),
        Err(error) => match error.downcast::<TransportError>() {
            Ok(error) => *error,
            // Unreachable in practice: the inner transport's error type is
            // `TransportError`, so the box holds either that or `Elapsed`.
            Err(error) => TransportErrorKind::custom_str(&error.to_string()),
        },
    }
}
```

Only other change needed is adding `"util"` to the tower features in the workspace `Cargo.toml` (currently `["timeout"]`).

A few notes:

- **Zero runtime cost from the builder**: `ServiceBuilder`'s `Layer` impl has `type Service = L::Service` and just delegates to its inner `Stack`, so the compiled service is exactly `MapErr<Timeout<Http>, fn>` — identical to hand-composing the layers. The builder doesn't exist in the resulting stack.
- **Avoids the per-request boxing**: this PR's `type Future = BoxFuture<...>` + `.boxed()` adds a heap allocation per RPC request; `MapErr` is a concrete future type, so this version stays allocation-free like the original code.
- **Behavior-identical for timeouts**: same `Elapsed` → `"provider request timeout from client side"` local-usage error, same pass-through of inner `TransportError`s. One deliberate tweak: the unreachable fallback arm uses `TransportErrorKind::custom_str` rather than `local_usage_str`, so an unknown error wouldn't be mislabeled as a local usage error.

I compile-verified this shape against `main` (`cargo check -p rundler-provider` + `cargo +nightly fmt`) — net diff ends up around +24/−134.

Separately: would be great to add a small `#[tokio::test(start_paused = true)]` covering (a) a never-resolving service → the timeout error, and (b) a `TransportError` passing through the downcast intact — the downcast chain is the one piece that could silently break if tower/alloy change their error types.

dancoombs:
@pengqima also make sure you're using signed commits as we'll only be able to merge those.

pengqima:
> Thanks for the cleanup! Since the only reason the wrapper still exists is alloy's `Transport` contract (`Error = TransportError`) vs. tower's `Timeout` erasing to `BoxError`, I think we can go one step further and drop the custom `Layer`/`Service` types entirely — tower already ships the adapter as `map_err`, and `ServiceBuilder` itself implements `Layer`, so alloy's `ClientBuilder::layer()` accepts it as-is:

Updated, thanks for the detailed suggestion.

I removed the custom Layer/Service wrapper and now compose the timeout directly with ServiceBuilder::map_err(...).timeout(...), keeping only the error-mapping function in mod.rs. This should avoid the per-request BoxFuture allocation while preserving the existing timeout behavior.

I also added the tower "util" feature and added start_paused tests covering:
- Elapsed timeout errors being mapped to the client-side timeout TransportError
- inner TransportError values passing through the downcast path intact

Verified with:
- cargo check -p rundler-provider
- cargo test -p rundler-provider --features test-utils timeout_layer_

pengqima:
> @pengqima also make sure you're using signed commits as we'll only be able to merge those.

Signed. Please review it again.

dancoombs (approved):

Issue #1301, opened 2026-07-06
Does not conform to EIP-7562 [OP-062]


shawn-zil:
**Describe the bug**
According to EIP7562 [OP-062]
```
[OP-062] Precompiles:
    Only allow known accepted precompiles on the network, that do not access anything in the blockchain state or environment.
    The core precompiles 0x1 .. 0x11
    The RIP-7212 secp256r1 precompile, on networks that accepted it.
```

**To reproduce**
Call a BLS12-381 precompile in `validateUserOp()` will result in:
```
error code -32502: account uses banned precompile: 0x000000000000000000000000000000000000000d in contract xxxx
```

**Expected behavior**
That error should not happen as precompile 0xd is allowed in OP-062.

shawn-zil:
A 1-line change will fix this here: https://github.com/alchemyplatform/rundler/blob/8a218ee3fa9c73580c78a14d7cbf1b67f61bdd17/crates/sim/tracer/src/validationTracerV0_7.ts#L348

dancoombs:
Fixed with #1302

PR #1302, opened 2026-07-06
fix: allow OP-062 core precompiles.

shawn-zil:
Refs: #1301

[Closes/Fixes] #1301

## Proposed Changes
  - Allow the full set of core precompiles in EIP7562.


dancoombs (approved):
Thanks for the fix!

PR #1304, opened 2026-07-13
fix(builder): remove ops after internal RPC error

dancoombs:
Issue: N/A

## Proposed Changes

- Preserve unrecognized RPC error details and derive the submitted transaction hash from the signed transaction bytes.
- On a terminal `-32000: internal error`, log the transaction hash and remove every user operation selected for that bundle from the mempool.
- Preserve fallback sender behavior and leave bundle operations untouched for other unrecognized RPC errors.

## Root Cause

Unrecognized provider RPC errors were flattened into a generic error before reaching the bundle sender. That discarded the locally derivable transaction hash and left the bundle sender without a structured signal for cleaning up the affected user operations.

## Impact

Ambiguous internal RPC submission failures no longer leave the same user operations eligible for repeated bundling, and operators can correlate the failure with the exact transaction hash.

## Validation

- `cargo test -p rundler-builder` (100 tests passed)
- `cargo clippy -p rundler-builder --all-features --tests -- -D warnings`
- `cargo +nightly fmt --all --check`
- Repository pre-commit hooks: workspace clippy, rustfmt, buf, and cargo-sort

### DIFF
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -1019,6 +1019,8 @@ where
                 tx_hash,
             }) => {
                 if Self::is_internal_rpc_error(code, &message) {
+                    // Treat this error shape as terminal. Unlike an onchain bundle revert, do not
+                    // retry the bundle's operations.
                     let uo_hashes: Vec<B256> = ops.iter().map(|(_, hash)| *hash).collect();
                     let uo_count = uo_hashes.len();
                     error!(
@@ -1027,7 +1029,7 @@ where
                         rpc_error_code = code,
                         rpc_error_message = %message,
                         uo_count,
-                        "Bundle transaction failed to land due to internal RPC error; removing all user operations from the pool"
+                        "Bundle submission hit a terminal internal RPC error; removing all user operations from the pool"
                     );
 
                     let rejection_error =
@@ -1038,7 +1040,7 @@ where
                             BuilderEvent::rejected_op(
                                 self.builder_tag.clone(),
                                 *uo_hash,
-                                OpRejectionReason::BundleTransactionFailedToLand {
+                                OpRejectionReason::TerminalBundleSubmissionError {
                                     tx_hash,
                                     error: Arc::clone(&rejection_error),
                                 },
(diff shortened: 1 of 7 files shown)


crates/builder/src/bundle_sender.rs
```diff
@@ -998,6 +1013,44 @@ where
...
+                if Self::is_internal_rpc_error(code, &message) {
+                    let uo_hashes: Vec<B256> = ops.iter().map(|(_, hash)| *hash).collect();
+                    let uo_count = uo_hashes.len();
+                    error!(
+                        %tx_hash,
+                        ?entry_point,
+                        rpc_error_code = code,
+                        rpc_error_message = %message,
+                        uo_count,
+                        "Bundle transaction failed to land; removing all user operations from the pool"
```
dancoombs: Update this to be clear this is the "internal error" shape that cause the bundle transaction to fail to land

crates/builder/src/bundle_sender.rs
```diff
@@ -998,6 +1013,44 @@ where
...
+                    );
+                    if let Err(remove_error) = self.pool.remove_ops(entry_point, uo_hashes).await {
```
dancoombs, line 1051: We need to emit a reason why these ops are being removed for tracing. Look at other places this happens. At a minimum we need to see the UO hash that is being removed and why.

crates/builder/src/sender/raw.rs
```diff
@@ -43,19 +43,24 @@ where
...
         } else {
-            self.submit_provider.send_raw_transaction(raw_tx).await?
+            self.submit_provider
+                .send_raw_transaction(raw_tx)
+                .await
+                .map_err(TxSenderError::from)
         };
 
-        Ok(tx_hash)
+        send_result.map_err(|error| error.with_tx_hash(signed_tx_hash))
```
Richard-Dang, line 63: Cancel still uses `?` without `with_tx_hash`, so unrecognized errors land as `Other` and the new cancel `UnrecognizedRpc` arm never fires. Should we mirror this there too (same in `polygon_private.rs`)?

```rust
let signed_tx_hash = keccak256(&raw_tx);
self.submit_provider
    .send_raw_transaction(raw_tx)
    .await
    .map_err(TxSenderError::from)
    .map_err(|error| error.with_tx_hash(signed_tx_hash))
    .map(|tx_hash| CancelTxInfo {
        tx_hash,
        soft_cancelled: false,
    })
```

Richard-Dang (approved):

jakehobbs (approved):

PR #1305, opened 2026-07-13
fix(provider): filter handled submission errors from metrics

dancoombs:
Issue: N/A

## Proposed Changes

- Centralize classification of transaction submission errors that Rundler handles itself.
- Exclude those handled outcomes from the generic Alloy provider RPC response-status metric for raw, conditional, and private transaction submission methods.
- Preserve generic RPC error metrics for unknown submission errors, transport failures, and non-submission methods.

## Root Cause

The provider metrics middleware classified every JSON-RPC error in the `-32000` range as a server error before the builder converted known submission outcomes such as nonce-too-low and replacement-underpriced into retryable or otherwise handled states. This caused expected transaction lifecycle races to contribute to provider RPC alerts.

## Impact

Handled submission outcomes no longer inflate provider RPC error alerts. Request count, latency, HTTP status, warning logs, and the builder's dedicated outcome counters remain unchanged.

## Validation

- `cargo test -p rundler-provider -p rundler-builder` (120 tests passed; 10 existing provider tests ignored)
- Repository pre-commit hooks: workspace Clippy, nightly rustfmt, Buf, and cargo-sort
- `cog check v0.11.0..HEAD`

### DIFF
--- a/crates/provider/src/alloy/metrics.rs
+++ b/crates/provider/src/alloy/metrics.rs
@@ -22,6 +22,8 @@ use rundler_types::task::{
 };
 use tower::{Layer, Service};
 
+use crate::transaction;
+
 /// Alloy provider metric layer.
 #[derive(Default)]
 pub(crate) struct AlloyMetricLayer {}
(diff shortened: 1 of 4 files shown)


crates/provider/src/alloy/metrics.rs
```diff
@@ -95,7 +97,11 @@ where
             match &response {
                 Ok(resp) => {
                     method_logger.record_http(HttpCode::TwoHundreds);
-                    method_logger.record_rpc(get_rpc_status_code(resp));
+                    if resp.as_error().is_none_or(|error| {
+                        should_record_rpc_status(&method_name, &error.message, error.code)
```
dancoombs: I think instead of not recording when we see a known submission error, we should map to 0 or a known code that won't trigger our alerting.

dancoombs: Done — handled submission errors are now recorded as `RpcCode::Success` (code 0) instead of being skipped, so the datapoint is kept but won't trigger alerting.

pavelm (approved):

PR #1308, opened 2026-07-15
fix(builder): re-validate EIP-7702 auth nonces at bundle assembly

dancoombs:
## Proposed Changes

Closes the EIP-7702 staleness window behind the empty-revert bundle livelock incident (fix 2 from the investigation; fix 1, no-data revert handling, landed in #1300).

Auth tuple nonces are only validated at pool entry, and every later simulation bypasses real 7702 semantics via an unconditional code state override — so a tuple that goes stale in the pool reaches the bundle-validation `eth_call`, where the node skips it and the codeless sender reverts the whole `handleOps` with empty, unattributable revert data.

  - At bundle proposal time, batch-fetch the transaction count and delegation status (`eth_getTransactionCount` + `eth_getCode`) of every sender carrying an auth tuple. Runs concurrently with re-simulation and the paymaster balance fetch, so no added latency and no extra RPCs for bundles without 7702 ops.
  - In `assemble_context`, reject an op whose tuple nonce is stale unless the sender is already delegated to the tuple's address. Rejection uses the normal `rejected_ops` path, so only the poison op is removed from the pool — innocent ops in the same bundle proceed.
  - New `OpRejectionReason::StaleEip7702Auth` event variant and `builder_proposer_stale_eip7702_auth_rejections` counter for observability.
  - New `authorization_utils::eip7702_delegate_from_code` helper that strictly parses the `0xef0100 ‖ address` delegation designator.

Verdict table for an op carrying an auth tuple:

| Chain state | Outcome |
|---|---|
| Tuple nonce == sender tx count | Keep (tuple still valid) |
| Stale nonce, sender delegated to tuple's address | Keep (delegation already landed) |
| Stale nonce, sender codeless or delegated elsewhere | Reject just that op |

Tests cover all three verdicts plus the multi-op case (poison op rejected, innocent op still bundled) and designator parsing edge cases.

### DIFF
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -47,7 +47,9 @@ use rundler_types::{
     pool::{PoolOperation, SimulationViolation},
     proxy::SubmissionProxy,
 };
-use rundler_utils::{emit::WithEntryPoint, eth, guard_timer::CustomTimerGuard, math};
+use rundler_utils::{
+    authorization_utils, emit::WithEntryPoint, eth, guard_timer::CustomTimerGuard, math,
+};
 use tokio::sync::broadcast;
 use tracing::{debug, error, info, warn};
 
@@ -267,6 +269,12 @@ where
         );
 
         // (3) simulate ops
+        let eip7702_senders = ops
+            .iter()
+            .filter(|op| op.op.uo.authorization_tuple().is_some())
+            .map(|op| op.op.uo.sender())
+            .collect::<HashSet<Address>>();
+
         let simulation_futures = ops
             .into_iter()
             .map(|op| self.simulate_op(op, block_hash))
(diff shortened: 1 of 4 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -1231,6 +1270,34 @@ where
         }
     }
 
+    // Fetch the current transaction count and delegation status of each EIP-7702
+    // sender. Authorization tuple nonces are only checked against chain state at
+    // pool entry, and simulation overrides the sender's code unconditionally, so
+    // this is the builder's only chance to catch a tuple that went stale while
+    // the op sat in the pool.
+    async fn get_eip7702_sender_states(
```
dancoombs, line 1278: Is there away to share this code with what we do in precheck? Don't force it.

dancoombs: Consolidated the piece that genuinely overlaps: precheck's local `is_7702_bytecode` is now replaced by the shared `authorization_utils::eip7702_delegate_from_code`, so precheck and the builder agree on what a delegation designator is (the shared parser is slightly stricter — exactly 23 bytes — which matches what EIP-7702 actually writes on chain).

I left the fetch itself separate: precheck's `get_eip7702_authority_data` is per-op and optionally fetches the pending tx count but not code, while the builder needs a batch per-sender fetch of latest count + code, and the two sides make different decisions with the data (violation at entry vs. keep-or-reject at assembly). A shared fetch helper would need to be parameterized over all of that, which felt forced.

pavelm (approved):

PR #1311, opened 2026-07-16
fix(types): lower user op time range buffer to 20s

jakehobbs:
## Proposed Changes

  - Lower `TIME_RANGE_BUFFER` from 60s to 20s — the minimum remaining validity window a user op must have to be included.
  - Effect: user ops with short-lived expirations (e.g. a 1-minute `validUntil`) are no longer rejected at `eth_sendUserOperation` or skipped during bundle assembly, as long as at least 20s of validity remain.
  - This is enforced against the intersected account + paymaster `validUntil`, so it applies to all user ops, not just bundler-sponsored ones.

### DIFF
--- a/crates/types/src/user_operation/mod.rs
+++ b/crates/types/src/user_operation/mod.rs
@@ -29,7 +29,7 @@ pub mod v0_7;
 use crate::{Entity, aggregator::AggregatorCosts, authorization::Eip7702Auth, chain::ChainSpec};
 
 /// A user op must be valid for at least this long into the future to be included.
-pub const TIME_RANGE_BUFFER: Duration = Duration::from_secs(60);
+pub const TIME_RANGE_BUFFER: Duration = Duration::from_secs(20);
 
 /// Overhead for bytes required for each bundle
 /// 4 bytes for function signature


crates/types/src/user_operation/mod.rs
```diff
@@ -29,7 +29,7 @@ pub mod v0_7;
 use crate::{Entity, aggregator::AggregatorCosts, authorization::Eip7702Auth, chain::ChainSpec};
 
 /// A user op must be valid for at least this long into the future to be included.
-pub const TIME_RANGE_BUFFER: Duration = Duration::from_secs(60);
+pub const TIME_RANGE_BUFFER: Duration = Duration::from_secs(20);
```
Richard-Dang, line 32: Should we make this configurable via chain config so it doesn't require a redeploy each time?

Richard-Dang (approved):

pavelm (approved):

crates/types/src/user_operation/mod.rs
pavelm: I was going to suggest the same but that will probably be a bigger refactor to propagate the config value to all call sites.

jakehobbs: @dancoombs said no

PR #1313, opened 2026-07-17
feat(pool): track poison user operation suspects in the mempool

dancoombs:
Part of the poison user operation design (PR 2 of 5, see `docs/designs/poison-user-operations.md`). Stacked on #1312.

## Proposed Changes

- Add pool-owned suspect state: each operation carries a single failure counter measured against two thresholds — suspect at `--pool.rpc_failures_before_suspect` (default 3), removed `--pool.max_suspect_rpc_failures` (default 3) failures later. Suspect status is derived from the counter; any successful submission resets it.
- Suspects are excluded from `best_operations` and retried alone, spaced by jittered exponential backoff (`--pool.suspect_rpc_backoff_initial_secs` 1s doubling to `--pool.suspect_rpc_backoff_max_secs` 600s).
- Add `Pool::report_bundle_outcome(entry_point, ops, outcome, provider_event_active)` applying the failure-flow table: reset on success, count non-terminal failures, remove terminal singletons, mark multi-op terminal/mined-revert bundles suspect. New `OpRemovalReason::{TerminalRpcError, RepeatedNonTerminalRpcFailure}`.
- During a provider event, suspect analysis continues but removal progress pauses: non-suspects still accrue toward suspicion (a poison herd cannot shield itself by tripping the signal with its own failures), while suspects' counters freeze and only their isolation backoff refreshes — provider issues can never remove an operation.
- Gate the entire flow behind a master switch, `--pool.suspect_tracking_enabled` (default `false`): when disabled, bundle outcomes change no UO state.
- `get_ops_summaries` gains a `max_suspects` cap and appends due suspects (backoff elapsed) beyond `max_ops`, flagged via a new `suspect` field on `PoolOperationSummary` — no separate query method for the assigner.
- Wire the full surface: local pool server, gRPC proto/client/server, mocks, CLI args, `docs/cli.md`, and align the design doc with the single-counter model.

**Inert by design**: the master switch defaults to off, nothing calls `report_bundle_outcome` in production, and the assigner passes `max_suspects: 0` — no operation can become suspect until PR 3 wires the bundle sender and the flag is deliberately enabled.

### DIFF
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -174,6 +174,49 @@ pub struct PoolArgs {
         env = "POOL_MAX_TIME_IN_POOL_SECS"
     )]
     pub max_time_in_pool_secs: Option<u64>,
+
+    /// Number of non-terminal RPC submission failures before a user operation
+    /// becomes a suspect and is only submitted alone.
+    #[arg(
+        long = "pool.rpc_failures_before_suspect",
+        name = "pool.rpc_failures_before_suspect",
+        env = "POOL_RPC_FAILURES_BEFORE_SUSPECT",
+        default_value = "3"
+    )]
+    pub rpc_failures_before_suspect: u32,
+
+    /// Number of non-terminal RPC submission failures a suspect is allowed
+    /// before it is removed from the pool. 0 disables removal.
+    ///
+    /// The suspect backoff schedule spaces these failures; a provider incident
+    /// outlasting the cumulative backoff can remove healthy operations. Raise
+    /// this threshold to tolerate longer incidents.
+    #[arg(
+        long = "pool.max_suspect_rpc_failures",
+        name = "pool.max_suspect_rpc_failures",
+        env = "POOL_MAX_SUSPECT_RPC_FAILURES",
+        default_value = "3"
+    )]
+    pub max_suspect_rpc_failures: u32,
+
+    /// Initial delay in seconds between suspect isolation attempts, doubling
+    /// with each failure.
+    #[arg(
+        long = "pool.suspect_rpc_backoff_initial_secs",
+        name = "pool.suspect_rpc_backoff_initial_secs",
+        env = "POOL_SUSPECT_RPC_BACKOFF_INITIAL_SECS",
+        default_value = "1"
+    )]
+    pub suspect_rpc_backoff_initial_secs: u64,
+
+    /// Maximum delay in seconds between suspect isolation attempts.
+    #[arg(
+        long = "pool.suspect_rpc_backoff_max_secs",
+        name = "pool.suspect_rpc_backoff_max_secs",
+        env = "POOL_SUSPECT_RPC_BACKOFF_MAX_SECS",
+        default_value = "600"
+    )]
+    pub suspect_rpc_backoff_max_secs: u64,
 }
 
 impl PoolArgs {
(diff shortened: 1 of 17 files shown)


crates/pool/src/mempool/pool.rs
```diff
@@ -224,7 +236,98 @@ where
...
+                for hash in ops {
+                    let Some(op) = self.by_hash.get(hash) else {
+                        continue;
+                    };
+                    // While a provider event is active only suspects accrue
+                    // failures: suspect removal is protected by backoff-spaced
+                    // evidence rather than by outage detection.
+                    if !op.is_suspect(suspect_threshold) && provider_event_active {
+                        continue;
+                    }
```
dancoombs: Is this the logic we want? Should we accrue at all during provider events?

crates/pool/src/mempool/pool.rs
```diff
@@ -224,7 +238,113 @@ where
     }
 
     pub(crate) fn best_operations(&self) -> impl Iterator<Item = Arc<PoolOperation>> + '_ {
-        self.best.iter().map(|o| o.po.clone())
+        self.best
+            .iter()
+            .filter(|o| !o.is_suspect(self.config.rpc_failures_before_suspect))
```
jakehobbs: With `--pool.suspect_tracking_enabled=false` and `--pool.rpc_failures_before_suspect=0`, every fresh op has `failures == 0`, so `is_suspect(0)` is true and this filter removes all ops from `best_operations`. That violates the master switch guarantee and can stop builders from receiving any work due only to a threshold config; gate the filter on `suspect_tracking_enabled` or reject zero.

posted by codex ✨

dancoombs: Good catch — fixed by gating the `is_suspect` filter in `best_operations()` and `due_suspect_operations()` on `suspect_tracking_enabled` directly, rather than relying on the threshold value alone. That way the master switch guarantee holds regardless of `rpc_failures_before_suspect`, including a threshold of 0. Added `disabled_suspect_tracking_ignores_zero_threshold` to cover this case.

bin/rundler/src/cli/pool.rs
```diff
@@ -174,6 +174,59 @@ pub struct PoolArgs {
...
+    /// before it is removed from the pool. 0 disables removal.
+    ///
+    /// The suspect backoff schedule spaces these failures; a provider incident
+    /// outlasting the cumulative backoff can remove healthy operations. Raise
+    /// this threshold to tolerate longer incidents.
+    #[arg(
+        long = "pool.max_suspect_rpc_failures",
+        name = "pool.max_suspect_rpc_failures",
+        env = "POOL_MAX_SUSPECT_RPC_FAILURES",
+        default_value = "3"
```
dancoombs: This should be higher than 3, maybe 8? The goal is to ensure that we've had a chance to kick in any provider event tracking before we remove any suspects from the mempool.

dancoombs: Agreed, bumped to 8. With the exponential backoff defaults (1s initial, doubling, 600s cap), the cumulative delay across the allowed suspect failures before removal was only ~3s at max_suspect_rpc_failures=3 (1+2s of backoff plus attempt latency) — nowhere near enough time for the provider-event window (needs at least 10 observations) to fill and pause removal during a real outage. At 8, cumulative backoff is ~255s (1+2+4+8+16+32+64+128s before jitter), giving the event signal real headroom. Updated the CLI default, docs/cli.md, and the design doc (the "fast eviction" framing no longer applied once the default moved off 3, so I reworded that paragraph too).

PR #1314, opened 2026-07-17
feat(builder): wire the poison user operation failure flow, isolation scheduling, and provider-event signal

dancoombs:
Part of the poison user operation design (PRs 3, 4, and 5 of 5 — completing the series, see `docs/designs/poison-user-operations.md`). Stacked on #1313. PRs 3 and 4 are combined because PR 3 alone would leave suspects excluded from bundles with nothing draining them; PR 5 rides along because it reduces to a small module plus two lines at the choke point PR 3 created.

### Failure flow in the bundle sender (PR 3)

- Report every bundle submission outcome to the pool via `report_bundle_outcome`: `Success` on acceptance (resets failure counters, clearing suspect status), and on error, classify via `TxSenderError::classify()` — terminal errors report `TerminalFailure` (pool removes a sole op, marks larger bundles suspect), non-terminal errors report `NonTerminalFailure` (pool advances failure counters). Neutral operational errors (underpriced, nonce-too-low, etc.) keep their dedicated handling and are not reported.
- Split `process_revert` into a `RevertOutcome` of removals and suspect markings: an unattributed mined revert (`None | Revert | PostOpRevert`, or no revert data) removes a sole op but marks every op of a larger bundle suspect instead of removing all of them. Attributed reverts (`FailedOp`, signature validation, submission proxy) still remove.

### Suspect scheduling in the assigner (PR 4)

- Fetch due suspects alongside normal summaries (`max_suspects`) and hand them out as tagged single-operation isolation assignments, never mixed into normal bundles.
- While normal work is eligible, cap concurrent isolation at `max(1, num_signers / 2)` so suspects cannot starve normal bundles; once normal work is exhausted, any idle builder may isolate. The limit is recomputed per assignment and in-flight isolation work is never canceled.
- Suspects respect the same sender locks and fee filtering as normal work; replacement attempts never start new isolation work (an accepted in-flight bundle has already cleared its ops' suspect status).
- With a single builder signer, the scheduler cannot guarantee progress for both work classes when both remain continuously eligible — documented on the CLI args.

### Provider-event signal (PR 5)

- `ProviderEventSignal`: a rolling window of the last 20 non-neutral final submission outcomes; enter an event at ≥ 10 observations with ≥ 30% failures, exit below 10% (hysteresis prevents flapping). One signal per submission route, shared by the route's builders; exposed as a `builder_sender_provider_event_active` gauge.
- Outcomes are recorded at the bundle-sender choke point where final results are already classified — downstream of the fallback sender's internal retries and failover, so only final outcomes enter the window. Successes and non-terminal provider failures are recorded; neutral and terminal errors are excluded, as are cancellation transactions.
- The live signal state is passed as `provider_event_active` when reporting outcomes; a failure is recorded before it is reported so the report is gated by the freshest state. Pool-side (from PR 2), an active event pauses suspect removal progress only.

**Still dark**: all suspect state changes remain behind PR 2's `--pool.suspect_tracking_enabled` master switch (default `false`). With it off, reports change no UO state and the pool returns no suspects to isolate.

### DIFF
--- a/crates/builder/src/assigner.rs
+++ b/crates/builder/src/assigner.rs
@@ -52,6 +52,9 @@ pub(crate) struct WorkAssignment {
     pub filter_id: Option<String>,
     /// The assigned operations
     pub operations: Vec<PoolOperation>,
+    /// Whether this is an isolation bundle for a single suspect operation.
+    /// See `docs/designs/poison-user-operations.md`.
+    pub is_isolation: bool,
 }
 
 /// Detailed result of attempting to assign work.
@@ -96,6 +99,9 @@ struct State {
     entrypoint_last_selected: HashMap<ProposerKey, u64>,
     /// Tracks which entrypoint a builder is pinned to for replacement attempts
     builder_to_pinned_proposer: HashMap<Address, ProposerKey>,
+    /// Builders currently holding an isolation assignment for a suspect
+    /// operation. Bounds isolation capacity while normal work is eligible.
+    isolating_builders: HashSet<Address>,
 }
 
 /// A candidate entrypoint for assignment, used during priority-based selection.
@@ -204,7 +210,7 @@ impl Assigner {
                 self.pool.get_ops_summaries(
                     ep.address,
                     self.max_pool_ops_per_request,
-                    0, // suspect isolation scheduling is not implemented yet
+                    self.num_signers as u64,
                     ep.filter_id.clone(),
                 )
             })
@@ -212,6 +218,7 @@ impl Assigner {
         let query_results = futures_util::future::join_all(query_futures).await;
 
         let mut candidates = Vec::new();
+        let mut suspect_candidates = Vec::new();
         let mut fee_filtered_candidate_exists = false;
         for (entrypoint_idx, (ep, result)) in self
             .entrypoints
@@ -219,7 +226,10 @@ impl Assigner {
             .zip(query_results.into_iter())
             .enumerate()
         {
-            let ops = result?;
+            let (suspects, ops): (Vec<_>, Vec<_>) = result?.into_iter().partition(|op| op.suspect);
+            if !suspects.is_empty() {
+                suspect_candidates.push((ep.clone(), suspects));
+            }
             let (assignable_count, eligible_count) =
                 self.count_ops(&ops, builder_address, max_sim_block_number, &required_fees);
             let ep_metrics = ep_metrics_for(ep);
@@ -233,6 +243,32 @@ impl Assigner {
                 });
             } else if assignable_count > 0 {
                 fee_filtered_candidate_exists = true;
+            }
+        }
+
+        // Isolation is filled before normal work: while normal work is
+        // eligible, at most `max(1, num_signers / 2)` builders isolate so
+        // suspects cannot starve normal bundles; once normal work is
+        // exhausted, any idle builder may drain suspects. Recomputed on every
+        // assignment.
+        if !suspect_candidates.is_empty() {
+            let isolation_limit = if candidates.is_empty() {
+                self.num_signers
+            } else {
+                (self.num_signers / 2).max(1)
+            };
+            let isolating_count = self.state.lock().unwrap().isolating_builders.len();
+            if isolating_count < isolation_limit
+                && let Some(assignment) = self
+                    .assign_isolation_work(
+                        builder_address,
+                        suspect_candidates,
+                        max_sim_block_number,
+                        &required_fees,
+                    )
+                    .await?
+            {
+                return Ok(AssignmentResult::Assigned(assignment));
             }
         }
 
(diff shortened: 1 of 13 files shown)


crates/builder/src/bundle_sender.rs
```diff
@@ -1073,9 +1149,16 @@ where
             "Unknown entry point config: {ep_address:?}, filter_id: {filter_id:?}"
         ))?;
 
-        let to_remove = proposer.process_revert(tx_hash).await?;
+        let outcome = proposer.process_revert(tx_hash).await?;
 
-        self.remove_ops_from_pool_by_hash(ep_address, to_remove)
+        self.report_bundle_outcome(
```
jakehobbs, line 1154: When `--pool.suspect_tracking_enabled` is false (the default), this `MarkSuspect` report is ignored by the pool. For multi-op unattributed reverts, `to_mark_suspect` is the only output, so the remove-all behavior from main is lost and the same reverting ops can remain eligible instead of being removed. (posted by codex ✨)

crates/builder/src/assigner.rs
```diff
@@ -236,6 +246,32 @@ impl Assigner {
...
+        // suspects cannot starve normal bundles; once normal work is
+        // exhausted, any idle builder may drain suspects. Recomputed on every
+        // assignment.
+        if !suspect_candidates.is_empty() {
+            let isolation_limit = if candidates.is_empty() {
+                self.num_signers
+            } else {
+                (self.num_signers / 2).max(1)
+            };
+            let isolating_count = self.state.lock().unwrap().isolating_builders.len();
```
jakehobbs: When multiple builder tasks call `assign_work` concurrently, they can all read this count before any of them inserts into `isolating_builders` inside `assign_isolation_work`, so each can take a suspect assignment and exceed the `max(1, num_signers / 2)` cap while normal work is eligible. Reserve the slot under the same lock as the capacity check. (posted by codex ✨)

dancoombs:
Open question: should we use the provider-event signal to trigger the fallback behavior, and remove the existing sequential failures counter?

crates/builder/src/sender/health.rs
```diff
@@ -0,0 +1,209 @@
...
+use metrics_derive::Metrics;
+
+/// Number of most recent non-neutral final submission outcomes considered.
+const WINDOW_SIZE: usize = 20;
+/// Minimum observations in the window before an event can be entered.
+const MIN_OBSERVATIONS: usize = 10;
+/// Failure ratio at or above which an event is entered.
+const ENTER_FAILURE_RATIO: f64 = 0.30;
+/// Failure ratio below which an active event is exited.
+const EXIT_FAILURE_RATIO: f64 = 0.10;
```
dancoombs, line 32: Is it worth making this configurable?

crates/builder/src/assigner.rs
```diff
@@ -204,22 +210,26 @@ impl Assigner {
...
+        let mut suspect_candidates = Vec::new();
         let mut fee_filtered_candidate_exists = false;
         for (entrypoint_idx, (ep, result)) in self
             .entrypoints
             .iter()
             .zip(query_results.into_iter())
             .enumerate()
         {
-            let ops = result?;
+            let (suspects, ops): (Vec<_>, Vec<_>) = result?.into_iter().partition(|op| op.suspect);
```
dancoombs, line 229: We need to make it very clear in the pool/builder API that suspects come after non-suspects always. If this isn't documented somewhere very clear we should change that.

crates/builder/src/bundle_sender.rs
dancoombs: Confirmed — good catch. When `suspect_tracking_enabled` is false, the master switch was a blanket no-op for every `BundleOutcome`, including `MarkSuspect`. But the multi-op unattributed-revert path used to unconditionally remove every op before this PR — so with the switch off, ops from that path were silently left live with no path to eviction. That's the exact livelock class in `INVESTIGATION-empty-revert-bundle-livelock.md`.

Fixed in `apply_bundle_outcome`: when disabled, `MarkSuspect` now falls back to full removal (`OpRemovalReason::Requested`), matching the pre-feature behavior; every other outcome stays a no-op since those are genuinely new functionality with nothing to fall back to. Added regression tests at both the `PoolInner` and `UoPool` levels — the existing `PoolInner` test had actually encoded the buggy no-op as expected behavior, so that assertion was corrected too. Also updated the design doc to describe the fallback explicitly.

crates/builder/src/assigner.rs
dancoombs: Confirmed — the capacity check (`isolating_count = ...len()`) and the eventual insert (inside `assign_isolation_work`, past an `.await`) were under separate lock acquisitions, so concurrent builders could all read a stale count before any of them registered.

Fixed by reserving the slot atomically: check-and-insert into `isolating_builders` under a single lock acquisition *before* calling `assign_isolation_work`, releasing the reservation if no suspect ends up claimable (or the call errors). The insert that used to happen inside `assign_isolation_work` on success is now redundant and removed.

Added a concurrency regression test using real OS-thread parallelism (`tokio::spawn` + an artificial delay in the mock's `get_ops_by_hashes`, since mockall's async returns are immediately-ready and won't yield on a single-threaded runtime). Verified it manually against the pre-fix code: failed 15/15 runs; passes reliably against the fix.

crates/builder/src/sender/health.rs
dancoombs: Landed on leaving these hardcoded for now. The design doc frames the window/ratios as "initial parameters," and unlike the sibling suspect thresholds (`rpc_failures_before_suspect`, `suspect_rpc_backoff_*`), PR5's plan doesn't call for CLI config here. Since the whole mechanism ships dark behind `suspect_tracking_enabled`, it seems better to tune based on real operational data once it's live rather than guessing at config surface now. Easy to add flags later — mirroring the existing pool-side pattern — if experience shows we need to.

crates/builder/src/assigner.rs
dancoombs: Confirmed the invariant holds (`get_ops_summaries` does `summaries.extend(due_suspect_operations...)`, non-suspects first), but it wasn't documented anywhere a future implementor or caller would actually see it — the trait doc comment only mentioned the counts, not the ordering.

Expanded the doc comment on `Pool::get_ops_summaries` (crates/types/src/pool/traits.rs) to state the ordering explicitly, and to spell out that callers must partition on the `suspect` flag rather than assume position — which is what this code already does via `.partition(|op| op.suspect)`, so no behavior changed, just made the contract explicit.

dancoombs:
Landed on no. They're solving different problems at different timescales:

- `FallbackTransactionSender.consecutive_failures` reacts to a handful of consecutive errors from *one builder's* primary sender, so it can reroute traffic quickly and keep that builder submitting.
- The provider-event signal needs a much larger, hysteretic sample (≥10 of the last 20 observations, shared across *all* of a route's builders) before it gates something as consequential as pausing pool-wide suspect removal — fast-flapping that gate would be worse than the outage it's meant to protect against.

It's also circular to derive one from the other: the signal is fed downstream of `FallbackTransactionSender`'s retries/failover (per PR 5's design), so it only ever observes the *outcome* of a failover decision that's already been made — it can't drive that same decision.

Added a note to the design doc (`docs/designs/poison-user-operations.md`, end of the PR 5 section) making this explicit so it doesn't come up again as an implicit gap.

PR #1315, opened 2026-07-20
feat(pool,builder): add poison user operation telemetry (PR 6)

dancoombs:
## Summary

Implements PR 6 from `docs/designs/poison-user-operations.md` (also added in
this PR): the telemetry section discussed on top of PR 3+4. Stacked on
#1314.

- **docs**: adds the Telemetry section (Metrics, Logs, Dashboards, Alerts)
  and a PR 6 entry in the implementation plan.
- **pool** (`op_pool` scope): live `num_suspect_ops` gauge maintained
  incrementally at every suspect/clear/removal transition (not a per-call
  pool scan — removal is hooked at the single `remove_operation_internal`
  choke point so the gauge can't leak regardless of *why* a suspect leaves
  the pool); `num_marked_suspect_multi_op_failure` /
  `num_marked_suspect_threshold` / `num_suspects_cleared` /
  `num_suspects_removed` counters; a `suspect_tracking_enabled` gauge set
  once at construction. Two new `OpPoolEvent` variants, `MarkedSuspect` and
  `SuspectCleared`, logged for free via the existing generic event sink (no
  proto impact — `OpPoolEvent` never crosses the pool's gRPC boundary).
- **builder** (`builder_assigner` / `builder_sender` scopes): global
  `isolation_limit` / `isolating_builders` gauges in the assigner, recomputed
  every assignment cycle (this data is global, not per-entrypoint, unlike
  the existing `ep_isolation_assignments` counter); a per-sender
  `report_bundle_outcome` classification breakdown
  (`builder_bundle_outcome_{success,non_terminal_failure,terminal_failure,mark_suspect}`)
  plus a `builder_bundle_outcome_report_failed` counter, both added at the
  bundle sender's single existing `report_bundle_outcome` helper; a
  `provider_events_total` counter alongside the existing
  `provider_event_active` gauge.

No config, CLI, or behavior changes — purely additive observability on top
of the already-landed suspect-tracking/isolation/provider-event mechanisms.

## Test plan

- [x] `cargo check -p rundler-pool -p rundler-builder`
- [x] `cargo clippy -p rundler-pool -p rundler-builder --all-targets -- -D warnings` (clean)
- [x] `cargo test -p rundler-pool -p rundler-builder --lib` (264 passed, all pre-existing suspect/assigner/health tests green — telemetry is additive and behavior-preserving)
- [x] `make fmt` (no changes)

### DIFF
--- a/crates/builder/src/assigner.rs
+++ b/crates/builder/src/assigner.rs
@@ -251,21 +251,29 @@ impl Assigner {
         // suspects cannot starve normal bundles; once normal work is
         // exhausted, any idle builder may drain suspects. Recomputed on every
         // assignment.
+        let isolation_limit = if candidates.is_empty() {
+            self.num_signers
+        } else {
+            (self.num_signers / 2).max(1)
+        };
+        self.metrics.isolation_limit.set(isolation_limit as f64);
+
         if !suspect_candidates.is_empty() {
-            let isolation_limit = if candidates.is_empty() {
-                self.num_signers
-            } else {
-                (self.num_signers / 2).max(1)
-            };
             // Reserve the isolation slot under the same lock as the capacity
             // check: concurrent builders calling assign_work must not both
             // read a count below the limit and then both insert, exceeding
             // isolation_limit. Released below if no suspect ends up
-            // claimable, so a failed attempt doesn't waste the slot.
+            // claimable, so a failed attempt doesn't waste the slot. The
+            // gauge is refreshed inside every critical section that touches
+            // `isolating_builders` so it can't go stale between mutations.
             let reserved = {
                 let mut state = self.state.lock().unwrap();
-                state.isolating_builders.len() < isolation_limit
-                    && state.isolating_builders.insert(builder_address)
+                let reserved = state.isolating_builders.len() < isolation_limit
+                    && state.isolating_builders.insert(builder_address);
+                self.metrics
+                    .isolating_builders
+                    .set(state.isolating_builders.len() as f64);
+                reserved
             };
             if reserved {
                 match self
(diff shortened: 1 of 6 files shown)


jakehobbs:
Codex review note (posted by Codex/AI):

I found two issues in this PR:

1. `isolating_builders` is sampled before `assign_isolation_work` inserts the builder, and release paths remove builders without updating the gauge. After a successful isolation assignment, or after an isolation slot is freed with no immediate later `assign_work` call, the metric no longer reflects the current isolation set, which breaks saturation/idle dashboards.

`crates/builder/src/assigner.rs` around line 261

2. The telemetry doc added in this branch defines `ep_isolation_limit` and `ep_isolating_builders` as per-entrypoint assigner metrics, but the implementation puts these fields on `GlobalMetrics`, so they emit unlabeled global series named `isolation_limit` / `isolating_builders`. In multi-entrypoint configs, route-level dashboards cannot query the documented `ep_*` metrics and the global value is just the most recent assignment cycle.

`crates/builder/src/assigner.rs` around line 884

Verification: `git diff --check` passed. A narrow `cargo test -p rundler-pool apply_bundle_outcome --no-default-features` did not reach the crate tests because `metrics-0.24.1` failed to compile on the installed toolchain (`E0521`); `rustup` was also unavailable for the documented nightly fmt command.

dancoombs:
Rebased onto the updated `danc/poison-ops-pr3-4-failure-flow-scheduling` (which had since gained the atomic isolation-slot-reservation fix) and addressed both issues:

1. **Gauge staleness**: merged the atomic reservation logic with gauge updates — `isolating_builders` is now refreshed inside every critical section that mutates it: the reserve in `assign_work`, both release paths (`Ok(None)` / `Err`) after a failed isolation attempt, and the two out-of-band removal sites in `confirm_senders_drop_unused` and `release_all`. It can no longer drift from the actual set size.
2. **Doc/impl mismatch**: these are genuinely global concepts (isolation capacity and the isolating-builder set are shared across all configured entrypoints, not partitioned per entrypoint), so I updated the design doc's metric names from `ep_isolation_limit`/`ep_isolating_builders` to `isolation_limit`/`isolating_builders` (and noted they're global) to match the implementation, rather than restructuring the implementation to be per-entrypoint.

`cargo check`/`clippy -D warnings`/`test --lib` (268 passed) and `make fmt` all clean.

PR #1316, opened 2026-07-20
feat(pool,builder): poison user operation detection, isolation, and telemetry

dancoombs:
N/A

## Proposed Changes

Combined view of the poison user operations stack (design + PRs 1-6), squashed
across #1306, #1312, #1313, #1314, #1315 onto `main`:

- **docs**: add `docs/designs/poison-user-operations.md` describing the
  liveness/fairness/provider-resilience goals, the failure-flow table, and a
  telemetry section (metrics, logs, dashboards, alerts).
- **builder**: classify submission errors as terminal, non-terminal, or
  neutral; treat intrinsic-gas-too-low and internal RPC errors as terminal
  (the latter behind a chain-spec flag); wrap `TxSenderError` in
  `TransactionTrackerError`.
- **pool**: track poison user operation suspects in the mempool, gated behind
  a `--pool.suspect_tracking_enabled` master switch (default `false`);
  pause suspect *removal* (not analysis) during provider events; fall back to
  removal for unattributed reverts when suspect tracking is disabled; raise
  the default `max_suspect_rpc_failures` to 8.
- **builder**: wire the failure flow end-to-end — report bundle submission
  outcomes to the pool, schedule suspect isolation bundles in the assigner,
  detect provider-health events via a rolling outcome window, record
  submission outcomes in the provider-event signal, and make isolation slot
  reservation atomic in the assigner.
- **telemetry**: live suspect-count gauges, marked/cleared/removed counters,
  isolation gauges, per-sender bundle-outcome counters, and provider-event
  counters — purely additive observability, no config/CLI/behavior changes.

No behavior changes beyond what's already described in the stacked PRs; this
PR is the single merge target for that work into `main`.

## Test plan

- [x] `cargo check` (workspace)
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --lib` (pool + builder suites green)
- [x] `make fmt`

### DIFF
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -186,7 +186,10 @@ pub struct PoolArgs {
     pub suspect_tracking_enabled: bool,
 
     /// Number of non-terminal RPC submission failures before a user operation
-    /// becomes a suspect and is only submitted alone.
+    /// becomes a suspect and is only submitted alone. Must be at least 1: at
+    /// 0, every fresh operation would satisfy `failures >= threshold`
+    /// immediately, and a successful submission resetting failures to 0
+    /// would make it look suspect again on its very next attempt.
     ///
     /// With a single builder signer, the scheduler cannot guarantee progress
     /// for both suspect and normal work when both remain continuously
@@ -195,7 +198,8 @@ pub struct PoolArgs {
         long = "pool.rpc_failures_before_suspect",
         name = "pool.rpc_failures_before_suspect",
         env = "POOL_RPC_FAILURES_BEFORE_SUSPECT",
-        default_value = "3"
+        default_value = "3",
+        value_parser = clap::value_parser!(u32).range(1..)
     )]
     pub rpc_failures_before_suspect: u32,
 
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -187,6 +187,10 @@ pub struct PoolArgs {
 
     /// Number of non-terminal RPC submission failures before a user operation
     /// becomes a suspect and is only submitted alone.
+    ///
+    /// With a single builder signer, the scheduler cannot guarantee progress
+    /// for both suspect and normal work when both remain continuously
+    /// eligible.
     #[arg(
         long = "pool.rpc_failures_before_suspect",
         name = "pool.rpc_failures_before_suspect",
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -174,6 +174,49 @@ pub struct PoolArgs {
         env = "POOL_MAX_TIME_IN_POOL_SECS"
     )]
     pub max_time_in_pool_secs: Option<u64>,
+
+    /// Number of non-terminal RPC submission failures before a user operation
+    /// becomes a suspect and is only submitted alone.
+    #[arg(
+        long = "pool.rpc_failures_before_suspect",
+        name = "pool.rpc_failures_before_suspect",
+        env = "POOL_RPC_FAILURES_BEFORE_SUSPECT",
+        default_value = "3"
+    )]
+    pub rpc_failures_before_suspect: u32,
+
+    /// Number of non-terminal RPC submission failures a suspect is allowed
+    /// before it is removed from the pool. 0 disables removal.
+    ///
+    /// The suspect backoff schedule spaces these failures; a provider incident
+    /// outlasting the cumulative backoff can remove healthy operations. Raise
+    /// this threshold to tolerate longer incidents.
+    #[arg(
+        long = "pool.max_suspect_rpc_failures",
+        name = "pool.max_suspect_rpc_failures",
+        env = "POOL_MAX_SUSPECT_RPC_FAILURES",
+        default_value = "3"
+    )]
+    pub max_suspect_rpc_failures: u32,
+
+    /// Initial delay in seconds between suspect isolation attempts, doubling
+    /// with each failure.
+    #[arg(
+        long = "pool.suspect_rpc_backoff_initial_secs",
+        name = "pool.suspect_rpc_backoff_initial_secs",
+        env = "POOL_SUSPECT_RPC_BACKOFF_INITIAL_SECS",
+        default_value = "1"
+    )]
+    pub suspect_rpc_backoff_initial_secs: u64,
+
+    /// Maximum delay in seconds between suspect isolation attempts.
+    #[arg(
+        long = "pool.suspect_rpc_backoff_max_secs",
+        name = "pool.suspect_rpc_backoff_max_secs",
+        env = "POOL_SUSPECT_RPC_BACKOFF_MAX_SECS",
+        default_value = "600"
+    )]
+    pub suspect_rpc_backoff_max_secs: u64,
 }
 
 impl PoolArgs {
@@ -229,6 +272,10 @@ impl PoolArgs {
                 .verification_gas_limit_efficiency_reject_threshold,
             max_time_in_pool: self.max_time_in_pool_secs.map(Duration::from_secs),
             max_expected_storage_slots: common.max_expected_storage_slots.unwrap_or(usize::MAX),
+            rpc_failures_before_suspect: self.rpc_failures_before_suspect,
+            max_suspect_rpc_failures: self.max_suspect_rpc_failures,
+            suspect_rpc_backoff_initial: Duration::from_secs(self.suspect_rpc_backoff_initial_secs),
+            suspect_rpc_backoff_max: Duration::from_secs(self.suspect_rpc_backoff_max_secs),
         };
 
         let mut pool_configs = vec![];
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -175,6 +175,16 @@ pub struct PoolArgs {
     )]
     pub max_time_in_pool_secs: Option<u64>,
 
+    /// Master switch for poison user operation handling (suspect tracking,
+    /// isolation, and removal). Disabled by default.
+    #[arg(
+        long = "pool.suspect_tracking_enabled",
+        name = "pool.suspect_tracking_enabled",
+        env = "POOL_SUSPECT_TRACKING_ENABLED",
+        default_value = "false"
+    )]
+    pub suspect_tracking_enabled: bool,
+
     /// Number of non-terminal RPC submission failures before a user operation
     /// becomes a suspect and is only submitted alone.
     #[arg(
@@ -272,6 +282,7 @@ impl PoolArgs {
                 .verification_gas_limit_efficiency_reject_threshold,
             max_time_in_pool: self.max_time_in_pool_secs.map(Duration::from_secs),
             max_expected_storage_slots: common.max_expected_storage_slots.unwrap_or(usize::MAX),
+            suspect_tracking_enabled: self.suspect_tracking_enabled,
             rpc_failures_before_suspect: self.rpc_failures_before_suspect,
             max_suspect_rpc_failures: self.max_suspect_rpc_failures,
             suspect_rpc_backoff_initial: Duration::from_secs(self.suspect_rpc_backoff_initial_secs),
--- a/bin/rundler/src/cli/pool.rs
+++ b/bin/rundler/src/cli/pool.rs
@@ -205,7 +205,7 @@ pub struct PoolArgs {
         long = "pool.max_suspect_rpc_failures",
         name = "pool.max_suspect_rpc_failures",
         env = "POOL_MAX_SUSPECT_RPC_FAILURES",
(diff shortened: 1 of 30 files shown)


crates/builder/src/assigner.rs
```diff
@@ -382,6 +508,10 @@ impl Assigner {
             .get_ops_summaries(
                 entry_point,
                 self.max_pool_ops_per_request,
+                // Replacement attempts never start new isolation work: an
+                // in-flight accepted bundle has already cleared its ops'
```
jakehobbs, line 577: [P1] Preserve singleton isolation across replacements

After an isolation transaction is accepted, `Success` clears the suspect flag. If it remains pending long enough to need replacement, this pinned path requests ordinary operations and `assign_ops_internal` can build a multi-op replacement or omit the original suspect entirely. A reverting poison can therefore contaminate healthy operations again or escape quarantine. Please retain the isolation hash through replacement attempts and keep those replacements singleton until resolution.

From Codex (sol).

bin/rundler/src/cli/pool.rs
```diff
@@ -174,6 +174,63 @@ pub struct PoolArgs {
...
+    /// With a single builder signer, the scheduler cannot guarantee progress
+    /// for both suspect and normal work when both remain continuously
+    /// eligible.
+    #[arg(
+        long = "pool.rpc_failures_before_suspect",
+        name = "pool.rpc_failures_before_suspect",
+        env = "POOL_RPC_FAILURES_BEFORE_SUSPECT",
+        default_value = "3"
+    )]
+    pub rpc_failures_before_suspect: u32,
```
jakehobbs, line 204: [P2] Reject a zero suspicion threshold

With tracking enabled and `rpc_failures_before_suspect=0`, every fresh operation satisfies `failures >= threshold`. A successful submission resets failures to zero, so the operation remains suspect while emitting `SuspectCleared` and decrementing the gauge. Since the CLI currently accepts zero without documenting special semantics, please require this value to be at least one.

From Codex (sol).

crates/pool/src/mempool/pool.rs
```diff
@@ -765,6 +958,13 @@ where
         block_number: Option<u64>,
     ) -> Option<Arc<PoolOperation>> {
         let op = self.by_hash.remove(&hash)?;
+        // Recompute the current suspect population on every removal path
+        // (mined, expired, replaced, entity-removed, poison-ops removal,
+        // etc.) so the gauge never leaks regardless of why a suspect left
+        // the pool.
+        if op.is_suspect(self.config.rpc_failures_before_suspect) {
```
jakehobbs, line 971: [P2] Reset the suspect gauge during bulk clear

This decrement only runs through `remove_operation_internal`, while `PoolInner::clear()` directly clears the maps. Calling `debug_clearState` while suspects exist therefore leaves `num_suspect_ops` permanently stale, invalidating the new dashboard and alerts. Please reset the gauge in the bulk-clear path as well.

From Codex (sol).

crates/builder/src/assigner.rs
dancoombs: Fixed in d58ded10. Added `builder_to_isolation_hash` tracking to `Assigner::State`, set alongside `builder_to_pinned_proposer` when `assign_isolation_work` succeeds. `assign_work`'s pinned-builder path now checks for an active isolation hash before falling through to `assign_work_for_entrypoint`; if one exists it calls a new `assign_isolation_replacement`, which re-fetches that exact op by hash via `Pool::get_op_by_hash` and returns it alone (`is_isolation: true`), never mixing in other senders or re-deriving eligibility from pool state. Cleared at the same two points `builder_to_pinned_proposer`/`isolating_builders` already are (`confirm_senders_drop_unused` when a builder's senders empty out, and `release_all`). Added `test_isolation_replacement_stays_singleton_and_pinned`, which reproduces the exact scenario in the comment: isolate a suspect, confirm its send (the state a stuck tx would be in when a replacement is needed), then assert the next assignment for that builder is still the same op alone - not merged with the two normal ops also present in the pool.

bin/rundler/src/cli/pool.rs
dancoombs: Fixed in d58ded10. Added `value_parser = clap::value_parser!(u32).range(1..)` to `--pool.rpc_failures_before_suspect`, so 0 is now rejected at CLI parse time (verified with a live `rundler node` invocation: `error: invalid value '0' for '--pool.rpc_failures_before_suspect ...': 0 is not in 1..=4294967295`). Also expanded the field doc comment to spell out why, mirroring the existing `num_signers` precedent for `clap::value_parser!(...).range(...)` in this codebase.

crates/pool/src/mempool/pool.rs
dancoombs: Fixed in d58ded10. `PoolInner::clear()` now sets `self.metrics.num_suspect_ops.set(0.0)` directly rather than trying to decrement per-op - a bulk clear doesn't track how many of the cleared ops were suspects, so `set(0.0)` is the correct (and only fully accurate) op here, unlike the single-op removal path which knows exactly one op's suspect status and can decrement.

Richard-Dang (dismissed):

jakehobbs:
## Code review (posted by Claude)

Reviewed the full diff for the poison-user-operations feature (error classification, pool-owned suspect state, bundle-sender failure flow, assigner isolation scheduling, provider-event signal, telemetry, plus gRPC/proto and CLI plumbing).

Overall this is a high-quality, heavily-tested change. I traced the pool state machine (suspect promotion, freeze-during-event, backoff, removal thresholds), the `num_suspect_ops` gauge accounting on every removal path, the assigner's atomic isolation-slot reservation, and the remote conversions — they line up with the design doc and their tests. I found one issue worth attention.

### Intrinsic-gas-too-low reports a terminal failure to the pool, evicting/suspecting user ops

**File:** `crates/builder/src/sender/mod.rs` (`TxSenderError::classify`)

`classify()` returns `Terminal` for `IntrinsicGasTooLow`, so the bundle sender reports `BundleOutcome::TerminalFailure` for it. With `suspect_tracking_enabled` on, that removes the sole op of a single-op bundle (`TerminalRpcError`) or marks every op suspect in a multi-op bundle.

But intrinsic-gas-too-low means the *outer bundle transaction's* `gas_limit` is below the EVM intrinsic minimum — a bundler-side tx-construction error, not a per-UO rejection, and unattributable to any single op. So a builder gas-estimation bug would start evicting/suspecting innocent user ops, which contradicts the feature's goal of not removing innocent ops. The dedicated `IntrinsicGasTooLow` arm in `bundle_sender` already preserves the prior log-and-hard-error behavior, and (unlike `TerminalRpcError`) no test asserts a pool report for this variant — suggesting the pool interaction is an unintended consequence of the `Terminal` classification.

Suggested fix: classify `IntrinsicGasTooLow` as `Neutral` so it stays a hard bundler error without touching pool suspect state.

Severity is low in practice: it only bites when the (default-off) feature is enabled *and* an intrinsic-gas-too-low error actually occurs. But it's a discrete, easy fix.

No other correctness, concurrency, or plumbing bugs found in the suspect state machine, provider-event signal, assigner, or remote layer.

crates/builder/src/assigner.rs
```diff
@@ -204,21 +229,26 @@ impl Assigner {
                 self.pool.get_ops_summaries(
                     ep.address,
                     self.max_pool_ops_per_request,
+                    self.num_signers as u64,
```
jakehobbs: **Suspect query can permanently hide eligible work**

`max_suspects` is capped at `num_signers` before the assigner filters by fees, sponsorship budget, block height, and locks. If those first suspects are unassignable, later eligible suspects are never returned or attempted, so they receive no backoff and can remain starved indefinitely. Request a larger page, such as `max_pool_ops_per_request`, and cap actual isolation assignments locally.

Posted by Sol.

crates/builder/src/assigner.rs
```diff
@@ -360,12 +449,71 @@ impl Assigner {
...
+    /// slot in `isolating_builders` before calling this, and releases it if
+    /// `None` or an error is returned.
+    async fn assign_isolation_work(
+        &self,
+        builder_address: Address,
+        suspect_candidates: Vec<(EntrypointInfo, Vec<PoolOperationSummary>)>,
+        max_sim_block_number: u64,
+        required_fees: &GasFees,
+    ) -> anyhow::Result<Option<WorkAssignment>> {
+        for (ep, suspects) in suspect_candidates {
```
jakehobbs, line 507: **Isolation bypasses entrypoint starvation protection**

Isolation candidates are traversed in fixed entrypoint configuration order. If an earlier entrypoint continuously has enough due suspects to fill isolation capacity, later entrypoints never receive isolation assignments; the starvation-aware ordering used for normal work is bypassed. Suspect entrypoints need equivalent rotation or last-selected tracking.

Posted by Sol.

jakehobbs (commented):
Review findings posted by Sol.

crates/builder/src/assigner.rs
dancoombs: Fixed in ee3a3e3c. The suspect query now requests `max_pool_ops_per_request` (a full page) instead of `num_signers` for the `max_suspects` arg in `assign_work`'s entrypoint query. Since `due_suspect_operations` returns due suspects best-fee-first and the lock/fee/block-height/sponsorship filtering all happens later in `assign_ops_internal` (on the already-truncated set), the old `num_signers` cap could permanently hide a lower-fee-but-eligible suspect behind a handful of unassignable ones ahead of it in the page. The isolation-slot cap itself is unchanged — it's still enforced locally via `isolating_builders`/`isolation_limit`, so widening the query only affects visibility, not how many builders isolate concurrently. Added `test_suspect_query_requests_full_page_not_num_signers` asserting the query requests the full page (and it fails if reverted to `num_signers`).

crates/builder/src/assigner.rs
dancoombs: Fixed in ee3a3e3c. Added `entrypoint_last_isolated: HashMap<ProposerKey, u64>` and an `isolation_cycle` counter to `Assigner::State`, mirroring the existing `entrypoint_last_selected`/`global_cycle` mechanism that normal work already uses. Before dispatching to `assign_isolation_work`, `suspect_candidates` is now stably sorted by least-recently-isolated first, and `assign_isolation_work` bumps `isolation_cycle` and records the chosen entrypoint on each successful isolation assignment. So an entrypoint that always has enough due suspects to fill isolation capacity no longer starves the others — assignments rotate. Tracked on a separate counter from `global_cycle` so isolation rotation doesn't perturb normal-work starvation accounting. Added `test_isolation_rotates_across_entrypoints`, which sets up EP1 (2 due suspects + a normal op, so isolation stays capped at 1 slot) and EP2 (1 due suspect), and asserts four consecutive isolation assignments alternate `EP1, EP2, EP1, EP2` rather than draining EP1 forever; verified it fails without the sort.

dancoombs:
Re: the intrinsic-gas-too-low finding — after tracing the failure flow we decided to **keep it classified as `Terminal`** rather than move it to `Neutral`, and documented the reasoning in ee3a3e3c.

The key point is what happens on the alternative. Reclassifying to `Neutral` doesn't make the failure attributable — it just makes it *non-terminal*, so no op is removed or marked suspect. But the bundle gas limit is computed as a deterministic sum over the included ops' gas (`get_bundle_gas_limit_inner` in `bundle_proposer.rs` — per-op authorization gas, EIP-7623 calldata-floor gas, DA gas). So the identical op set produces the identical intrinsic-gas-too-low failure on the very next trigger, with zero pool state mutation in between — a livelock on that entrypoint. That's the same structural failure mode we hit in the empty-`handleOps`-revert incident (hard error → no pool mutation → assigner re-serves the same ops → repeat).

You're right that it's a bundle-level failure, not cleanly attributable to one op. But `Terminal` is exactly what makes it *self-healing*: in a multi-op bundle the ops are marked suspect and isolated one at a time; an innocent op retried alone has small calldata and passes, clearing its suspicion, while the genuine offender keeps failing whether bundled or isolated until it's removed. Breaking the livelock is the whole point of poison-op handling, so we treat "the bundler can't construct a valid tx from these ops" as terminal-for-these-ops even though it surfaces on the bundle as a whole. The dedicated `IntrinsicGasTooLow` arm in `bundle_sender` still runs its log-and-hard-error afterward; the classification only adds the pool report that lets the state machine converge.

dancoombs:
Note: the codecov comment above (85.19% patch coverage) is not blocking — the "missing" lines are concentrated in the gRPC/proto plumbing (`remote/server.rs`, `remote/client.rs`, `protos.rs`), the CLI wiring (`cli/builder.rs`), and the manual-testing RPC (`debug.rs` — the `debug_bundler_forceReportBundleOutcome` helper), none of which carry the state-machine logic. The suspect state machine, assigner isolation/rotation, and failure-classification paths that are the substance of this PR are covered by unit tests.

jakehobbs (approved):

PR #1318, opened 2026-07-22
fix(builder): recognize Robinhood's chain-policy rejection as terminal

dancoombs:
## Summary

Robinhood changed their `-32000` rejection message from `internal error` to
`Transaction rejected by chain policy`. `promote_terminal_internal_error`
only matched the exact old string, so this message silently stopped being
classified as terminal and fell back to ambiguous `UnrecognizedRpc` —
reintroducing the false-failover/mis-suspicion risk this branch's terminal
classification exists to prevent.

## Changes

- Add `TxSenderError::promote_terminal_chain_policy_rejection`, applied
  unconditionally in `RawTransactionSender::map_provider_error` (not gated
  by `ChainSpec::internal_rpc_error_is_terminal`) — the new message is
  unambiguous on its own, unlike the generic `internal error` string.
- Keep it as a separate function from `promote_terminal_internal_error` so
  the `internal error` match (and the flag) can be dropped independently
  once we're confident Robinhood no longer emits it.
- Tests covering: the new message promotes regardless of the flag,
  `internal error` still requires the flag, and near-miss messages/codes
  don't falsely promote.

Targets `danc/poison-ops` (#1316) — this error class only exists in that
branch's classification path, not on `main`.

## Test plan

- [x] `cargo test -p rundler-builder sender::`
- [x] `cargo clippy -p rundler-builder --all-features --tests -- -D warnings`
- [x] `cargo +nightly fmt --all --check`

### DIFF
--- a/crates/builder/src/sender/mod.rs
+++ b/crates/builder/src/sender/mod.rs
@@ -81,8 +81,10 @@ pub(crate) enum TxSenderError {
     /// The provider rejected the transaction with an error response known to be
     /// terminal for this chain: retrying the identical transaction cannot succeed.
     ///
-    /// Currently produced for `-32000: internal error` on chains with
-    /// `ChainSpec::internal_rpc_error_is_terminal`.
+    /// Produced for `-32000: internal error` on chains with
+    /// `ChainSpec::internal_rpc_error_is_terminal`, and unconditionally for
+    /// `-32000: Transaction rejected by chain policy` (Robinhood's unambiguous
+    /// wording for the same rejection, unlike the generic "internal error").
     #[error("terminal RPC error {code}: {message}")]
     TerminalRpcError {
         /// RPC error code.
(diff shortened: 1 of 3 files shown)


crates/builder/src/sender/mod.rs
```diff
@@ -159,6 +161,29 @@ impl TxSenderError {
...
+    /// Promotes Robinhood's `-32000: Transaction rejected by chain policy` from
+    /// ambiguous to terminal.
+    ///
+    /// Unlike the bare "internal error" message, this wording is unambiguous on its
+    /// own — no chain legitimately emits it for a transient condition — so it is
+    /// recognized unconditionally rather than behind
+    /// `ChainSpec::internal_rpc_error_is_terminal`. Kept separate from
+    /// `promote_terminal_internal_error` so the "internal error" match can be
+    /// dropped independently once Robinhood no longer emits it.
+    pub(crate) fn promote_terminal_chain_policy_rejection(self) -> Self {
```
dancoombs: This can be combined with the promote_terminal_internal_error function above. Lets call it `promote_terminal_error` and have it take a chain spec instance so that we're not leaking this process outside of this function.

dancoombs: Fixed in 7ea891c2. Merged `promote_terminal_internal_error` and `promote_terminal_chain_policy_rejection` into a single `promote_terminal_error(self, chain_spec: &ChainSpec) -> Self` on `TxSenderError`. It now internally decides both conditions — chain-policy wording unconditionally, "internal error" gated on `chain_spec.internal_rpc_error_is_terminal` — so `RawTransactionSender::map_provider_error` collapses to one call: `TxSenderError::from(error).promote_terminal_error(&self.chain_spec)`, with no more caller-side if/else branching on the flag.

`RawTransactionSender` and `RawSenderArgs` now hold the `ChainSpec` directly instead of a pre-extracted bool, matching how `ChainSpec` is already held by value elsewhere in this crate (`bundle_sender.rs`, `task.rs`). One side effect: embedding the full `ChainSpec` in `RawSenderArgs` made that variant much larger than its siblings in `TransactionSenderArgs`, tripping `clippy::large_enum_variant` — added `#[allow(clippy::large_enum_variant)]` there, mirroring the same allow already on the sibling runtime enum `TransactionSenderEnum` (which has the identical shape issue via `RawTransactionSender`).

Also rebased this branch onto the updated `danc/poison-ops` (now includes ee3a3e3c's assigner fixes).

PR #1319, opened 2026-07-24
feat(rpc): expand one-sided block ranges for user operation lookups

jakehobbs:
## Proposed Changes

  - When a max event block range is enforced (`--user_operation_event_block_distance` or `X-Rundler-Max-Block-Range`), a one-sided block range on the UO lookup methods (`eth_getUserOperationByHash`, `eth_getUserOperationReceipt`, `rundler_getUserOperationStatus`) is no longer rejected — the missing bound is expanded to a max-wide window anchored at the supplied bound: `{ fromBlock }` → `[fromBlock, fromBlock + max]` (capped at the chain head via a single `get_block_number` call), `{ toBlock }` → `[toBlock - max, toBlock]`.
  - A one-sided `fromBlock` past the chain head still returns `-32602` invalid params; ranges with both bounds present keep their existing over-width and reversed-range validation.
  - Block-option resolution now happens once per request (into `ResolvedEventBlockOptions`) and is deferred to just before the log-query fan-out. Paths that resolve receipts from a known bundle transaction (preconfirmed lookup) and the unsupported-pending-tag rejection no longer perform a head lookup, so a transient `eth_blockNumber` failure can't replace an `InvalidParams` or a valid preconfirmed response.
  - Updated the block-option docs and added tests covering the expansion, head-capping, and past-head cases.


crates/rpc/src/eth/events/common.rs
```diff
@@ -359,27 +359,54 @@ where
...
+                            )));
+                        }
+                        bo
+                    }
+                    // One-sided range with a max enforced: expand the missing bound to a
+                    // max_distance-wide window anchored at the supplied bound instead of
+                    // rejecting the request. Cap toBlock at the head so the resulting
+                    // window never runs past the chain tip.
+                    (Some(max_distance), Some(BlockNumberOrTag::Number(from)), None) => {
+                        let head = self.provider.get_block_number().await?;
```
dancoombs: I don't love doing this async call here, can we parallelize with `resolve_block_option` above? Maybe do it as part of `resolve_block_option`?

jakehobbs: Good call — removed the async call entirely rather than parallelizing it. The head lookup only existed to cap `toBlock` at the tip; instead I now set `toBlock = fromBlock + max` directly. The window is exactly `max` wide by construction, and `eth_getLogs` clamps an over-the-tip `toBlock` to the head rather than erroring.

I confirmed this against a live node: querying with `toBlock` ~41M blocks past the current head returned logs from the head block with no error. So the one-sided expansion branch is now pure/synchronous, no extra round trip.

(For transparency: this change was made by Claude.)

jakehobbs: Correction to my previous reply — you're right, and I've reverted the "rely on getLogs clamping" approach. `toBlock > latest` clamping is not in the JSON-RPC spec (geth/Alchemy clamp, but it's not guaranteed across clients), so synthesizing a `toBlock` past the head is unsafe.

I've now done what you originally suggested: folded the head lookup into `resolve_block_option`, which now handles tag resolution, window validation, and one-sided expansion together. For `{ fromBlock }` the synthesized `toBlock` is capped at the chain head (`min(fromBlock + max, head)`), so we never emit a `toBlock` past the tip. The `get_event_by_hash` call site is now a single `resolve_block_option(...)` call instead of an inline match plus a separate `get_block_number`.

(For transparency: this was done by Claude.)

crates/rpc/src/eth/events/mod.rs
```diff
@@ -43,8 +43,10 @@ impl EventBlockOptions {
     /// to multiple entry point routes without each route re-resolving the tags via RPC.
     pub(crate) async fn resolve<P: EvmProvider>(self, provider: &P) -> EventProviderResult<Self> {
         Ok(Self {
+            // Resolve tags only here (no max): the effective max distance is known per
+            // entry point route, which applies validation and one-sided expansion.
             block_option: match self.block_option {
-                Some(bo) => Some(resolve_block_option(provider, bo).await?),
+                Some(bo) => Some(resolve_block_option(provider, bo, None).await?),
```
dancoombs: Not sure I understand why we don't provide a max here?

dancoombs: I think it might be worth refactoring this a bit so that `max_block_range` is always set to `let event_block_distance = block_options.max_block_range.or(self.event_block_distance);` prior to resolving. And then we only resolve once.

We probably should have 2 types (1) unresolved which comes through the API and (2) resolved which passes to the events and always uses block numbers. Wdyt?

jakehobbs: Done — implemented in 351b0193.

Both types now exist: `EventBlockOptions` (unresolved, straight from the API) and `ResolvedEventBlockOptions` (concrete block numbers, handed to the event query layer). Resolution happens exactly once per request in the RPC handler before the fan-out: `EventBlockOptions::resolve(provider, event_block_distance)` computes the effective max as `self.max_block_range.or(event_block_distance)` and applies tag resolution, width validation, and one-sided expansion in that single pass. Every route then receives already-resolved block numbers.

To make the effective max available at the handler (it was previously per-provider), the static `event_block_distance` moved onto the `EntryPointRouter` — it's shared config, identical across routes — and the handler reads it via `router.event_block_distance()`. The per-route providers now only hold the fallback distance. Behavior is unchanged; 47 rpc tests pass, clippy clean.

(For transparency: this was done by Claude.)

jakehobbs: Addressed in 351b0193 — the max is now provided. This `resolve(provider, None)` call is gone: resolution moved up to the RPC handler, which passes the effective max (`max_block_range.or(router.event_block_distance())`) so tags, validation, and one-sided expansion all happen together in one pass. See the longer reply on your refactor comment below. (Done by Claude.)

jakehobbs:
@dancoombs one judgment call on the refactor (351b0193) I want your read on:

I put the shared static `event_block_distance` on `EntryPointRouter` (exposed via `router.event_block_distance()`) rather than threading it into `EthApi` and `RundlerApi` separately. Reasoning: it's a single global config value, identical across every route, and the router is already the fan-out owner the handler talks to — so the handler reads it there and passes it into `EventBlockOptions::resolve(...)`. The per-route providers now only hold the fallback distance.

The alternative is to store it directly on the two API structs (plumbed from settings in `task.rs`), leaving the router purely about routing.

Do you have a preference? Happy to move it to the API structs if you'd rather keep the router narrow.

dancoombs (approved):

PR #1322, opened 2026-07-29
fix(builder): treat rate limits as pacing, not poison evidence

dancoombs:
Follow-up to #1316. Prompted by the arb-sepolia bundle-send 429 storm on 07-29 (07:24-08:35 UTC, zero customer impact, 8 high-urgency pages).

`https://sepolia-rollup.arbitrum.io/rpc` began rate limiting our conditional submissions. A 429 collapses into `SenderUnavailable`, which classifies `NonTerminal`, so it entered the provider-health window and incremented the failure counter of every op in the bundle. At the default `--pool.rpc_failures_before_suspect` of 3, a few seconds of that made the whole pool suspect and every builder switched to one-op isolation bundles — against the endpoint that had just asked us for fewer requests. ~98% of submissions during the storm were isolation bundles.

Nothing paced the retries either: `--provider_rate_limit_retry_enabled` defaults to `false`, so alloy's `RetryBackoffLayer` is not installed on the submit provider, and the bundle sender had no 429 handling of its own. Each builder was free to re-submit on every trigger for as long as it had work — and because nothing landed, nothing left the pool, so every builder had work on every trigger.

Two independent problems, fixed separately here: a 429 is not evidence about a user operation, and a 429 needs backoff.

Worth noting what was *not* wrong: submission rate is capped at one attempt per builder per `bundle_max_send_interval_millis` regardless of isolation, and de-batching changes ops-per-submission rather than submissions-per-second. Isolation made the storm less efficient, not faster.

## Proposed Changes

  - Add `ProviderError::is_rate_limited` (`crates/provider/src/traits/error.rs`), recognising HTTP 429, `429`/`-32005`-style JSON-RPC error responses, and transports that leave the status in the error text. Every arm delegates to alloy so the provider-specific codes stay in one place: `HttpError::is_rate_limit_err`, `ErrorPayload::is_retry_err` (which carries `429`, Infura's `-32005`, QuickNode's credit limits and the rest), and the `Custom` check from `TransportErrorKind::is_retry_err`. HTTP 503 stays excluded — alloy groups it with rate limits as "retryable", but a temporarily unavailable endpoint is genuine provider-health evidence and must stay `NonTerminal`.
  - Add `TxSenderError::RateLimited`, classified `RpcOutcomeClass::Neutral`: no bundle outcome is reported to the pool and the provider-event signal never sees it. No failure counter, no suspect, no isolation, and no provider event freezing suspects' counters. Detection runs *before* the error-response paths, since a JSON-RPC-shaped rate limit would otherwise land in `UnrecognizedRpc` and reach `NonTerminal` by a different route.
  - Add `RateLimitBackoff` to the bundle sender: consecutive rate-limited attempts double the wait from `--builder.rate_limit_backoff_initial_millis` (100ms) to `--builder.rate_limit_backoff_max_millis` (2s), with ±20% jitter so builders sharing a route do not retry in the same instant. The backoff *replaces* the trigger wait rather than stacking on top of it — a rate-limited attempt retries as soon as the backoff elapses. Held on `BundleSenderImpl` rather than in the state machine, so it survives state resets and applies whichever state the last attempt left behind. The signer is held for the duration: `retry()` leaves `wait_for_trigger: false`, so the idle-release check does not fire. That is at most 2s, and releasing the key while the builder still holds assignments risks `begin_cycle` leasing a different address and orphaning them.
  - Clear the backoff only on outcomes the node actually judged (`endpoint_judged_transaction`). Clearing on any non-429 result would let an idle pool between rate-limited attempts silently hold the wait at its initial value, reproducing a pinned backoff in a new place.
  - Return `Ok(RateLimited)` instead of `Err`. The error path reset the transaction tracker, which a request refused before processing never warranted (unlike `SenderUnavailable`, acceptance is not ambiguous), and logged two `error!` lines per attempt per builder — ~64 error logs/s across a 32-signer fleet, which is what flapped `RundlerHighLogErrorCount` six times.
  - Leave `RateLimited` out of the fallback sender's failover arm. Failover stays for errors; pacing the primary is the backoff's job. This is a deliberate behaviour change from today, where a 429 counted toward failover as `SenderUnavailable`.
  - Correct `docs/designs/poison-user-operations.md`: its outcome table listed "exhausted rate limit" under `ProviderFailure`, and the claim that "existing provider rate limiting handles request pacing" was not true of any deployment, since that layer is off by default.

## Testing

`make test-unit` (514 passed), `make lint`, `cargo +nightly fmt --all --check`.

New coverage: rate-limit detection across all four shapes plus the 503 exclusion; `RateLimited` classifies `Neutral`; a rate-limited attempt reports no bundle outcome, records no provider-event observation, and leaves the backoff as the only thing the next attempt waits on; backoff escalation over the shipped 100ms→2s ladder, capping under jitter, clearing, and that the wait is consumed once rather than re-served every loop.

## Not in scope

  - The backoff is per-builder, not shared per submission route. `ProviderEventSignal` is already shared per route if we want one fleet-wide gate later.
  - `--pool.rpc_failures_before_suspect` is unchanged. Genuine `SenderUnavailable` outages still make the whole pool suspect within ~3s at the default of 3.
  - arb-sepolia's unkeyed submit endpoint is addressed separately in OMGWINNING/ws-tools#811.

### DIFF
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -11,7 +11,7 @@
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
 
-use std::{net::SocketAddr, str::FromStr, sync::Arc};
+use std::{net::SocketAddr, str::FromStr, sync::Arc, time::Duration};
 
 use alloy_primitives::Address;
 use anyhow::Context;
@@ -227,6 +227,28 @@ pub struct BuilderArgs {
         default_value = "0.50"
     )]
     assigner_starvation_ratio: f64,
+
+    /// Initial delay a builder waits before its next submission after the
+    /// submission endpoint rate limited it, doubling with each consecutive
+    /// rate-limited attempt and applied on top of the normal bundle trigger
+    /// interval. Any attempt that is not rate limited resets it.
+    #[arg(
+        long = "builder.rate_limit_backoff_initial_millis",
+        name = "builder.rate_limit_backoff_initial_millis",
+        env = "BUILDER_RATE_LIMIT_BACKOFF_INITIAL_MILLIS",
+        default_value = "1000"
+    )]
+    rate_limit_backoff_initial_millis: u64,
+
+    /// Maximum delay a builder waits between submissions while the submission
+    /// endpoint is rate limiting it.
+    #[arg(
+        long = "builder.rate_limit_backoff_max_millis",
+        name = "builder.rate_limit_backoff_max_millis",
+        env = "BUILDER_RATE_LIMIT_BACKOFF_MAX_MILLIS",
+        default_value = "30000"
+    )]
+    rate_limit_backoff_max_millis: u64,
 }
 
 impl BuilderArgs {
@@ -322,6 +344,10 @@ impl BuilderArgs {
             chain_spec,
             assigner_max_ops_per_request: self.assigner_max_ops_per_request,
             assigner_starvation_ratio: self.assigner_starvation_ratio,
+            rate_limit_backoff_initial: Duration::from_millis(
+                self.rate_limit_backoff_initial_millis,
+            ),
+            rate_limit_backoff_max: Duration::from_millis(self.rate_limit_backoff_max_millis),
         })
     }
 
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -11,7 +11,7 @@
 // You should have received a copy of the GNU General Public License along with Rundler.
 // If not, see https://www.gnu.org/licenses/.
 
-use std::{net::SocketAddr, str::FromStr, sync::Arc};
+use std::{net::SocketAddr, str::FromStr, sync::Arc, time::Duration};
 
 use alloy_primitives::Address;
 use anyhow::Context;
@@ -228,6 +228,28 @@ pub struct BuilderArgs {
         default_value = "0.50"
     )]
     assigner_starvation_ratio: f64,
+
+    /// Initial delay a builder waits before its next submission after the
+    /// submission endpoint rate limited it, doubling with each consecutive
+    /// rate-limited attempt and applied on top of the normal bundle trigger
+    /// interval. Any attempt that is not rate limited resets it.
+    #[arg(
+        long = "builder.rate_limit_backoff_initial_millis",
+        name = "builder.rate_limit_backoff_initial_millis",
+        env = "BUILDER_RATE_LIMIT_BACKOFF_INITIAL_MILLIS",
+        default_value = "1000"
+    )]
+    rate_limit_backoff_initial_millis: u64,
+
+    /// Maximum delay a builder waits between submissions while the submission
+    /// endpoint is rate limiting it.
+    #[arg(
+        long = "builder.rate_limit_backoff_max_millis",
+        name = "builder.rate_limit_backoff_max_millis",
+        env = "BUILDER_RATE_LIMIT_BACKOFF_MAX_MILLIS",
+        default_value = "30000"
+    )]
+    rate_limit_backoff_max_millis: u64,
 }
 
 impl BuilderArgs {
@@ -323,6 +345,10 @@ impl BuilderArgs {
             chain_spec,
             assigner_max_ops_per_request: self.assigner_max_ops_per_request,
             assigner_starvation_ratio: self.assigner_starvation_ratio,
+            rate_limit_backoff_initial: Duration::from_millis(
+                self.rate_limit_backoff_initial_millis,
+            ),
+            rate_limit_backoff_max: Duration::from_millis(self.rate_limit_backoff_max_millis),
         })
     }
 
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -231,13 +231,12 @@ pub struct BuilderArgs {
 
     /// Initial delay a builder waits before its next submission after the
     /// submission endpoint rate limited it, doubling with each consecutive
-    /// rate-limited attempt and applied on top of the normal bundle trigger
-    /// interval. Any attempt that is not rate limited resets it.
+    /// rate-limited attempt. Any attempt that is not rate limited resets it.
     #[arg(
         long = "builder.rate_limit_backoff_initial_millis",
         name = "builder.rate_limit_backoff_initial_millis",
         env = "BUILDER_RATE_LIMIT_BACKOFF_INITIAL_MILLIS",
-        default_value = "1000"
+        default_value = "100"
     )]
     rate_limit_backoff_initial_millis: u64,
 
@@ -247,7 +246,7 @@ pub struct BuilderArgs {
         long = "builder.rate_limit_backoff_max_millis",
         name = "builder.rate_limit_backoff_max_millis",
         env = "BUILDER_RATE_LIMIT_BACKOFF_MAX_MILLIS",
-        default_value = "30000"
+        default_value = "2000"
     )]
     rate_limit_backoff_max_millis: u64,
 }
--- a/crates/builder/src/bundle_sender.rs
+++ b/crates/builder/src/bundle_sender.rs
@@ -35,6 +35,7 @@ use tokio::{
         mpsc::{UnboundedReceiver, UnboundedSender},
         oneshot,
     },
+    time::Instant,
 };
 use tracing::{debug, error, info, warn};
 
@@ -61,6 +62,8 @@ pub(crate) struct Settings {
     pub(crate) max_replacement_underpriced_blocks: u64,
     pub(crate) max_cancellation_fee_increases: u64,
     pub(crate) max_blocks_to_wait_for_mine: u64,
+    pub(crate) rate_limit_backoff_initial: Duration,
+    pub(crate) rate_limit_backoff_max: Duration,
 }
 
 pub(crate) struct BundleSenderImpl<T, C> {
@@ -85,6 +88,10 @@ pub(crate) struct BundleSenderImpl<T, C> {
     /// builders on the route. Final submission outcomes are recorded here and
     /// its state gates poison user operation evidence in the pool.
     provider_event_signal: Arc<ProviderEventSignal>,
+    /// Backoff against a submission endpoint that is rate limiting this builder.
+    /// Outlives the bundle state machine: it is a property of the endpoint, not
+    /// of any one bundle attempt, so state resets must not clear it.
+    rate_limit_backoff: RateLimitBackoff,
 }
 
 pub enum BundleSenderAction {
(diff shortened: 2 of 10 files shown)


bin/rundler/src/cli/builder.rs
```diff
@@ -227,6 +227,28 @@ pub struct BuilderArgs {
...
+
+    /// Initial delay a builder waits before its next submission after the
+    /// submission endpoint rate limited it, doubling with each consecutive
+    /// rate-limited attempt and applied on top of the normal bundle trigger
+    /// interval. Any attempt that is not rate limited resets it.
+    #[arg(
+        long = "builder.rate_limit_backoff_initial_millis",
+        name = "builder.rate_limit_backoff_initial_millis",
+        env = "BUILDER_RATE_LIMIT_BACKOFF_INITIAL_MILLIS",
+        default_value = "1000"
```
dancoombs: This is too high, lets start at 100

bin/rundler/src/cli/builder.rs
```diff
@@ -227,6 +227,28 @@ pub struct BuilderArgs {
...
+    )]
+    rate_limit_backoff_initial_millis: u64,
+
+    /// Maximum delay a builder waits between submissions while the submission
+    /// endpoint is rate limiting it.
+    #[arg(
+        long = "builder.rate_limit_backoff_max_millis",
+        name = "builder.rate_limit_backoff_max_millis",
+        env = "BUILDER_RATE_LIMIT_BACKOFF_MAX_MILLIS",
+        default_value = "30000"
```
dancoombs: This is too high, lets start at 2s

crates/builder/src/sender/fallback.rs
```diff
@@ -170,7 +170,13 @@ impl TransactionSender for FallbackTransactionSender {
...
             }
             Err(
-                e @ (TxSenderError::SenderUnavailable(_) | TxSenderError::UnrecognizedRpc { .. }),
+                e @ (TxSenderError::SenderUnavailable(_)
+                | TxSenderError::UnrecognizedRpc { .. }
+                // A rate limit is capacity we cannot use on this route, so the
+                // fallback is exactly the right place to send the bundle. The
+                // caller still backs off the primary if the fallback is not
+                // configured or is itself rate limited.
+                | TxSenderError::RateLimited(_)),
```
dancoombs: Lets remove this for now. Lets only fallback on errors and let backoff handle the primary.

crates/builder/src/sender/mod.rs
```diff
@@ -149,7 +158,15 @@ impl TxSenderError {
...
                 RpcOutcomeClass::NonTerminal
             }
-            TxSenderError::Underpriced
+            // A rate limit is a statement about our request volume, not about
+            // the bundle: the endpoint never judged the transaction. Counting it
+            // as provider-health evidence would also make suspects of every op
+            // in the bundle, so a rate-limited endpoint would push the whole
+            // pool into isolation - one submission per bundle - while we are
+            // already being told to send fewer requests. Backing off the
+            // endpoint is the dedicated handling instead.
```
dancoombs: Comment is way too long.

crates/builder/src/sender/mod.rs
```diff
@@ -95,6 +95,15 @@ pub(crate) enum TxSenderError {
         /// RPC error message.
         message: String,
     },
+    /// The submission endpoint is rate limiting us.
+    ///
+    /// The transaction was never judged, so this says nothing about the bundle's
+    /// contents - only that we are asking for more capacity than the endpoint
+    /// will give us. Handled by backing the builder off the endpoint rather than
+    /// by retrying at the normal trigger cadence. When a fallback sender is
+    /// configured this triggers failover.
```
dancoombs: Comment is too long

crates/provider/src/traits/error.rs
```diff
@@ -64,6 +65,31 @@ impl ProviderError {
...
+    /// about the endpoint's health, which callers classify differently from
+    /// being asked to send fewer requests.
+    pub fn is_rate_limited(&self) -> bool {
+        let ProviderError::RPC(error) = self else {
+            return false;
+        };
+        match error {
+            RpcError::Transport(TransportErrorKind::HttpError(http)) => http.is_rate_limit_err(),
+            RpcError::Transport(TransportErrorKind::Custom(err)) => {
+                err.to_string().contains("429 Too Many Requests")
```
dancoombs: Where have we seen this type of err before? Where is -32005-style errors pulled from?

crates/builder/src/sender/mod.rs
```diff
@@ -416,6 +433,13 @@ fn create_hard_cancel_tx(to: Address, nonce: u64, gas_fees: GasFees) -> Transact
 
 impl From<ProviderError> for TxSenderError {
     fn from(value: ProviderError) -> Self {
+        // Checked before the error-response paths below: providers signal a rate
+        // limit both as HTTP 429 and as a JSON-RPC error response, and the latter
+        // would otherwise land in `UnrecognizedRpc` and be treated as an ambiguous
+        // judgement of the transaction.
```
dancoombs: Unneeded comment

crates/builder/src/bundle_sender.rs
```diff
@@ -133,6 +140,75 @@ enum SendBundleAttemptResult {
...
+}
+
+/// Per-builder backoff against a submission endpoint that is rate limiting us.
+///
+/// A rate limit is the endpoint asking for fewer requests. Retrying at the normal
+/// trigger cadence answers that by asking for exactly as many, which is enough to
+/// keep a whole fleet of builders pinned against the limit for as long as they
+/// have work: nothing lands, so nothing leaves the pool, so every builder has
+/// work on every trigger. Consecutive rate-limited attempts double the wait;
+/// any attempt that is not rate limited clears it.
```
dancoombs: Remove this part of the comment

crates/builder/src/bundle_sender.rs
```diff
@@ -1064,6 +1179,16 @@ where
                         warn!("Bundle attempt condition not met");
                         Ok(SendBundleAttemptResult::ConditionNotMet)
                     }
+                    // Not an error result: erroring here would reset the
+                    // transaction tracker and log twice per attempt per builder,
+                    // when the bundle is fine and the only thing to do is send
+                    // less often. The caller warns once, with the backoff it
+                    // picked; the endpoint's own wording is only needed when
+                    // diagnosing which limit was hit.
```
dancoombs: Remove this comment.

crates/builder/src/bundle_sender.rs
```diff
@@ -454,6 +550,25 @@ where
...
+                );
+                // Warn rather than error: a rate limit is the endpoint pacing us,
+                // not a fault in the bundle, and at fleet scale erroring on every
+                // attempt buries the logs that matter.
+                warn!(
+                    "Submission endpoint rate limited, backing off {}ms before the next attempt",
+                    delay.as_millis()
+                );
+                self.increment_counter("builder_bundle_txns_rate_limited", &pinned, 1);
+                state.update(InnerState::Building(inner.retry_on_next_trigger()));
```
dancoombs: Why do we need to wait until the next trigger? We should just wait until the backoff is done.

crates/provider/src/traits/error.rs
dancoombs: Both come from alloy rather than from us.

`-32005` is in `ErrorPayload::is_retry_err` (`alloy-json-rpc-1.3.0/src/response/error.rs:86`), which is where alloy keeps the whole provider list, commented per provider:

- `429` — "alchemy throws it this way"
- `-32005` — Infura, "exceeded project rate limit"
- `-32016` + `"rate limit"` in the message — Alchemy, per-IP
- `-32012` + `"credits"` — QuickNode
- `-32007` + `"request limit reached"` — QuickNode
- `1008` — websocket concurrency limit
- `-32055` — Cloudflare
- plus message matching on `rate limit` / `rate exceeded` / `too many requests` / `credits limited` / `request limit` / `maximum number of concurrent requests`

That is the reason detection delegates instead of re-listing wordings here — that list moves, and it is alloy's job to track it.

To answer the first question directly: **we have not seen the stringified shape in rundler.** arb-sepolia returned HTTP 429, which hits the `HttpError` arm via `HttpError::is_rate_limit_err` (`alloy-transport-1.3.0/src/error.rs:145`). The `Custom` arm is lifted verbatim from alloy's `TransportErrorKind::is_retry_err` (`alloy-transport-1.3.0/src/error.rs:114`) and only matters for a transport that loses the status code and leaves `"429 Too Many Requests"` in the error text. I mirrored it because alloy carries it, but it is speculative for our transports — say the word and I will drop it so we only handle shapes we have actually observed.

The doc comment now names all three helpers so the provenance is traceable.

crates/builder/src/bundle_sender.rs
```diff
@@ -164,10 +241,18 @@ where
             // and waiting for the next block trigger. The signing key is not needed
             // during the wait, so it is returned to the pool where it can be borrowed
             // for other work such as sponsored undelegation.
-            if state.is_signer_releasable() {
+            // A rate-limited builder waits out its backoff instead of the trigger,
+            // so an outstanding backoff is idle time too. `end_cycle` is a no-op
+            // while transactions are pending.
+            if state.is_signer_releasable() || self.rate_limit_backoff.is_waiting() {
```
dancoombs: What happens here if UOs are assigned to a signer?

dancoombs: Nothing — the key is never returned while operations are still assigned to it, because `end_cycle`'s own guard is strictly stronger than the assigner's.

`end_cycle` releases only when `transactions.is_empty()`, whereas `num_pending_transactions()` counts entries with `tx_hash.is_some()`. So `transactions.is_empty()` implies `num_pending_transactions() == 0`, which is exactly the condition the `RateLimited` arm released the assigner locks under one statement earlier. Case by case at this line:

- **No transactions tracked** — `release_all` ran in the `RateLimited` arm, and `end_cycle` returns the key. Both released, consistent.
- **Transactions tracked, none with a hash** — `release_all` ran, but `end_cycle` no-ops and the key is kept. Conservative in the safe direction.
- **A transaction in flight** — `release_all` was skipped and `end_cycle` no-ops. Both held, so the tracker keeps the key it needs to escalate or cancel that transaction.

The dangerous ordering would be key-released-while-locks-held: `begin_cycle` can lease a *different* address (`NeedsFunding`), `self.sender_eoa` is then reassigned, and a later `release_all(self.sender_eoa)` would target the new address and orphan the old one's locks. That requires `transactions.is_empty()` and `num_pending_transactions() > 0` simultaneously, which the definitions make impossible.

Worth noting the third case explicitly: a rate-limited builder with an in-flight transaction holds its signer for the whole backoff, up to `--builder.rate_limit_backoff_max_millis` (2s). That is deliberate — same reason the `num_pending_transactions() == 0` guard exists throughout `handle_building_state` — but it does mean the backoff is not always signer-free.

`RateLimitBackoff::record` is called only in that one arm, and `wait()` takes `retry_after`, so `is_waiting()` is true for at most the single loop iteration immediately following a rate-limited attempt. Nothing runs in between that could assign more operations.

I have tightened the comment to state the invariant rather than just "`end_cycle` is a no-op while transactions are pending", since that clearly did not answer the question. Happy to add a regression test if you want the invariant pinned, though it would need a real tracker rather than the mock to be meaningful.

dancoombs: The replacement case is already safe, but not for the reason the guard appears to give — and chasing it turned up something more interesting about this line.

**Replacement attempts are covered.** A successful send always pushes a `PendingTransaction { tx_hash: Some(..) }`, so while that transaction exists `num_pending_transactions() >= 1` and `release_all` is skipped. Confirmed sender locks therefore cannot be dropped out from under an in-flight replacement. The invariant is: confirmed locks exist ⇒ a send succeeded ⇒ a hashed entry exists ⇒ the guard holds.

Worth knowing that `num_pending_transactions()` counts only entries with a hash, and the underpriced ladder pushes `tx_hash: None` entries (that is how `get_state()` keeps escalating fees). So mid-ladder with no prior successful send the count is 0 and `release_all` does run — but in that case nothing was ever confirmed either, so nothing meaningful is lost.

**The thing I did not expect:** this arm's `release_all` is very nearly a no-op for the work itself. `send_bundle` already ends with

```rust
_ => self.assigner.confirm_senders_drop_unused(self.sender_eoa, &[])?,
```

on every non-success result, `RateLimited` included, which drops every lock still in `LockState::Assigned` *before* `handle_building_state` runs. So this attempt's freshly assigned operations go back to the pool no matter what the arm does. What `release_all` adds is clearing the pin and isolation bookkeeping, plus confirmed locks — which the guard protects.

So to answer directly: the arm is not what returns the work, and removing it does not make a rate-limited retry resubmit the same bundle. If we want that — and I think there is a real argument for it, since a refused request judged nothing and we intend to resubmit within `rate_limit_backoff_max` (2s) — the change belongs in that `confirm_senders_drop_unused` call in `send_bundle`, giving `RateLimited` its own arm that drops nothing. That also touches how the underpriced and condition-not-met paths behave, so I would rather not fold it into this PR without you weighing in.

For transparency: I removed the arm's `release_all` first, then wrote a test asserting the builder still held its assignments — the test failed at 0, which is what exposed the `confirm_senders_drop_unused` behaviour above. Reverted in 6ea5377a, along with the signer-release-during-backoff change from 2eaa5781, whose justification depended on the same wrong premise. The signer is now simply held for the duration of the backoff (`retry()` leaves `wait_for_trigger: false`, so `is_signer_releasable()` is false), which is at most 2s.

pavelm (approved):

PR #1324, opened 2026-08-05
fix(builder): treat unsupported transaction type as a terminal RPC error

dancoombs:
Soneium's RPC rejects type `0x04` transactions with `-32000: failed to unmarshal tx: transaction type not supported`. Rundler makes a bundle type `0x04` if *any* op carries an `eip7702Auth`, so such an op fails every bundle it joins until it gets quarantined.

That rejection was classified non-terminal, which is wrong: the node cannot decode the transaction envelope, so it never evaluated the transaction and never will until it is upgraded. Retrying it unchanged cannot succeed.

## What went wrong

Quarantine worked. On soneium-mainnet 2026-08-05 the ops crossed the suspect threshold, were excluded from normal bundles, and were isolated — all as designed. Removal never fired:

| Signal (09:00–15:00 UTC) | Value |
|---|---|
| `transaction type not supported` (log lines / 30 min) | 48 before 10:30 → 14,928 → 27,880 |
| `num_suspect_ops` | steady at 8 |
| `ep_isolation_assignments` | 68.7k → 227k |
| `provider_event_active` | 1 continuously from 10:20 UTC |
| `num_suspects_removed` | **0** |

Because the error was non-terminal, every rejection fed the provider-event window — and an active event pauses suspect removal progress. The ops' own rejections therefore froze their removal counters. With removal paused and the success that would clear them impossible by construction, they had no exit and re-isolated until they aged out of the pool.

The frozen counter also pinned the retry at the 1s floor: `reschedule_retry` derives its backoff exponent from `failures - suspect_threshold`, and `failures` was stuck at exactly the threshold, so the exponent stayed 0 and the backoff curve never engaged.

Blast radius was bounded. The ops stayed quarantined the whole time, so bundling was not blocked and other ops were unaffected — this is not the entrypoint-blocking case the poison-op design was built to prevent. The cost was wasted submissions plus a handful of ops that never resolved.

## Proposed Changes

  - `promote_terminal_error` recognizes `transaction type not supported` as terminal. A terminal error removes a singleton immediately and bypasses provider-health gating, so the op dies on its first isolation attempt.
  - Matched as a substring with no code gate: op-geth wraps it (`failed to unmarshal tx: ...`), and the phrase names its own condition, so unlike `internal error` it needs no code to disambiguate. Only `-32000` has been observed for it; matching distinctive messages alone rather than codes is the existing convention in `classify_submission_error`.
  - Extracts the rules into `is_terminal_rejection`, one early-return each. The previous nested boolean mixed three message rules with two different guards (error code and chain spec) and was no longer readable. No behavior change.

## Testing

`make test-unit` (495 passed), `make lint`, `cargo +nightly fmt --all --check` all clean. New tests:

  - `promotes_unsupported_transaction_type_to_terminal` — op-geth's wrapped wording, bare wording, mixed case, surrounding whitespace, and both `-32000` and `-32603`
  - `does_not_promote_unrelated_unsupported_messages` — near-misses stay non-terminal

## Not in scope

**The general case.** Error responses feed the provider-event window, so any deterministically rejected transaction can still manufacture the event that freezes its own removal. This PR fixes one error string, so we remain a message behind on the general case.

An earlier revision closed that by excluding all error responses from the window. It was dropped as unsafe: the same chain returns `Transaction verification service temporarily unavailable` in bursts of 2–3k per 30 min, an error *response* that genuinely does mean the provider cannot serve, and excluding all responses would lose outage detection exactly where it is warranted. A workable version needs to separate provider-unavailability responses from judgments about the transaction — likely by classifying recognized unavailability wording as `SenderUnavailable` (non-terminal, health evidence, triggers fallback) and excluding only responses that stay unrecognized. Worth its own PR.

**The concurrent bundling outage**, which this does not explain. Successful bundle transactions went 436 per 10 min → ~0 at 10:30 UTC and stayed there. Two tempting explanations are both ruled out:

  - Not a pool full of 7702 ops: `num_7702_ops_added` totals 36 for the life of the pod, five of them between 07:20 and 08:50 UTC and none during the outage, against ~1200 ops per 10 min of inflow.
  - Not isolation starving normal work: `Selected bundle for` ran 35k–53k per 30 min after 10:20 with `Sending isolation bundle` at 10k–27k of it, leaving ~20k–36k normal bundles per 30 min — higher than the ~27k before the incident.

Bundles kept being formed and sent and essentially none mined. `replacement transaction underpriced` dominates the logs at 52k–95k per 30 min, but it was *already* 52k–60k from 09:00–10:00 while bundling was healthy, so it does not discriminate. Separate investigation.

**Soneium's node upgrade.** The superchain registry shows Isthmus active since 2025-05-09, so type `0x04` should work at the protocol level; the sequencer is behind. `eip7702_enabled = false` in the soneium chain spec remains the faster mitigation and needs someone with deploy access.

### DIFF
--- a/crates/builder/src/sender/mod.rs
+++ b/crates/builder/src/sender/mod.rs
@@ -188,8 +188,10 @@ fn is_terminal_rejection(message: &str, code: i64, chain_spec: &ChainSpec) -> bo
 
     // The node cannot decode the transaction envelope, so it never evaluated
     // the transaction and never will until it is upgraded. Substring-matched
-    // under any code: clients wrap it (op-geth reports "failed to unmarshal
-    // tx: transaction type not supported") and number it differently.
+    // because op-geth wraps it ("failed to unmarshal tx: ..."); un-gated by
+    // code because the phrase names its own condition, unlike the generic
+    // messages below. Matching distinctive messages alone is the convention in
+    // `classify_submission_error`.
     if trimmed
         .to_ascii_lowercase()
         .contains("transaction type not supported")
--- a/crates/builder/src/sender/mod.rs
+++ b/crates/builder/src/sender/mod.rs
@@ -161,44 +161,56 @@ impl TxSenderError {
     }
 
     /// Promotes an ambiguous `UnrecognizedRpc` response to `TerminalRpcError`
-    /// when its message is known to be a terminal, per-transaction rejection
-    /// rather than a transient or provider-health condition.
+    /// when its message is a known terminal, per-transaction rejection rather
+    /// than a transient or provider-health condition.
     ///
-    /// Three message shapes are recognized. The first two require code
-    /// `-32000`:
-    /// - `"internal error"` — only promoted on chains with
-    ///   `ChainSpec::internal_rpc_error_is_terminal`, since some providers also use
-    ///   this generic wording for transient outages.
-    /// - `"Transaction rejected by chain policy"` — Robinhood's unambiguous wording
-    ///   for the same rejection; promoted unconditionally, since no chain
-    ///   legitimately emits this phrase for a transient condition.
-    /// - `"transaction type not supported"` — the node cannot decode the
-    ///   transaction envelope, so it never evaluated the transaction and never
-    ///   will until it is upgraded. Matched as a substring under any error code,
-    ///   because clients wrap it differently (op-geth reports
-    ///   `"failed to unmarshal tx: transaction type not supported"`) and report
-    ///   it under codes that vary by client. Retrying is futile and the
-    ///   rejection is unambiguous, so no chain gate applies.
+    /// The recognized messages and their conditions are in
+    /// [`is_terminal_rejection`].
     pub(crate) fn promote_terminal_error(self, chain_spec: &ChainSpec) -> Self {
         match self {
-            TxSenderError::UnrecognizedRpc { code, message } => {
-                let trimmed = message.trim();
-                let is_terminal = trimmed
-                    .to_ascii_lowercase()
-                    .contains("transaction type not supported")
-                    || (code == -32000
-                        && (trimmed.eq_ignore_ascii_case("Transaction rejected by chain policy")
-                            || (chain_spec.internal_rpc_error_is_terminal
-                                && trimmed.eq_ignore_ascii_case("internal error"))));
-                if is_terminal {
-                    TxSenderError::TerminalRpcError { code, message }
-                } else {
-                    TxSenderError::UnrecognizedRpc { code, message }
-                }
+            TxSenderError::UnrecognizedRpc { code, message }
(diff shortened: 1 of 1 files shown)


crates/builder/src/sender/mod.rs
```diff
@@ -158,26 +160,36 @@ impl TxSenderError {
...
                 let is_terminal = trimmed
-                    .eq_ignore_ascii_case("Transaction rejected by chain policy")
-                    || (chain_spec.internal_rpc_error_is_terminal
-                        && trimmed.eq_ignore_ascii_case("internal error"));
+                    .to_ascii_lowercase()
+                    .contains("transaction type not supported")
+                    || (code == -32000
+                        && (trimmed.eq_ignore_ascii_case("Transaction rejected by chain policy")
+                            || (chain_spec.internal_rpc_error_is_terminal
+                                && trimmed.eq_ignore_ascii_case("internal error"))));
```
dancoombs: Can we clean this up? Its starting to get out of hand.

dancoombs: Agreed — cleaned up in 6467648.

Pulled the rules out into `is_terminal_rejection`, one early-return each with its matching strategy and rationale stated locally:

```rust
fn is_terminal_rejection(message: &str, code: i64, chain_spec: &ChainSpec) -> bool {
    let trimmed = message.trim();

    // ... substring-matched under any code: clients wrap and number it differently
    if trimmed.to_ascii_lowercase().contains("transaction type not supported") {
        return true;
    }

    // The remaining rules are specific to this code.
    if code != -32000 {
        return false;
    }

    // ... no chain legitimately emits this phrase for a transient condition
    if trimmed.eq_ignore_ascii_case("Transaction rejected by chain policy") {
        return true;
    }

    // ... terminal only where the chain spec says so
    chain_spec.internal_rpc_error_is_terminal && trimmed.eq_ignore_ascii_case("internal error")
}
```

The `-32000` guard is now a single early return instead of being threaded through every branch, and `promote_terminal_error` is down to a match guard. The long doc comment enumerating all three messages is gone too — that rationale lives next to each rule now instead of being duplicated above the function.

No behavior change. Existing tests unchanged and passing.

jakehobbs:
Codex review question (AI-assisted): Could treating “transaction type not supported” as terminal be too strong when the raw sender is acting as a fallback? For example, if a Flashbots, Bloxroute, or private primary is temporarily unavailable but supports type 0x04, while the raw fallback node does not, the fallback returns TerminalRpcError and poison-op handling may remove a valid singleton before the primary recovery interval elapses. Is there an assumption that every configured route has identical transaction-type support, or should this remain nonterminal in that fallback scenario?

jakehobbs (approved):
Approved by Codex (AI-assisted review). The fallback-routing concern was left separately as a non-blocking question.

PR #1327, opened 2026-08-13
feat(deps): enable gzip/brotli response compression for RPC clients

GregoryElliott:
## Proposed Changes

- Enable `gzip`/`brotli` on the shared `reqwest` dependency (`Cargo.toml`) so rundler's HTTP client advertises `Accept-Encoding` and transparently decompresses responses on every RPC request, for every network.

That's it — one line in `Cargo.toml`, plus the resulting `Cargo.lock` churn.

## Why

Archv2 supports gzip/brotli response compression when a client asks for it via `Accept-Encoding`, but rundler never did — `reqwest` was compiled without those features and nothing sent the header.

An earlier version of this PR scoped compression to `d.alchemy.com` hosts only, with a Flashbots-relay-specific opt-out, to avoid changing behavior for other endpoints. Per [review feedback](https://github.com/alchemyplatform/rundler/pull/1327#discussion_r3779431918), that scoping was unnecessary complexity: HTTP content negotiation is designed to be a safe no-op when unsupported — a client sends `Accept-Encoding`, and a server that doesn't support compression just ignores it and returns the body uncompressed, identical to today. There's no failure mode to guard against, so every RPC client in the binary (archv2, a self-hosted node via `NODE_HTTP`, the Flashbots relay, `localhost` in dev/tests) now negotiates compression the same way, and only the endpoints that actually support it will ever compress a response.

## Validation

- `cargo test -p rundler-provider --lib --features test-utils`: 36/36 passed (unchanged from before this PR — no new branching logic to test)
- `cargo test -p rundler-builder --lib`: 143/143 passed (unchanged)
- `cargo clippy --all --all-features --tests -- -D warnings`: clean
- `cargo +nightly fmt --check`: clean
- Manually measured against a real compression-supporting JSON-RPC endpoint: a 443,917-byte response compressed to 196,187 bytes on the wire (~56% reduction), decoding back to byte-identical JSON.

## Rollout

This changes wire behavior for every RPC call on every network (an `Accept-Encoding` header is now sent everywhere), though it's a no-op for any endpoint that doesn't support compression. I'd still suggest a staged rollout (e.g. one low-traffic network first) and confirming via response headers against a real archv2 endpoint before wider rollout, purely as standard caution for a change that touches every request path.


crates/builder/src/sender/flashbots.rs
```diff
@@ -252,7 +252,15 @@ struct FlashbotsClient {
...
-            http_client: Client::new(),
+            // Compression is only negotiated with Alchemy's internal RPC
+            // backend (see rundler_provider::alloy); explicitly opt out here
+            // so enabling gzip/brotli support in the binary doesn't change
+            // negotiated behavior with the Flashbots relay.
+            http_client: Client::builder()
+                .no_gzip()
+                .no_brotli()
+                .build()
+                .expect("failed to build reqwest client"),
```
dancoombs: Is this required? My assumption would be that the negotiation would fail if flashbots doesn't support gzip/brotli, but if it did we'd get it? Seems reasonable to leave as default unless this actually breaks something.

crates/provider/src/alloy/mod.rs
```diff
@@ -80,10 +80,41 @@ pub fn new_alloy_evm_provider(
...
+    host == ALCHEMY_INTERNAL_RPC_DOMAIN || host.ends_with(ALCHEMY_INTERNAL_RPC_SUBDOMAIN_SUFFIX)
+}
+
+/// Builds the `reqwest::Client` used for RPC requests to `rpc_url`, requesting
+/// response compression only when `rpc_url` is on Alchemy's internal RPC
+/// domain.
+fn build_http_client(rpc_url: &Url) -> reqwest::Result<reqwest::Client> {
+    let mut builder = reqwest::Client::builder();
+    if !host_supports_response_compression(rpc_url) {
+        builder = builder.no_gzip().no_brotli();
```
dancoombs: Similar question here, can we just let HTTP content negotiation handle this automatically?

Cargo.toml
```diff
@@ -76,6 +76,7 @@ auto_impl = "1.2.0"
 aws-config = { version = "1.5.6", default-features = false, features = ["rt-tokio", "rustls"] }
 cargo-husky = { version = "1", default-features = false, features = ["user-hooks"] }
 const-hex = "1.12.0"
+flate2 = "1"
```
dancoombs: Why was this needed?

dancoombs: Looks like testing

dancoombs (approved):

dancoombs:
Can you update the PR description with the latest?

PR #1328, opened 2026-08-17
fix(builder): fall back to node_http, not submit_url

dancoombs:
Fixes the failover bug behind today's `RundlerRPCProviderHTTPSuccessRateBelowThreshold` page on polygon-mainnet prod ([PagerDuty Q1IF1NQHEN89EB](https://down-to-lunch.pagerduty.com/incidents/Q1IF1NQHEN89EB)).

## What broke

The raw sender injected as the failover target resolved its submit URL the same way the primary does, so it inherited `BUILDER_SUBMIT_URL`:

```rust
submit_url: self.submit_url.clone().unwrap_or_else(|| rpc_url.into()),
```

For the senders that actually read that variable it is the *primary's own* endpoint. On polygon-mainnet prod it is Polygon's private gateway (`helm/prod/use1/rundler-polygon-mainnet/values.yaml`), so when the gateway began rejecting our access key, failover built a raw sender pointed back at the same dead gateway — this time calling plain `eth_sendRawTransaction` instead of the private variant, producing exactly the observed error:

```
-32601: jsonrpc method 'eth_sendRawTransaction' is unavailable for this access key
```

The public mempool was never reached and 100% of bundle sends failed for the duration of the incident.

The rest of the failover chain was already correct: the gateway's `-32601` maps to `TxSenderError::UnrecognizedRpc`, which `crates/builder/src/sender/fallback.rs` counts toward the failure threshold and does activate the fallback on. The fallback's URL was the only defect.

## Proposed Changes

  - The fallback raw sender always submits to `rpc_url` (`NODE_HTTP`) and never reads `submit_url`, since `submit_url` belongs to the primary
  - Four tests covering the fallback URL for `polygonprivate` and `flashbots` primaries, plus the raw-primary cases that must keep honoring `submit_url`. The two fallback tests fail against the old logic (verified by reverting the fix) and pass with it
  - Corrects the stale `submit_url` doc comment that invited the mistake — it claimed the flag is only read when the sender is `raw`, but `polygonprivate` reads it too — and documents `sender_recovery_interval_secs` / `sender_failure_threshold` in `docs/cli.md`, which were previously undocumented

## Not addressed here

`BUILDER_SENDER_RECOVERY_INTERVAL_SECS` is silently a no-op when `BUILDER_SENDER` is `raw`, because `maybe_with_fallback` is only applied to the non-raw match arms. `helm/stg/use1/rundler-base-sepolia/values.yaml` sets the interval to `60` with no `BUILDER_SENDER` and a dedicated Base sequencer `submit_url`, so that deployment has no failover despite being configured as though it does. The flag's own docs say "for non-raw sender types", so this is current documented behavior rather than the reported bug — left out to avoid changing failover behavior for the Base/OP fleets in an incident fix. Happy to wire it up in a follow-up.

### DIFF
--- a/bin/rundler/src/cli/builder.rs
+++ b/bin/rundler/src/cli/builder.rs
@@ -96,9 +96,8 @@ pub struct BuilderArgs {
     /// If present, the url of the ETH provider that will be used to send
     /// transactions. Defaults to the value of `node_http`.
     ///
-    /// Only used when BUILDER_SENDER is "raw" or "polygonprivate". Never used
-    /// for the raw sender that BUILDER_SENDER_RECOVERY_INTERVAL_SECS falls back
-    /// to, which always submits to `node_http`.
+    /// Only used when BUILDER_SENDER is "raw" or "polygonprivate". Never used by
+    /// the fallback sender, which always submits to `node_http`.
     #[arg(
         long = "builder.submit_url",
         name = "builder.submit_url",
@@ -332,12 +331,10 @@ impl BuilderArgs {
         chain_spec: &ChainSpec,
         rpc_url: &str,
     ) -> anyhow::Result<TransactionSenderArgs> {
-        // The raw sender used to stand in for an unavailable primary. It always
-        // submits to `rpc_url` (`NODE_HTTP`) and never to `submit_url`:
-        // `submit_url` is the primary's own endpoint for the senders that read it
-        // (e.g. Polygon's private gateway for `polygonprivate`), so honoring it
-        // here would point the fallback straight back at the outage it exists to
-        // route around.
+        // The raw sender that stands in for an unavailable primary. It submits
+        // to `rpc_url` rather than `submit_url` because `submit_url` is the
+        // primary's own endpoint (`polygonprivate` submits there), and a
+        // fallback is only useful on an endpoint independent of the primary.
         let fallback_args = || RawSenderArgs {
             submit_url: rpc_url.into(),
             use_conditional_rpc: false,
@@ -614,8 +611,7 @@ mod tests {
             }
             other => panic!("expected a polygon private primary, got {other:?}"),
         }
-        // ...while the fallback goes to the public mempool. Reusing submit_url
-        // here would send the fallback back into the primary's outage.
+        // ...while the fallback reaches the public mempool through node_http.
         assert_eq!(unwrap_raw(*fallback.fallback).submit_url, RPC_URL);
     }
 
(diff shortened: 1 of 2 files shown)


Richard-Dang (dismissed):

bin/rundler/src/cli/builder.rs
```diff
@@ -96,7 +96,9 @@ pub struct BuilderArgs {
     /// If present, the url of the ETH provider that will be used to send
     /// transactions. Defaults to the value of `node_http`.
     ///
-    /// Only used when BUILDER_SENDER is "raw"
+    /// Only used when BUILDER_SENDER is "raw" or "polygonprivate". Never used
+    /// for the raw sender that BUILDER_SENDER_RECOVERY_INTERVAL_SECS falls back
```
dancoombs: This is confusing `BUILDER_SENDER_RECOVERY_INTERVAL_SECS` is a config. We should just call this by its feature name.

bin/rundler/src/cli/builder.rs
```diff
@@ -330,8 +332,14 @@ impl BuilderArgs {
...
         rpc_url: &str,
     ) -> anyhow::Result<TransactionSenderArgs> {
-        let raw_args = || RawSenderArgs {
-            submit_url: self.submit_url.clone().unwrap_or_else(|| rpc_url.into()),
+        // The raw sender used to stand in for an unavailable primary. It always
+        // submits to `rpc_url` (`NODE_HTTP`) and never to `submit_url`:
+        // `submit_url` is the primary's own endpoint for the senders that read it
+        // (e.g. Polygon's private gateway for `polygonprivate`), so honoring it
+        // here would point the fallback straight back at the outage it exists to
+        // route around.
```
dancoombs: Don't refer to old code in comments.

bin/rundler/src/cli/builder.rs
dancoombs: Good call — now reads "Never used by the fallback sender, which always submits to `node_http`." Same wording in `docs/cli.md`.

bin/rundler/src/cli/builder.rs
dancoombs: Rewritten to describe only what this code does:

```rust
// The raw sender that stands in for an unavailable primary. It submits
// to `rpc_url` rather than `submit_url` because `submit_url` is the
// primary's own endpoint (`polygonprivate` submits there), and a
// fallback is only useful on an endpoint independent of the primary.
```

Dropped the same kind of back-reference from the test comment too.

pavelm (approved):

PR #1329, opened 2026-08-18
fix(builder): treat underpriced bundle validation as retryable, not as a bundle failure

dancoombs:
## Proposed Changes

Bundle building on `hyperliquid-mainnet` has been failing continuously and paging on-call (`RundlerRPCProviderRPCSuccessRateBelowThreshold`, `eth_call`). Peak measured today: **38 failed bundle attempts/sec** and ~70 error log lines/sec from a single pod.

**Root cause.** HyperEVM interleaves two block types in one block-number sequence, each with its own EIP-1559 base fee market: small blocks (3M gas, ~1s) and big blocks (30M gas, rare). `eth_call` validates `maxFeePerGas` against the **big block** base fee, while `eth_feeHistory`, `eth_gasPrice`, and the block header all report the **small block** base fee. Verified against the chain by binary-searching the enforced threshold:

```
eth_gasPrice         (small lane) :    152381883  (0.1524 gwei)
eth_bigBlockGasPrice (big lane)   :   1352814512  (1.3528 gwei)
eth_call enforced threshold       :   1352814552  (1.3528 gwei)

threshold / bigBlockGasPrice = 1.000000     <- exact
threshold / small gasPrice    = 8.88x
```

Pinning the call to a specific small block does not help; that block's header reports 0.130 gwei and the call still rejects 0.2876 gwei. The public `rpc.hyperliquid.xyz/evm` behaves identically, so this is chain behavior rather than an Alchemy node issue.

Our bundles are capped at ~1.89M gas (`CHAIN_BLOCK_GAS_LIMIT=2000000` x 0.9 max ratio, +5% overhead), so they land in small blocks where the small-lane base fee is the real cost. The fee is correct; the validation check is comparing against the wrong lane.

**Why it livelocked.** `-32003` arrives as an RPC transport error, so `call_handle_ops` returned `Err`, which became a hard error at `bundle_proposer.rs` (`.context("should call handle ops with candidate bundle")?`) and landed in the bundle sender's catch-all `Err` arm. No op rejected, no pool mutation, no fee escalation — the next trigger rebuilt an identical bundle. Same structural shape as the empty-revert livelock investigated in July.

Single-op v0.7+ bundles skip the validation call entirely, which is why hyperliquid kept landing 20–49 bundles/hr throughout; only multi-op bundles and all v0.6 bundles hit the call.

Two changes:

  - **Classify the rejection as retryable.** New `HandleOpsOut::Underpriced`, returned from v0.6/v0.7 `call_handle_ops` when the node reports a base-fee-too-low error. The proposer surfaces it as `BundleProposerError::SimulationUnderpriced` and the sender maps it to the existing `SendBundleAttemptResult::Underpriced`, so fees escalate and the round terminates instead of spinning. This is chain-agnostic and also covers the story-mainnet class of fee-floor problem.

  - **Add a `bundle_simulation_omit_gas_fees` chain spec option** that sends the validation call without fee caps, skipping the node's fee validation. Verified that both an unset and a zero `maxFeePerGas` pass on HyperEVM. Only the validation call is affected — `get_send_bundle_transaction` always passes `Some(gas_fees)`, so the submitted transaction keeps its real caps. Defaults to `false`, so no chain changes behavior without opting in.

  - **The post-op-revert recheck now aborts on an underpriced rejection** rather than treating it as evidence. Previously it would have blamed the op it could not test, or fallen through to "couldn't identify the offender, remove everything" and evicted the whole bundle.

  - `chore(docker)`: exclude `.claude/worktrees` from the Docker build context. Those are full repo checkouts; they bloat the planner `COPY`, give `cargo chef prepare` nested workspaces to trip over, and get walked by the `recipe-original` rsync.

Not customer-facing: the gas estimation path builds its calls with `max_fee_per_gas(0)`, which passes the check, so only the builder's validation call was affected.

**Tests:** 5 new unit tests — error-message classification, fee caps present vs omitted on the built request (asserting nothing else about the request changes), `Underpriced` producing `SimulationUnderpriced` with no ops rejected, and the post-op-revert recheck aborting instead of evicting. `make test-unit` 525 passed, clippy and `cargo +nightly fmt --check` clean.

**Follow-ups (not in this PR):**
  - `ws-tools`: set `CHAIN_BUNDLE_SIMULATION_OMIT_GAS_FEES: "true"` on `rundler-hyperliquid-mainnet` / `-testnet` to actually enable the second fix. The helm values configure the chain spec purely via `CHAIN_*` env vars (no `--network`), so no hardcoded chain spec is needed.
  - `CHAIN_BLOCK_GAS_LIMIT` is `2000000` for hyperliquid, but small blocks are 3M today — likely stale from onboarding. Raising it toward 3M stays in the small lane; going above 3M would push bundles into big-block territory where the big-lane fee genuinely applies. Wants a wider sample of block gas limits before changing.
  - The assigner logs "doesn't meet fee requirements" at INFO for every builder x op evaluation (~150 lines/sec for one stuck op) — worth rate-limiting.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,9 +2,6 @@
 /.github
 /.githooks
 /.vscode
-# Agent worktrees are full repo checkouts. Left in the build context they bloat the
-# planner COPY, confuse `cargo chef prepare` with nested workspaces, and get walked
-# by the recipe-original rsync.
 /.claude
 /test
 *.iml
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,10 @@
 /.github
 /.githooks
 /.vscode
+# Agent worktrees are full repo checkouts. Left in the build context they bloat the
+# planner COPY, confuse `cargo chef prepare` with nested workspaces, and get walked
+# by the recipe-original rsync.
+/.claude
 /test
 *.iml
 *.rs.bk
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -101,10 +101,6 @@ pub(crate) enum BundleProposerError {
     NoOperationsAfterFeeFilter,
     /// The node rejected the bundle validation call because the bundle's fee caps
     /// were below the base fee it validates against.
-    ///
-    /// Carries no information about individual operations, so no pool state is
-    /// mutated. The caller should retry at a higher fee rather than treat this as a
-    /// bundle failure.
     #[error("Bundle validation call underpriced")]
     SimulationUnderpriced,
     #[error(transparent)]
--- a/crates/builder/src/bundle_proposer.rs
+++ b/crates/builder/src/bundle_proposer.rs
@@ -99,12 +99,24 @@ pub(crate) type BundleProposerResult<T> = std::result::Result<T, BundleProposerE
 pub(crate) enum BundleProposerError {
     #[error("No operations after fee filtering")]
     NoOperationsAfterFeeFilter,
+    /// The node rejected the bundle validation call because the bundle's fee caps
+    /// were below the base fee it validates against.
+    ///
+    /// Carries no information about individual operations, so no pool state is
+    /// mutated. The caller should retry at a higher fee rather than treat this as a
+    /// bundle failure.
+    #[error("Bundle validation call underpriced")]
+    SimulationUnderpriced,
     #[error(transparent)]
     ProviderError(#[from] rundler_provider::ProviderError),
     /// All other errors
     #[error(transparent)]
     Other(#[from] anyhow::Error),
 }
+
+/// A boxed future resolving to the bundle indices a post-op-revert check wants removed.
+type PostOpRevertCheck<'a> =
+    Pin<Box<dyn Future<Output = BundleProposerResult<Vec<usize>>> + Send + 'a>>;
 
 /// Version-agnostic bundle data for transaction submission.
 /// This is the type-erased result of bundle proposal.
@@ -552,7 +564,13 @@ where
                         .unwrap_or_default(),
                 ))
             }
-            None | Some(HandleOpsOut::Revert(_)) | Some(HandleOpsOut::PostOpRevert) => {
+            // `Underpriced` is a pre-execution rejection from an `eth_call` and is
+            // never produced by decoding an onchain revert, but it is unattributable
+            // if it ever shows up here.
+            None
+            | Some(HandleOpsOut::Revert(_))
+            | Some(HandleOpsOut::PostOpRevert)
+            | Some(HandleOpsOut::Underpriced) => {
                 warn!(
                     "unattributed revert for bundle of {} ops",
                     all_op_hashes.len()
@@ -1268,6 +1286,15 @@ where
 
         match handle_ops_out {
             HandleOpsOut::Success => Ok(Some(gas_limit)),
+            HandleOpsOut::Underpriced => {
+                // The node rejected the call before executing it, so there is nothing
+                // to attribute to an op. Bail out without touching the bundle so the
+                // sender can escalate fees instead of retrying an identical bundle.
+                warn!(
+                    "Bundle validation call rejected as underpriced by the node, will retry with higher fees."
+                );
+                Err(BundleProposerError::SimulationUnderpriced)
+            }
             HandleOpsOut::FailedOp(index, message) => {
                 self.emit(BuilderEvent::rejected_op(
                     self.builder_tag.clone(),
@@ -1476,10 +1503,10 @@ where
         context: &mut ProposalContext<EP::UO>,
         gas_limit: u64,
         bundle_fees: GasFees,
-    ) -> anyhow::Result<()> {
+    ) -> BundleProposerResult<()> {
         let agg_groups = context.to_ops_per_aggregator();
         let mut op_index = 0;
-        let mut futures: Vec<Pin<Box<dyn Future<Output = Vec<usize>> + Send>>> = vec![];
+        let mut futures: Vec<PostOpRevertCheck<'_>> = vec![];
 
         for agg_group in agg_groups {
             // For non-aggregated ops, re-simulate each op individually
@@ -1509,7 +1536,14 @@ where
         }
 
         let results = future::join_all(futures).await;
-        let mut to_remove = results.into_iter().flatten().collect::<Vec<_>>();
+        // A re-simulation rejected on fees alone is inconclusive: it tells us nothing
+        // about which op reverted. Bail out before rejecting anything so the sender can
+        // retry at a higher fee, rather than falling through to the catch-all below and
+        // evicting the whole bundle.
+        let mut to_remove = Vec::new();
+        for result in results {
+            to_remove.extend(result?);
+        }
         if to_remove.is_empty() {
             // if we can't identify the offending user ops, remove all user ops from the bundle and from the pool
             error!(
@@ -1540,7 +1574,7 @@ where
         op_index: usize,
         gas_limit: u64,
         bundle_fees: GasFees,
-    ) -> Vec<usize> {
+    ) -> BundleProposerResult<Vec<usize>> {
         let op_hash = op.hash();
         let bundle = vec![UserOpsPerAggregator {
             aggregator: Address::ZERO,
@@ -1560,21 +1594,24 @@ where
             )
             .await;
         match ret {
+            // Rejected on fees alone, so this tells us nothing about whether the op
+            // reverted. Abort rather than blame an op we haven't actually tested.
+            Ok(HandleOpsOut::Underpriced) => Err(BundleProposerError::SimulationUnderpriced),
             Ok(out) => {
                 if let HandleOpsOut::PostOpRevert = out {
                     warn!(
                         "PostOpRevert error found, removing {:?} from bundle",
                         op_hash
                     );
-                    vec![op_index]
+                    Ok(vec![op_index])
                 } else {
-                    vec![]
+                    Ok(vec![])
                 }
             }
             Err(e) => {
                 // If we get an error here, we can't be sure if the op is the offending op or not, so we remove it to be safe
                 error!("Failed to call handle ops: {e} during postOpRevert handling, removing op");
-                vec![op_index]
+                Ok(vec![op_index])
             }
         }
     }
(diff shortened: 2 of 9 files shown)


crates/builder/src/bundle_proposer.rs
```diff
@@ -99,13 +99,25 @@ pub(crate) type BundleProposerResult<T> = std::result::Result<T, BundleProposerE
 pub(crate) enum BundleProposerError {
     #[error("No operations after fee filtering")]
     NoOperationsAfterFeeFilter,
+    /// The node rejected the bundle validation call because the bundle's fee caps
+    /// were below the base fee it validates against.
+    ///
+    /// Carries no information about individual operations, so no pool state is
+    /// mutated. The caller should retry at a higher fee rather than treat this as a
+    /// bundle failure.
```
dancoombs: Remove this part of the comment

crates/builder/src/bundle_sender.rs
```diff
@@ -931,6 +931,13 @@ where
                             state.condition_not_met = false;
                             Ok(SendBundleAttemptResult::NoOperationsAfterFeeFilter)
                         }
+                        // The node rejected the validation call on fees alone. Treat it
+                        // like any other underpriced bundle so fees escalate, rather
+                        // than as a bundle failure that retries an identical bundle.
```
dancoombs: Remove this comment

crates/provider/src/alloy/entry_point/mod.rs
```diff
@@ -21,6 +21,19 @@ use rundler_types::authorization::Eip7702Auth;
 pub(crate) mod v0_6;
 pub(crate) mod v0_7;
 
+/// Returns true if an `eth_call` error means the node rejected the call before
+/// execution because its fee caps were below the base fee.
+///
+/// This is a pre-execution check, so it carries no information about the operations
+/// in the bundle and must not be treated as a revert.
```
dancoombs: remove this part

crates/provider/src/alloy/entry_point/v0_6.rs
```diff
@@ -260,12 +261,21 @@ where
         proxy: Option<Address>,
         _validation_only: bool,
     ) -> ProviderResult<HandleOpsOut> {
+        // Some chains validate the fee caps on an `eth_call` against a base fee that
+        // doesn't correspond to the block being executed against, rejecting correctly
+        // priced bundles. Omitting the caps skips that validation.
```
dancoombs: Remove this

crates/provider/src/alloy/entry_point/v0_7.rs
```diff
@@ -316,12 +317,21 @@ where
             );
         }
 
+        // Some chains validate the fee caps on an `eth_call` against a base fee that
+        // doesn't correspond to the block being executed against, rejecting correctly
+        // priced bundles. Omitting the caps skips that validation.
```
dancoombs: Remove this

crates/provider/src/traits/entry_point.rs
```diff
@@ -54,6 +54,12 @@ pub enum HandleOpsOut {
     PostOpRevert,
     /// Call reverted
     Revert(Bytes),
+    /// The node rejected the call before execution because the bundle's fee caps
+    /// were below the base fee it validates against.
+    ///
+    /// This says nothing about the operations in the bundle, so no operation should
+    /// be rejected on its account. The bundle is retryable at a higher fee.
```
dancoombs: Remove this

crates/types/src/chain.rs
```diff
@@ -132,6 +132,21 @@ pub struct ChainSpec {
...
+    /// does not correspond to the block being executed against, so a correctly
+    /// priced bundle is rejected with `-32003: max fee per gas less than block base
+    /// fee`. HyperEVM is one such chain: it runs separate base fee markets for its
+    /// small (3M gas) and big (30M gas) blocks, and `eth_call` validates against the
+    /// big block base fee while `eth_feeHistory` and the block header report the
+    /// small block base fee, which can be an order of magnitude lower.
+    ///
+    /// Omitting the caps skips the node's fee validation while leaving simulation
+    /// otherwise unchanged. Only the validation call is affected; the bundle
+    /// transaction that is actually submitted always carries its real fee caps.
```
dancoombs: Remove this

.dockerignore
```diff
@@ -2,6 +2,10 @@
 /.github
 /.githooks
 /.vscode
+# Agent worktrees are full repo checkouts. Left in the build context they bloat the
+# planner COPY, confuse `cargo chef prepare` with nested workspaces, and get walked
+# by the recipe-original rsync.
```
dancoombs: Remove the comment

crates/builder/src/bundle_proposer.rs
dancoombs: Trimmed to the first sentence.

crates/builder/src/bundle_sender.rs
dancoombs: Removed.

crates/provider/src/alloy/entry_point/mod.rs
dancoombs: Trimmed to the first sentence.

crates/provider/src/alloy/entry_point/v0_6.rs
dancoombs: Removed.

crates/provider/src/alloy/entry_point/v0_7.rs
dancoombs: Removed.

crates/provider/src/traits/entry_point.rs
dancoombs: Cut to a single line, matching the other variants. Kept one line because the crate warns missing_docs and CI runs clippy with -D warnings.

crates/types/src/chain.rs
dancoombs: Cut to the one-line field description. Same missing_docs constraint as the HandleOpsOut variant.

.dockerignore
dancoombs: Removed.

jakehobbs:
Codex review findings:

1. **[P1] Simulation retries do not actually increase fees.** `SimulationUnderpriced` occurs before `TransactionTracker::send_transaction`, so no higher fee floor is recorded. `fee_increase_count` changes, but it is never applied to `bundle_fees`; a retry on the same block can therefore repeat the identical failing `eth_call`. The rejected simulation fees need to be persisted or explicitly increased before entering the underpriced state.

2. **[P2] Preserve pending condition rechecks.** If `condition_not_met` is already set, an underpriced simulation exits before `check_conditions_met` runs. Clearing the flag in this error arm lets the next attempt skip the required storage-condition recheck and retain stale operations. The flag should remain set until `make_bundle` completes successfully.

3. **[P2] Simulation errors clear unrelated submission backoff.** `Underpriced` counts as `endpoint_judged_transaction`, so mapping a simulation-only rejection to it clears the submission rate-limit backoff. When the simulation provider and submission endpoint differ, such as with a Flashbots relay, the relay never saw this request. A distinct result variant should avoid marking the submission endpoint as reached.

Reviewed by Codex.

pavelm (dismissed):

dancoombs:
Checked all three against the code — all confirmed. Fixed in e8cd6240.

**#2 and #3 share a root cause** and are fixed together: reusing `SendBundleAttemptResult::Underpriced`, a submission-time outcome, for a build-time rejection. There's now a distinct `SimulationUnderpriced` variant.

- **#3** — confirmed: `endpoint_judged_transaction()` returns `true` for `Underpriced`, so it cleared the backoff. `SimulationUnderpriced` returns `false`, alongside `RateLimited` and the `NoOperations*` variants. Folded into the existing `test_only_judged_attempts_end_the_backoff`.
- **#2** — confirmed, and my arm was directly contradicting the comment a few lines above it in `send_bundle` ("keep it set for hard failures so the next attempt still re-checks conditions"). No longer cleared; new `test_simulation_underpriced_keeps_condition_not_met`.

**#1 — confirmed, but I've fixed it by dropping the escalation claim rather than by bumping fees.** `bundle_fees` comes from `required_bundle_fees(block_hash, required_fees)`, where `required_fees` is set by `TransactionTracker` only after a send; `fee_increase_count` feeds `required_op_fees` (the op filter), never `bundle_fees`. So the retry did re-issue an identical `eth_call`, exactly as described.

I don't think persisting/bumping the rejected fees is the right remedy, for three reasons:

1. On HyperEVM the check wants ~9x the small-lane base fee. Escalation can't realistically get there, and if it did we'd be overpaying ~9x for a transaction that lands in the small-block lane at the un-escalated fee.
2. `required_op_fees` becomes `bundle_fees` once `fee_increase_count > 0`, so escalating also raises the bar user ops must clear — we'd price out ops to satisfy a check that is wrong about what our transaction costs.
3. For a *genuine* base fee move, the next trigger re-reads fees from the chain via the fee estimator and naturally bids correctly. That's real escalation; a synthetic bump on top of it isn't needed.

So `SimulationUnderpriced` now leaves `fee_increase_count` alone (nothing was submitted, so there's no replacement to bump) and waits for the next trigger to rebuild at re-read fees. Doc comment and `docs/architecture/builder.md` updated to say that instead of claiming escalation.

Worth being explicit about what this half of the PR does and doesn't buy: it makes the failure non-poisoning and correctly classified — no ops rejected, no tracker reset, no `builder_bundle_txns_failed` inflation, no ERROR per attempt, backoff and condition flag preserved. It does **not** stop the `eth_call` from failing, so it alone won't clear the page. That's what `bundle_simulation_omit_gas_fees` is for, and it needs the ws-tools env var to take effect.

One residual worth flagging since it's adjacent to #1: `required_bundle_fees` is cached per block hash, so a retry within the same block re-reads identical fees. With `bundle_max_send_interval_millis` at 1000ms and ~1s hyperliquid blocks that's mostly one attempt per block, but a same-block repeat is possible. I left it alone rather than add block-tracking state — say the word if you want the rebuild gated on a new block.

`make test-unit` 526 passed, clippy and fmt clean.

jakehobbs (approved):
Re-reviewed e8cd6240. The distinct simulation-underpriced path preserves conditional rechecks and submission backoff, while the revised fee behavior correctly waits for fresh chain fees instead of applying a synthetic replacement bump. The regression tests and documentation align with the implementation.

Approved by Codex.

PR #1333, opened 2026-09-04
fix(builder): classify wrapped sequencer errors

jakehobbs:
## Summary
- treat `-32000` RPC responses containing an embedded sequencer `429 Too Many Requests` as rate limits so builder pacing activates
- classify Nitro's max-fee-below-base-fee response as underpriced so bundles retry with higher fees instead of becoming poison-operation/provider evidence
- add regression tests for both production response shapes

## Test plan
- [x] `cargo +nightly fmt --all --check`
- [x] `cargo test -p rundler-provider -p rundler-builder --all-features`
- [x] `cargo clippy -p rundler-provider -p rundler-builder --all-features --tests -- -D warnings`

🤖 Generated with [Codex](https://Codex.com/Codex)

### DIFF
--- a/crates/builder/src/sender/mod.rs
+++ b/crates/builder/src/sender/mod.rs
@@ -560,7 +560,7 @@ mod tests {
 
         let wrapped_rpc_429 = TxSenderError::from(super::rpc_error_response(
             -32000,
-            "sequencer returned 429 Too Many Requests",
+            "upstream returned HTTP 429: Too Many Requests",
         ));
         assert!(matches!(wrapped_rpc_429, TxSenderError::RateLimited(_)));
 
--- a/crates/builder/src/sender/mod.rs
+++ b/crates/builder/src/sender/mod.rs
@@ -558,6 +558,12 @@ mod tests {
         let rpc_429 = TxSenderError::from(super::rpc_error_response(429, "rate limit exceeded"));
         assert!(matches!(rpc_429, TxSenderError::RateLimited(_)));
 
+        let wrapped_rpc_429 = TxSenderError::from(super::rpc_error_response(
+            -32000,
+            r#"429 Too Many Requests: {"jsonrpc":"2.0","error":{"code":429,"message":"Too Many Requests"}}"#,
+        ));
+        assert!(matches!(wrapped_rpc_429, TxSenderError::RateLimited(_)));
+
         let rpc_limit_exceeded =
             TxSenderError::from(super::rpc_error_response(-32005, "limit exceeded"));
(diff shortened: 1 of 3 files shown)


crates/builder/src/sender/mod.rs
```diff
@@ -558,6 +558,12 @@ mod tests {
         let rpc_429 = TxSenderError::from(super::rpc_error_response(429, "rate limit exceeded"));
         assert!(matches!(rpc_429, TxSenderError::RateLimited(_)));
 
+        let wrapped_rpc_429 = TxSenderError::from(super::rpc_error_response(
```
jakehobbs, line 561: do you think this is too specific?

crates/provider/src/traits/error.rs
```diff
@@ -68,18 +68,29 @@ impl ProviderError {
...
-            RpcError::ErrorResp(resp) => resp.is_retry_err(),
+            RpcError::ErrorResp(resp) => {
+                resp.is_retry_err()
+                    // Robinhood's Nitro submission path wraps the sequencer's
+                    // JSON-RPC 429 response in a generic -32000 error.
+                    || (resp.code == -32000
+                        && resp
+                            .message
+                            .to_ascii_lowercase()
+                            .contains("429 too many requests"))
```
jakehobbs: should this logic be within `is_retry_err`?

crates/builder/src/sender/mod.rs
jakehobbs: Yes—the exact embedded JSON was too specific, and calling it a sequencer response asserted provenance we do not have. I changed the fixture to `upstream returned HTTP 429: Too Many Requests`. The fallback now checks that the normalized message contains both `429` and `too many requests` independently, so wrapper punctuation can vary without relying on the full payload shape.

crates/provider/src/traits/error.rs
jakehobbs: Agreed that the generic casing fix ideally belongs in the Alloy `ErrorPayload::is_retry_err` method. The locked Alloy implementation already checks `msg.contains("too many requests")`, but case-sensitively, so the observed title-case message falls through. We cannot change that dependency method here; this remains in the same `ErrorResp` match arm as a narrow compatibility fallback after `resp.is_retry_err()`. The fallback now requires both `429` and `too many requests`. Using the outer `TransportErrorKind::is_retry_err` would be too broad because it also classifies HTTP 503 and missing batch responses as retryable, while this method specifically identifies rate limits.

pavelm (approved):

